PR #5, opened 2018-10-03
Move all ascii art to separate files

zxey:
It is similar to other language matching functions.
This keeps the ascii art more organized, and its easier to add more languages without bloating main.rs


o2sh:
Are you sure the added IO overhead resulting from reading a file from disk is worth it ?

zxey:
This does read from file when running, only at build time. What I think would be even better, is to have some configuration for every file, where you would specify what color to use and what ascii art to use.
Maybe it is not worth it as it is now, so feel free to close this.

o2sh:
Ok, I will merge it.
Don't hesitate to follow up on your idea, it sounds promising !

Issue #2, opened 2018-10-03
Better and more ASCII !! (╯°□°）╯︵ ┻━┻


o2sh:
Some of our ascii logos look a bit out of shape...

plz help ಥ_ಥ

Don't hesitate to add more languages, if you feel inspired. 

Bye.
ฅ^•ﻌ•^ฅ

jonathanfilbert:
Would it be okay to add Coffeescript?

o2sh:
I don't see why not...you might also want to add JavaScript as it's closely related.

felix91gr:
@o2sh, how did you draw these? Maybe with a formal procedure we could synthesize the logos straight from a PNG. There are, afaik, libraries that do this for you :D

o2sh:
Hi @felix91gr,
Except for typescript, R, lua and maybe clojure, I used an [online tool] (https://www.text-image.com/convert/ascii.html) that converts PNG into ascii which is OK but not great.

If you know better alternatives, plz share. :)

PR #6, opened 2018-10-03
Detect dominant language type

### LINKED ISSUE #1: Implement logic for dominant language
Create a function that will scan a project and determine the dominant language (the language is currently hardcoded in the main function)

An idea would be to scan all files based on a set of extension (rs for rust, py for python) and find which extension represents most of the lines of code (in case of a multiple language project).

zxey:
This detects language of source code repository at runtime.

Fixes #1

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,6 @@
 extern crate colored;
+extern crate tokei;
+
 use colored::*;
 use std::fmt;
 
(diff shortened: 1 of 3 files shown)


zxey:
Yeah, I might have a look sometimes, how tokei or loc do their analysis and maybe extract the relevant code, so it is not as big dependency. The reason I decided to go with this, it tokei had quite nice library api next to their command line utility, so I used that.

src/main.rs
```diff
@@ -107,6 +118,57 @@ fn main() {
...
+            tokei::LanguageType::C => Language::C,
+            tokei::LanguageType::Python => Language::Python,
+            tokei::LanguageType::CSharp => Language::Csharp,
+            tokei::LanguageType::Scala => Language::Scala,
+            tokei::LanguageType::Sh => Language::Shell,
+            tokei::LanguageType::Lisp => Language::Lisp,
+            tokei::LanguageType::Haskell => Language::Haskell,
+            tokei::LanguageType::Ruby => Language::Ruby,
+            tokei::LanguageType::R => Language::R,
+            _ => unimplemented!(),
```
Mouvedia: JavaScript is a valid `LanguageType`.

cf https://docs.rs/tokei/6.0.2/tokei/enum.LanguageType.html

o2sh: Of course, this is a work in progress. An issue was opened #2 for that purpose.

Feel free to contribute @Mouvedia 

Thanks 🥇

Issue #9, opened 2018-10-04
Implement logic for project Information


o2sh:
Create a function that will scan a project and retrieve its information: Name, Author, Number of lines, License... 

These info are currently hardcoded in the main function...

zxey:
As @o2sh states in #12 there are only three things left to implement:
Project name - this can be obtained from git by parsing remote origin or some similar info
Project repo - also can be parsed from git metadata
Project license - you can probabyl use [askalono](https://github.com/amzn/askalono) library to detect license of the project

PR #12, opened 2018-10-06
Detect authors

zxey:
Onefetch can now detect first three authors of a project by assuming the project is a git repository.
It counts the first three top commiters and puts them as an authors of the project.

This started as a work on #9 but since then @JoshBrudnak took and interest in working on that issue,
so I am contributing this as only a partial imlementation of full project info logic and maybe he can base his work on this :).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,11 +3,12 @@ extern crate tokei;
 
 use colored::*;
 use std::fmt;
+use std::process::{Command, Stdio};
 
 struct Info {
     project_name: String,
     language: Language,
-    author: String,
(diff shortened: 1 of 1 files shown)


zxey:
The output now looks something like this:
output

o2sh:
Well done @zxey, there is only three infos left to be implemented: Project name, repo and license.

@JoshBrudnak good luck !

Mouvedia:
For npm packages it should prioritize the [`author` and `contributors` fields](https://docs.npmjs.com/files/package.json#people-fields-author-contributors) and, if found, ignore commits.

PR #13, opened 2018-10-06
Added repository name and url detection

JoshBrudnak:
Added repository name and url detection by parsing the repositories remote url. I also formatted the code using rustfmt.
This resolves the url and name part of #9

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,20 +139,6 @@ version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "curl-sys"
-version = "0.4.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "cc 1.0.25 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


zxey:
Just a heads up, this may make it harder for people with macOS to build this project, because of the added git2 dependency, you might need to add 'default-features = false' which will disable some optional features of git2 (amongst them is ssh feature, which is the part that makes it harder to build for macOS), you can check more in git2-rs repository readme [here](https://github.com/alexcrichton/git2-rs#building-on-osx-1010).

This is one of the reason why I chose to call git command directly in #12. But if this is merged, we can change how authors are detected too.

Also will the repo url get correctly recognized if user tracks upstream or origin with ssh instead of https?

JoshBrudnak:
Yes, it will work with ssh and https. Thanks.

o2sh:
Good job @JoshBrudnak, but the way you detect the repo name relies on the presence of `.git` in the url. So, if I choose to clone a repo without it:

`git clone https://github.com/o2sh/onefetch`

It won't work...

JoshBrudnak:
Good point, In that case I'll assume the repository name is the last part of the url. :+1:

o2sh:
Thanks a lot @JoshBrudnak, @zxey what do we do for the author detection ? do we maintain what we have, or do we use git2-rs ?

We're almost there, only the License is left hardcoded.

PR #15, opened 2018-10-30
Add Lua ascii logo

francesco-dipi:
#2


o2sh:
Thanks @francesco-dipi

PR #18, opened 2018-12-11
Error types instead of process:exit() for main and get_color() moved to impl fn of Info

xynxynxyn:
I implemented a custom error type and proper error handling since main can return Result.

I also moved the `get_color(&language)` function to be an implemented function on Info such that the color of a project/language can be retrieved by calling `Info.color()`.


o2sh:
Great refactoring @xynxynxyn 
Clean code In Rust is not easy to achieve...

PR #20, opened 2018-12-14
Multi color ascii

xynxynxyn:
Implemented multi color ascii.
To color ascii just add {n} to the ascii to change every character after it to that color. N can be any single digit number, so only up to 10 different colors per ascii.

I only changed the C and rust logo to multiple colors so far, but the others still work even if they don't have the annotations.


o2sh:
Impressive work @xynxynxyn 
We are getting very close to what Neofetch can do...in Rust.

Thanks again

PR #29, opened 2019-04-10
Separate build and install targets

hoop33:
I don't have root on my work machine, and expected to be able to run `make` just to build the app. I was a little startled to see a `sudo` prompt when I just ran `make`.

Thanks for this utility!


o2sh:
Oups, sorry about that.
Thx a lot @hoop33

PR #30, opened 2019-04-10
Add assembly detection and ascii image

aeter:
Adds assembly detection and ascii image (https://github.com/o2sh/onefetch/issues/2)

The ascii is created through  
```
$ figlet -f 3D%20Diagonal.flf "asm"
```

asm_ascii


o2sh:
Looks great !
Thx @aeter

PR #35, opened 2019-06-18
Add Forth language

jephthai:
Add config for Forth language (already supported by Tokei), with ASCII art inspired by the most famous Forth language cover art in history (should be recognizable to most Forthers).


o2sh:
Thanks for your contribution @jephthai,
Too bad you didn't use multiple colors to make it look better.

Still, this is really cool!!

PR #42, opened 2019-10-01
Use clap to handle command-line arguments

### LINKED ISSUE #40: Add -h command-line option for usage
Executing `onefetch -h` causes onefetch to treat `-h` as a directory.
Since [clap](https://crates.io/crates/clap) is already needed to compile due to your tokei dependency, you might want to make use of clap so that `onefetch -h` would print a help message.

spenserblack:
`onefetch -h` will provide a default usage message (also adds `onefetch -V` option to print version). Should fix #40.

I've changed the usage `onefetch path/to/dir/` to be `onefetch -d path/to/dir/`, with the default dir being `.` if that option isn't specified. The reason I did this is because I think optional arguments should be preceded by an *option*, but I'm happy to revert the usage to `onefetch path/to/dir/` if you prefer it that way :smiley:.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,4 @@ git2 = {version = "0.7.5", default-features = false}
 tokei = "8.0"
 license = "0.7.1"
 bytecount = "0.5.1"
+clap = "2.33.0"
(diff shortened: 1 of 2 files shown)


o2sh:
Great work @spenserblack 

You may have missed some warnings though (tip: cargo clippy):

DeepinScreenshot_select-area_20191001200044

Please take care of it, so that I can merge your PR

spenserblack:
No problem, @o2sh!

PR #46, opened 2019-10-02
Add editorconfig

spenserblack:
Hello! While working on an issue, I noticed a few irregularities in the code (tab indentation mixed with space indentation, some trailing whitespace). I thought you might want to add an [`.editorconfig`](https://editorconfig.org/) file, so that contributor's IDEs and text editors would behave in a more uniform way.

Based on the files I've looked at, I think this should cover how you want the files in your project to be formatted. But it's your project, @o2sh, so please suggest any additions/changes you might want!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = LF
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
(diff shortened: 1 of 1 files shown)


spenserblack:
I just realized a lot (but not all) of the `.ascii` files have CRLF (Windows) line endings. Converting to LF (Unix) can result in some git diffs marking the whole file as changed. Not sure if you'd want `.ascii` files converted to LF at the expense of large diffs or not.

PR #52, opened 2019-10-04
Adding OCAML ascii art

Vinetos:
Check if I have updated `main.rs` correctly !


o2sh:
Thanks for your contribution @Vinetos 
I will check the main.rs in a bit.

However, the ascii art is way too big and "noisey"...
Plz, try and improve on it 👍 

After that, I will gladly merge the PR.

Vinetos:
Is this better ? @o2sh

o2sh:
I apologize for the late answer @Vinetos 
It's indeed better for the "noisey" part, unfortunately it's still too big.

You're almost there :1st_place_medal:

Vinetos:
Gosh !
What's is a good size ? Because if I reduce too much, we cannot see anything ! 😂

Issue #60, opened 2019-10-06
Display the creation date of the repo


o2sh:
We would like to display the repo's age.

That information should be displayed along side the "Project" name, as such:

Project: onefetch, created 1 year, 1 month ago

You can use `git log --reverse --pretty=oneline --format="%ar"` to fetch the repository age from oldest commit

:100: :1st_place_medal:

bojan88:
Hey, I'll create a PR for this. Already working on it

PR #61, opened 2019-10-06
Project creation date

bojan88:
Implements #60.

My code editor likes to remove trailing spaces. I hope you don't mind, but if you do, I can revert those changes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -182,9 +182,9 @@ fn count_newlines(s: &str) -> usize {
 /// escape characters for color display.
 ///
 /// Colors are specified with {0}, {1}... where the number represents
-/// the nth element in the colors Vec provided to the function.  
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -666,6 +670,23 @@ fn get_total_loc(languages: &tokei::Languages) -> usize {
...
+        .arg("--pretty=oneline")
+        .arg("--format=\"%ar\"")
+        .output()
+        .expect("Failed to execute git.");
+
+    let output = String::from_utf8_lossy(&output.stdout);
+
+    match output.lines().next() {
+        Some(val) => Some(val.to_string().replace('"', "")),
+        None => None
```
bojan88: We might not even need to handle `None` since the program will error and exit before this in case there are no commits

PR #62, opened 2019-10-07
Add Zig support

SamTebbs33:
This patch adds support for [Zig](https://github.com/ziglang/zig). Unfortunately colored doesn't support orange.

onefetch with Zig


o2sh:
Nice to see you took advantage of the multicolor feature for the ascii logo.
Unfortunately, it looks a little noisey and unpolished.

You're almost there.

SamTebbs33:
Thanks for the feedback @o2sh . I agree that it looked unpolished, so I chose to replace all the white symbols with Zs. Let me know if it's more suitable or if I should go back to the style with varied characters.

v2

o2sh:
Better! @SamTebbs33 
Try and replace the "Z"s with blank spaces and the "/" with a more visble character (#, $...)

SamTebbs33:
Much better :)
v3

PR #65, opened 2019-10-08
Exclude images from crate

spenserblack:
Next time this crate is published to crates.io, this should make the crate around 26KB, instead of 312KB (checked with `cargo package && ls -lh target/package/`).


o2sh:
Well done!

PR #67, opened 2019-10-10
Use raw images for screenshots

spenserblack:
Should fix images not rendering on [crates.io page](https://crates.io/crates/onefetch).
Looks like crates.io strips HTML alignment attributes, so it will still be left-aligned, unfortunately.
docs.rs Also seems to strip all HTML from the README.


o2sh:
Thanks for that @spenserblack 
I will publish a new version to crate later today to see how it renders.

Issue #45, opened 2019-10-01
Better and more ASCII !! (╯°□°）╯︵ ┻━┻ 


o2sh:
Some of our ascii logos look a bit out of shape...

plz help ಥ_ಥ

Don't hesitate to add more languages, if you feel inspired :100:

PR #68, opened 2019-10-11
Add HTML/CSS language support

MaxJohansen:
Helping out with #45!

I've run the HTML and CSS logos through an ASCII art generator and touched them up by hand.

CSS

HTML


o2sh:
Nice to see you took advantage of the multicolor feature for the ascii logo.
Unfortunately, it looks a little noisey/unpolished and also too big.

You're almost there @MaxJohansen 

MaxJohansen:
I've shrunk the logos down to 21 lines, which is more in line with the other ones :-)

Also made them more uniform to avoid the business of more granular ASCII art, what do you think?

css-squished
html-squished

PR #71, opened 2019-10-12
Add Vue.js language support

estevam31:
Helping with #45


o2sh:
Great work @estevam31, I didn't know Vue.js was considered a language on its own.

**You're missing a coma line 301.**

Plz fix so that I can merge your PR :+1:

estevam31:
In theory it would not be a language, but in github is considered as one. I believe it is because the files have the extension .vue

Just sent the fix to that missing comma, thanks for your time.

Issue #74, opened 2019-10-12
Choose Which Language's ascii art to print


o2sh:
We would like to allow users to pick which language's ascii art to print.
By default, the ascii art corresponds to the main language of the repo.

With this feature, the user can choose which art to display:

`onefetch --ascii_language rust`

which will show the rust logo even if the main language isn't rust.

:100:

ktsuench:
I'm interested it trying this out :)

PR #75, opened 2019-10-13
Feature: allow user to pick ASCII art with flag --ascii_language

ktsuench:
I implemented the feature using the package `Strum` to work with the `Language` enum. Makes it easier to go from string to `Language` enum. Also, I decided to convert the value from CLI to lowercase and turn all the `Language` enum values to lowercase for easier comparison plus ease of use.

Use:
Overrides showing the dominant language ascii logo
`-a, --ascii_language <ascii_language> [default: ]`

Examples:
1. Flag used with no value provided, defaults to dominant language ASCII logo

image

2. Flag used with unrecognized language, defaults to dominant language ASCII logo

image

3. Flag used with recognized language, overrides dominant language ASCII logo

image

4. Flag used with mixed case recognized language, override dominant language ASCII logo

image

Addresses issue #74


o2sh:
Great work @ktsuench 
And the way you treated the corner cases is just splendid!

Thanks again

PR #76, opened 2019-10-13
Add ability to disable info elements

### LINKED ISSUE #73: Disable info line from appearing in the output
We would like to allow users  to disable one (or many) info line(s) from appearing in the output via a command line flag:

`onefetch --disable size version`

:100:

mstruebing:
closes #73

I used `Options` to indicate if a information part is disabled or not.

Screenshot_2019-10-13_12-17-14
Screenshot_2019-10-13_12-14-50
Screenshot_2019-10-13_12-14-30

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,11 +9,11 @@ extern crate strum;
 #[macro_use]
 extern crate strum_macros;
 
+use clap::{App, Arg};
 use colored::Color;
 use colored::*;
-use git2::{Repository, Oid};
+use git2::{Oid, Repository};
 use license::License;
-use clap::{App, Arg};
 use std::{
     cmp,
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


o2sh:
Impressive stuff @mstruebing 

Few remarks :
* The info name provided shouldn't  be case sensitive (ex: `onefetch --disable sIzE` should work) --> you can check #75
* The enum of possible values should match with the one displayed (current_commit <> head, project_name <> project...), this can be confusing. By doing so, you won't have to specify the list of possible values in the help of the command.
* Don't  forget to add a little help to describe what the flag does, something like -->  Allows  you  to disable an info line from appearing in the output.

mstruebing:
Still to do: mixed case or uppercase arguments are counted as failures by clap.

Some arguments need clarification

How should I name:

`last_change`
`lines_of_code`
`dominant_language` 

I don't think that spaces are possible or a good idea. Looking for your feedback :)

o2sh:
You're right @mstruebing 

This might not be optimal but here is a suggestion.
You can change the display names as such:
* Lines of code --> LOC
* Last change --> Last
* dominant_language --> language

In the same spirit, if there is only one author. The display name is gonna be just "author" --> it should be added as a possible value for the flag.

:100:

mstruebing:
I think I've respected every change request :)

o2sh:
when I do `onefetch --disable language`, I get this:

DeepinScreenshot_select-area_20191013225109

indeed, `Info:language` is used in get_ascii() to return the corresponding asset --> you cannot assign it to None

mstruebing:
Thanks for you catchy eye, will fix sonnish :)

o2sh:
I've decided to merge #82, hope you don't mind... :pray:

PR #80, opened 2019-10-13
WIP: Split code into modules

### LINKED ISSUE #77: [Code organization] Split main.rs into modules
As suggested by @mstruebing in #70, 
It's time to split the main.rs file into multiple files/modules.

Feel free to re-organize the code in a way that improves the navigation inside the project overall.

Thx :100: :1st_place_medal:

mstruebing:
closes #77


o2sh:
Hi @mstruebing,

Can I review? Or is it still "WIP" ?

Please use snake case (commit_info.rs) for the filenames.

Thx again :+1:

mstruebing:
It's still WIP and has some errors I didn't had time to fix until now.

o2sh:
Don't hesitate to ask for help if need be :smile_cat:

src/language.rs
```diff
@@ -0,0 +1,89 @@
+extern crate strum;
+
+use std::fmt;
+
+#[derive(PartialEq, Eq, Hash, Clone, EnumString)]
```
mstruebing: i.e: `cannot find derive macro `EnumString` in this scope`

src/language.rs
```diff
@@ -0,0 +1,89 @@
...
+#[strum(serialize_all = "lowercase")]
```
mstruebing: i.e: `The attribute `strum` is currently unknown to the compiler and may have meaning added to it in the future. for more information, see https://github.com/rust-lang/rust/issues/29642`

spenserblack:
Looks like you need to add this to bring strum's macros into scope.
```rust
#[macro_use]
extern crate strum_macros;
```

mstruebing:
When I do so inside the `language.rs`-file it tells me: `an `extern crate` loading macros must be at the crate root`

But it is already at the root like you can see in the code.

spenserblack:
Got it. Sorry, I'm more used to writing in 2018 edition, where `#[macro_use]` isn't necessary.
I think the issue is that the macro is in the scope of `main.rs`, but not `lib.rs`, where `mod language` is used. So try moving the strum macros to the root of `lib.rs`.

mstruebing:
That worked, there are two small issues left. I comment the code, even if there is already a new PR I would love to know how to fix this.

src/lib.rs
```diff
@@ -0,0 +1,420 @@
...
+                entry.is_file()
+                    && entry
+                        .file_name()
+                        .map(OsStr::to_string_lossy)
+                        .unwrap_or_else(|| "".into())
+                        .starts_with("LICENSE")
+            }, // TODO: multiple prefixes, like COPYING?
+        )
+        .map(|entry| {
+            license::Kind::from_str(&fs::read_to_string(entry).unwrap_or_else(|_| "".into()))
```
mstruebing:
```
no variant or associated item named `from_str` found for type `license::kind::Kind` in the current scope: variant or associated item not found in `license::kind::Kind`
```

src/lib.rs
```diff
@@ -0,0 +1,420 @@
+extern crate colored;
+extern crate git2;
+extern crate strum;
+extern crate strum_macros;
+
+use colored::*;
```
mstruebing:
```
the following trait is implemented but not in scope, perhaps add a `use` for it:
```

src/lib.rs
spenserblack: `Kind::from_str` is implemented by the `std::str::FromStr` trait. If `FromStr` is not in scope, then foreign types will not have functions that the trait requires. So try adding `use::std::FromStr`.

Just curious, what version of Cargo are you using? It's usually smart enough to know which available traits you need to bring into scope to make a function available.

spenserblack:
I'm happy to continue this discussion to help you fix this, but I'll be closing this PR, since #77 and #97 are already closed and #98 has been merged. Thanks for your hard work, though :smiley:

PR #84, opened 2019-10-14
added xml ascii art

tacrazymage:


src/main.rs
```diff
@@ -1083,6 +1088,7 @@ impl Info {
             Language::Vue => vec![Color::BrightGreen, Color::Blue],
             Language::Perl => vec![Color::BrightBlue],
             Language::Php => vec![Color::BrightWhite],
+            Language::XML => vec![Color::Yellow],
```
o2sh: You seem to be using two colors in the ascii logo but you're only referencing a single value in the vector, it still works though

o2sh:
Plz check my comment.

tacrazymage:
Ooops, sorry for that :(
try one more time

PR #88, opened 2019-10-16
Fix compilation error on linux

spenserblack:
I was getting a compilation error when attempting to compile on Linux.
I tested this on both Windows and Linux, compiles on both. Will probably fail on MacOS, though.

Thanks for adding me as a collaborator, @o2sh!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,4 +17,6 @@ bytecount = "0.5.1"
 clap = "2.33.0"
 strum = "0.16.0"
 strum_macros = "0.16.0"
+
+[target.'cfg(windows)'.dependencies]
 ansi_term = "0.12"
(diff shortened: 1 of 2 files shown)


o2sh:
You can check [this](https://github.com/rivy/rust.tin-summer/commit/a8c8bd9c5d246157e75d30b3d3ab910a1e79bcce)

This will only apply to windows

spenserblack:
Thanks! I used a combination of the example you provided and `ansi_term`'s documentation to add this.

spenserblack:
Do you mind if I update `.travis.yml` to compile on both Linux and Windows (and possibly OSX)?

o2sh:
> Do you mind if I update `.travis.yml` to compile on both Linux and Windows (and possibly OSX)?

Sure 👍

spenserblack:
Travis CI is currently pointing to a cancelled build due to a force push. But [this build](https://travis-ci.org/o2sh/onefetch/builds/599191962) should effectively be the same.

o2sh:
Sure go ahead, it  should fix #90

PR #92, opened 2019-10-18
Add Tex support

KaindlJulian:
Screenshot from 2019-10-18 22-38-08


o2sh:
Looks great,
Thanks @KaindlJulian

Issue #106, opened 2019-10-23
Display git version + username


o2sh:
On top of the repository infos, we would like to show general information (not specific to the repo), like git username (git config user.name) and the version of Git installed on the machine (git --version). Something like -->

![Untitled](https://user-images.githubusercontent.com/13710835/67389745-7c8f1a80-f59b-11e9-943c-bf115685469c.png)

💯 🥇

pablodiegoss:
I've tried to replicate your prototype on #107, what do you think?
![Screenshot from 2019-10-23 11-12-51](https://user-images.githubusercontent.com/8556291/67401847-177cfa00-f586-11e9-8b55-f0927c6c1be8.png)

o2sh:
Great start @pablodiegoss 

Few remarks:
* The dotted line has to match the length of the displayed info, in this case: _Pablo Diego: git version_ 2.20.1
* You have to handle the case where the git username is unknown --> Don't print ":" at the start

Looking forward to merge your PR 👍

pablodiegoss:
Nice!
I'll work on those remarks then :)

pablodiegoss:
After formatting the line and username case it looks like this

With username:
![Screenshot from 2019-10-23 12-17-33](https://user-images.githubusercontent.com/8556291/67408432-56fc1400-f58f-11e9-805a-a5d1242bfa47.png)

And without:
![Screenshot from 2019-10-23 12-18-30](https://user-images.githubusercontent.com/8556291/67408440-59f70480-f58f-11e9-93d0-43015470a043.png)

o2sh:
Looking better @pablodiegoss 

Last remarks:

* The text has to be diplayed in bold (except the dotted line), you should check #99 (already merged) to make sure that we can disable boldness on your text too
* We can't see it in your example, but the text should be displayed in color (like the info titles); by text I don't include **the dotted line and the separator** --> only git username and git version

🥇

pablodiegoss:
I've updated the color and bold o/
![Screenshot from 2019-10-24 12-22-34](https://user-images.githubusercontent.com/8556291/67500426-05b85700-f659-11e9-874c-84ef3269e64a.png)

pablodiegoss:
I'm not quite sure how to do that, because the `write_buf` function that prints and paint texts breaks a line after printing something. So if i leave the ":" as white the next printed text with the git version will be in a new line. If i change the `write_buf`to not break lines, all the code should be changed to add line breaks.
[...]

PR #107, opened 2019-10-23
Add Git info for user and version

pablodiegoss:
Resolves: #106
I'm new to rust, but i think i've solved the issue or at least contributed to it :D

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -43,17 +43,15 @@ impl std::fmt::Display for Info {
             None => Color::White,
         };
         if !self.disable_fields.git_info{
-            let username = if self.git_username == "" {
-                String::from("")
-            }else{
-                format!("{}:",self.git_username)        
-            };
-
-            let git_info = format!("{}{}", username, self.git_version);
-            write_buf(&mut buf, "", &git_info, color)?;
-
+            let git_info;
+            if self.git_username != "" {
+                git_info = format!("{} : {}", self.git_username, self.git_version);   
+            } else {
+                git_info = self.git_version.clone();
+            }
+            write_buf(&mut buf, &self.get_formatted_info_label(&git_info, color), "")?;
             let separator = "-".repeat(git_info.len());
-            write_buf(&mut buf, "", &separator, color)?;
+            write_buf(&mut buf, &self.get_formatted_info_label("", color), &separator)?;
         }
         if !self.disable_fields.project {
             write_buf(&mut buf, &self.get_formatted_info_label("Project: ", color), &self.project_name)?;
--- a/src/info.rs
+++ b/src/info.rs
@@ -45,11 +45,12 @@ impl std::fmt::Display for Info {
         if !self.disable_fields.git_info{
             let git_info;
             if self.git_username != "" {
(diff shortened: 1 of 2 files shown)


src/info.rs
```diff
@@ -262,6 +273,25 @@ impl Info {
...
+            .arg("--version")
+            .output()
+            .expect("Failed to execute git.");
+        let version = String::from_utf8_lossy(&version.stdout);
+        let version = version.to_string().replace('\n',"");
+
+        let username = Command::new("git")
+            .arg("config")
+            .arg("--get")
+            .arg("user.name")
```
spenserblack: Good job!
But it looks like this will get `user.name` in the *current directory*, not the directory that could get passed by `onefetch -d path/to/project`. This could be an issue when the local `user.name` is different from the global `user.name`.
See `get_authors` for an example of running a git command for the specified directory.

pablodiegoss: Done, i've tested with the -d flag and seems fine now, can you verify?

spenserblack: Looks good to me :+1:

src/info.rs
```diff
@@ -39,7 +41,19 @@ impl std::fmt::Display for Info {
             Some(&c) => c,
             None => Color::White,
         };
+        if !self.disable_fields.git_info{
+            let username = if self.git_username == "" {
+                String::from("")
+            }else{
+                String::from(format!("{}:",self.git_username))
```
spenserblack: `format!` already returns type `String`, so you don't need to call `String::from` here. It's kind of like calling `String::from(String::from("stuff"))`.

src/info.rs
```diff
@@ -262,6 +279,27 @@ impl Info {
...
+        let username = Command::new("git")
+            .arg("-C")
+            .arg(dir)
+            .arg("config")
+            .arg("--get")
+            .arg("user.name")
+            .output()
+            .expect("Failed to execute git.");
+        let username = String::from_utf8_lossy(&username.stdout);
+        let username = username.to_string().replace('\n',"");
```
spenserblack: Even though the return type of `from_utf8_lossy` is `Cow<str>`, you can treat it as if it was a `&str`, so you don't need `to_string`.
See [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%20%7B%0A%20%20%20%20println!(%0A%20%20%20%20%20%20%20%20%22Using%20%60replace%60%20without%20%60to_string%60!%20%7B%3A%3F%7D%22%2C%0A%20%20%20%20%20%20%20%20String%3A%3Afrom_utf8_lossy(%26%5B100%2C%20121%5D).replace(%27d%27%2C%20%22x%22)%2C%0A%20%20%20%20)%3B%0A%7D).

src/info.rs
pablodiegoss: Oh, I didn't know that!
I've removed that redundancy.

src/info.rs
pablodiegoss: Thank you for that example, I've fixed both places that I did that.

o2sh:
@spenserblack Don't hesitate to merge this one when your code review is done.

PR #111, opened 2019-10-24
Optimize images

spenserblack:
Reduced filesize of assets.

See spenserblack#1 for stats


o2sh:
Great idea @spenserblack 

I plugged the Bot to the repo, this should be done automatically from now on

PR #113, opened 2019-10-26
Add support for displaying a custom image instead of ascii art

CephalonRho:
Add support for displaying an image specified with the --image option using the kitty terminal graphics protocol.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,11 @@ bytecount = "0.5.1"
 clap = "2.33.0"
 strum = "0.16.0"
 strum_macros = "0.16.0"
+image = "0.22.3"
 
 [target.'cfg(windows)'.dependencies]
 ansi_term = "0.12"
+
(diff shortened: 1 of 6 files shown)


o2sh:
DeepinScreenshot_select-area_20191027153151

o2sh:
Impressive stuff @CephalonRho 
This is way over my head, so I'm not even gonna try doing a code review on this one...(@spenserblack || @andymac-2 any volonteer?).

> This works pretty well, but has very limited support (right now only kitty itself implements that protocol) so I also added a basic trait to have multiple image backends for different terminal emulators.

That's the best way to do it _**program to interfaces, not implementations**_,  thus paving the way for more support in the future like [catimg](https://github.com/posva/catimg), [SIXEL](https://github.com/saitoha/libsixel), w3m...

spenserblack:
Looks great to me! :+1: 
Thanks for structuring this so that more backends can be supported in the future

PR #114, opened 2019-10-26
Limit shown languages to 6

### LINKED ISSUE #112: Limt the number of language statistics
For repositories using multiple languages, like this [one](https://github.com/dart-lang/sdk):

DeepinScreenshot_select-area_20191026152815

We would like to limit the number of `language_stat` to N (6 for example), like GH does: 

DeepinScreenshot_select-area_20191026153248

Which means adding a seventh `language_stat` named _Other_ with the cumulated % of the remaining languages.

:100: :1st_place_medal:

CephalonRho:
Closes #112

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -78,9 +78,10 @@ impl std::fmt::Display for Info {
                 let mut languages: Vec<(String, f64)> = {
                     let mut iter = self.languages.iter().map(|x| (format!("{}", x.0), x.1));
                     if self.languages.len() > 6 {
-                        let first_languages = iter.by_ref().take(6).collect::<Vec<_>>();
+                        let mut languages = iter.by_ref().take(6).collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


o2sh:
Great job @CephalonRho ,
However the _Other_ stat should always be printed at last, which is not the case when the cumulated percentage of remaining languages is not the smallest:

DeepinScreenshot_select-area_20191026234855

Looking forward to merge your PR :+1:

CephalonRho:
I removed the second sort, now it should always be printed last.
image
I first added it because I thought that it would represent repositories with a large variety of languages better, but this is fine too.

o2sh:
DeepinScreenshot_select-area_20191027125058

Other than that, LGTM :1st_place_medal:

CephalonRho:
Sorry, missed the warning. Should be fine now.

PR #126, opened 2019-10-29
Added instructions to README for use on Windows.

gurry:
It works just fine on Windows. All it needed is some instructions on how to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,20 +16,28 @@ Onefetch is a command line tool that displays information about your Git project
 <img src="https://raw.githubusercontent.com/o2sh/onefetch/master/assets/kitty.png" height="217px" width="420px">
 </p>
 
-## Usage
+## Installation
+
+### Unix
 
 Clone the repository and install to `/usr/local/bin` with `make install`.
 
 Or get the binary from [here](https://github.com/o2sh/onefetch/releases) and add it to your $PATH.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,20 +16,28 @@ Onefetch is a command line tool that displays information about your Git project
...
+
+### Unix
 
 Clone the repository and install to `/usr/local/bin` with `make install`.
 
 Or get the binary from [here](https://github.com/o2sh/onefetch/releases) and add it to your $PATH.
 
+### Windows
+
+Clone the repository and build with `cargo build --release`. Add the resulting binary to your $PATH.
```
spenserblack: Thanks for your contribution!

`cargo install --path path\to\onefetch\repo` will build this project and add it to the path. Might be a bit easier to use.

gurry: That's a good suggestion. Didn't know you could use `cargo install` on a local path. Thanks :). I've updated the README. 

One thing I was wondering about was maybe we can use the same installation instructions for Unix. The only reason we seem to be using `make` is because it installs the binary in `/usr/bin/local`. Why not let `cargo install` do its thing and install it in its own `bin` directory?

spenserblack: Yes, I think that's a good idea. That way it's the same instructions for all operating systems.

Issue #129, opened 2019-10-29
Run Onefetch in any of the repo's subfolders


o2sh:
As suggested by @davep in #100 , It would be handy if the tool could work anywhere within a git repo, rather than only in the root of the repo.

The idea would be to search for the `.git` folder in any of the parent directories as a marker for the root of the repo and to execute onefetch on it 👍

If it fails to find it, the error message should be:

`fatal: not a git repository (or any of the parent directories): .git`

rockisch:
Seems like the `git2` package supports this kind of usage with the `discover` method:
https://docs.rs/git2/0.10.1/git2/struct.Repository.html#method.discover

I'm opening a PR for that, although not sure I should use the error message on the description, since that's not how the other error messages are formatted.

spenserblack:
Closed by #132
Thanks, @rockisch!

o2sh:
The function `get_project_license` doesn't work in subfolders

rockisch:
Damn, I ran a little test but didn't check the individual fields, my bad.

I think `git2::Repository` has some kind of method to get the root dir of the repo. I can check later when I'm home.

davep:
Updated locally and it's working great. Just wanted to say thanks for doing this. Very much appreciated.

PR #98, opened 2019-10-20
Refactor modules into files

andymac-2:
Fixes #97 

The program is split into several modules, the largest (the info module) is already logically quite simple, and it would not make much sense to split it any further in my opinion.

I may have broken something, but as far as I can tell through my testing, this refactor went quite smoothly.

It might be good to further refactor the `Info::new` method to have less arguments.

I also ran `rustfmt` on the resulting code since I'm changing pretty much every line anyway.

PR #139, opened 2019-10-31
Improve running command from subfolder

rockisch:
Related Issue: https://github.com/o2sh/onefetch/issues/129

So, while doing some investigation, I discovered that whenever the `git` cmd is executed in a subfolder, it does a recursive search for the git repo on the current path.

Although the impact on this program is very low, since we only do ~10 git calls per run, I was already going to extract the working dir to fix the LICENSE issue, so why not improve the other calls?

Also, we can't call `workdir()` on bare git folders, but the program was already not working on them, since they don't have any content.

I can also reorder the functions that receive the `Repository` to keep the code more uniform, but some people don't like big git changes since it messes their git blame ¯\\_(ツ)_/¯

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,6 +11,8 @@ pub enum Error {
     /// Not in a Git Repo
     NotGitRepo,
     /// Error while getting branch info
+    BareGitRepo,
+    /// Repository is a bare git repo
     ReferenceInfoError,
     /// Image probably doesn't exist or has wrong format
     ImageLoadError,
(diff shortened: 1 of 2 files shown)


spenserblack:
>I can also reorder the functions that receive the `Repository` to keep the code more uniform, but some people don't like big git changes since it messes their git blame ¯\(ツ)/¯

We've already had some big changes (#98), so I don't think it's an issue.

rockisch:
I reordered the functions to keep the code more consistent, besides updating the branch to match master.

PR #140, opened 2019-10-31
Fixed several clippy warnings

ZapAnton:
This PR fixes some clippy warnings, in order to reduce the clippy output noise, so only relevant lints would be visible.


o2sh:
Thanks a lot @ZapAnton :+1:

PR #141, opened 2019-10-31
Feature/no merges

### LINKED ISSUE #131: Onefetch counts Merge Commits
Think that it's not ok to count merge commits since they are with no value at all and it makes wrong statistics.

spenserblack:
Resolves #131 
Let me know if I missed anything!


o2sh:
Looks great :+1: , If you could just add a `.short("n")` as a shortcut for `--no-merges`

spenserblack:
:+1:, but are we sure `-n` should be used for this? There's already a "no" flag (`-n`, `--no-bold`). How about `.short("m")`?

o2sh:
You're absolutely right, it's more consistent that way :+1:

spenserblack:
Done :smiley:

PR #142, opened 2019-10-31
Fix language name mismatch

### LINKED ISSUE #125: Language name support mismatch
This command should display each supported ASCII art.
```bash
# Runs `onefetch -a Assembly`, `onefetch -a C`, etc.
onefetch -l | while read line; do echo "$line"; onefetch -a $line; done;
```
Several of these languages will map to unknown, however, causing the default ASCII art to be displayed.
This is because some listed languages have special characters (C++, C#, Objective-C), yet custom ASCII art cannot take special characters.
```bash
# doesn't work
onefetch -a C++

# works
onefetch -a cpp
```

I think it can be a bit confusing if `onefetch -l` lists the language as "C++", yet that's an invalid input for `onefetch -a <ascii_langage>`. They should probably match.

It might also help if the help message showed the possible values for `-a`.

ZapAnton:
Closes #125.

New help message for the `-a` flag:

```text
    -a, --ascii_language <ascii_language>    Overrides showing the dominant language ascii logo. Possible values:
                                             [assembly, c, clojure, coffeescript, cpp, csharp, css, dart, elixir, elm,
                                             erlang, forth, fsharp, go, haskell, html, idris, java, javascript, julia,
                                             kotlin, lisp, lua, nim, objectivec, perl, php, prolog, purescript, python,
                                             r, racket, ruby, rust, scala, shell, swift, tcl, tex, typescript, vue, xml,
                                             zig] [default: ]

```


src/main.rs
```diff
@@ -110,7 +110,14 @@ fn main() -> Result<()> {
...
                 .takes_value(true)
                 .default_value("")
-                .help("Overrides showing the dominant language ascii logo"),
+                .help(&format!(
+                    "Overrides showing the dominant language ascii logo. Possible values: [{}]",
+                    Language::iter()
+                        .filter(|language| *language != Language::Unknown)
+                        .map(|language| language.to_string().to_lowercase())
+                        .collect::<Vec<String>>()
+                        .join(", ")
```
spenserblack: You can actually [set possible values](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.possible_values) for a `clap::Arg`. This will not only show in the help message \(assuming you don't [hide them](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.hide_possible_values)\), but will also make `clap` ensure that *only* the possible values are used. e.g. If someone runs `onefetch -a French`, they will receive an error message saying that's an unexpected input, and prompting them to view the help message.

src/language.rs
```diff
@@ -60,8 +60,8 @@ impl std::fmt::Display for Language {
             Language::C => write!(f, "C"),
             Language::Clojure => write!(f, "Clojure"),
             Language::CoffeeScript => write!(f, "CoffeeScript"),
-            Language::Cpp => write!(f, "C++"),
-            Language::Csharp => write!(f, "C#"),
+            Language::Cpp => write!(f, "Cpp"),
+            Language::Csharp => write!(f, "Csharp"),
```
spenserblack: Unfortunately, I was hoping for the valid *inputs* to be changed to "C++", "C#", etc. \(preferably case-insensitive\), instead of the *outputs* being changed to "Cpp", "Csharp", etc.

ZapAnton: This may be a stupid question, but how does `Language::from_str` works?
There is no explicit implementation of the `FromStr`.
Is it done implicitly via `#[strum(serialize_all = "lowercase")]` or `EnumString`.
The answer would help in resolving the issue with the valid inputs

spenserblack: `Language::from_str` is implemented from [`#[derive(EnumString)]`](https://docs.rs/strum_macros/0.16.0/strum_macros/derive.EnumString.html). Documentation can be found [here](https://github.com/Peternator7/strum/wiki/Derive-EnumString).

Hope this helps!

ZapAnton:
@spenserblack , I have resolved the issues you pointed out, could you give it another look?

src/main.rs
```diff
@@ -109,8 +114,15 @@ fn main() -> Result<()> {
                 .short("a")
                 .long("ascii_language")
                 .takes_value(true)
-                .default_value("")
-                .help("Overrides showing the dominant language ascii logo"),
+                .default_value("rust")
```
spenserblack: Why is the default value `"rust"`? Just ran your branch for a CoffeeScript project, and it incorrectly used Rust's ASCII art.

It might be better if there is no default value at all, actually. The logic can be:
```rust
let ascii_language = if let Some(ascii) = matches.value_of("ascii_language") {
    // Use user-selected ascii art
} else {
  // Calculate ascii art
}
```
Since `matches.value_of("ascii_language").unwrap()` would often panic with no default value.

PR #149, opened 2019-11-03
Replace license crate with askalono

### LINKED ISSUE #147: [License Detection] Switch to askalono
Onefeth currently relies on [license](https://github.com/evenorog/license/blob/master/src/lib.rs), for license detection.

After looking at the code, it seems only capable of detecting -->
```
pub fn from_id_ext(id: &str) -> Option<&'static dyn LicenseExt> {
    match id {
        "AGPL-3.0-only" => Some(&AGPL_3_0_only),
        "Apache-2.0" => Some(&Apache_2_0),
        "CC0-1.0" => Some(&CC0_1_0),
        "GPL-3.0-only" => Some(&GPL_3_0_only),
        "LGPL-3.0-only" => Some(&LGPL_3_0_only),
        "MIT" => Some(&MIT),
        "MPL-2.0" => Some(&MPL_2_0),
        "Unlicense" => Some(&Unlicense),
        _ => None,
    }
}
```

Which is clearly unsufficient...hence #133 #118 and #123 

[Askalono](https://github.com/amzn/askalono) seems to be the right candidate for the job,.

Let's migrate 💯

ebroto:
Resolves #147 

So askalono seems to use a more complex analysis that relies in a binary cache generated from a json file. For the moment I went for adding the cache directly under `resources/`. In askalono-cli the json file is checked out as a submodule from [this repo](https://github.com/spdx/license-list-data) and the cache is generated using `build.rs`.

We could go with a similar approach (maybe using git subtree instead), this way we would get updates of the supported licenses and won't have to update the cache manually. Since the solution is more involved I wanted to ask first what do you think about it.


o2sh:
Just tested it, works like a charm! 💯 

> We could go with a similar approach (maybe using git subtree instead), this way we would get updates of the supported licenses and won't have to update the cache manually.

Just to be sure I understood correctly, you mean adding [spdx](https://github.com/spdx/license-list-data) as a subtree to onefetch in order to automatically cash the licenses (generate a license-cache.bin.gz) at compile time via a build.rs?

Sounds great @ebroto !

ebroto:
Yep, that's it.

Alright, I will work on it soon!

o2sh:
Hi @ebroto, don't you think it would be better to use spdx as a `git submodule` instead, so that we only get references and not a full copy of the repo.

The rest is great!!

ebroto:
Yes, I was hesitating about the two options and went for the subtree because it's easier to manage, but indeed it seems overkill for our needs :+1: 

I will try with submodules instead. BTW thanks for the invitation to collaborate!

o2sh:
When I build from a fresh repo, I'm getting:
`thread 'main' panicked at 'Couldn't create a store from SPDX data. Have submodules been initialized ?: Os { code: 2, kind: NotFound, message: "No such file or directory" }'`

o2sh:
This is definitely the most elegant solution, however I don't think the license cache will change that much.

I will create a branch with your PR (to build the a new cache when changes on spdx occur) and use a static cache on master...

ebroto:
You can also use `git clone --recurse-submodules $REPO`. But yes, that was the kind of extra management I wanted to avoid initially because the user has to do extra steps. Besides of this, I think it was a good call to choose submodules over subtrees because they seem a better fit. From what I've seen, cargo and travis for example should fetch the submodules without extra configuration.

In any case, we can always go back to the first approach and bundle the cache directly in the repository. The main con for this is that we have to remember to regenerate the cache whenever we update askalono as the caches are incompatible between versions. 

So what are others doing? Askalono is currently used by two crates: 
- **askalono-cli**: the CLI that comes with the library, is using the submodules approach.
- [cargo-deny](https://github.com/EmbarkStudios/cargo-deny): they have a pre-generated cache versioned in the repository.

So there is no "better" option here I think. After having tried all the options (xD), I'm thinking that we can simplify and just bundle the cache directly for now and see if it works for us.

What do you think?

ebroto:
Hey it seems we posted at the same time :) Did not think about the possibility you suggested, works for me!

o2sh:
Great, thanks again for your PR @ebroto,
---
If you have any suggestion/idea to improve the project, don't hesitate to keep contributing :1st_place_medal:

PR #150, opened 2019-11-05
Added support for the OCaml language

KasraF:
Hey, I'm working with a lot of OCaml code at the moment, and since it was supported by tokei, I added support for it here too.


o2sh:
LGTM 👍 
Thanks for your contribution @KasraF

PR #152, opened 2019-11-06
Use a newer version of askalono so ISC is detected

### LINKED ISSUE #123: ISC license not recognized
It seems that the ISC license is not recognized. I saw comments about upgrading the `license` dependency, maybe that would fix it.

ebroto:
It's been some time without a formal release of askalono. Latest is 0.3.0 (Sep 27, 2018), but some changes have been made to the master branch. In order to support ISC, we have to use a newer version which sadly is not available from crates.io.

In addition I've added some code to remove duplicates if multiple files refer to exactly the same license.

Fixes #123


ebroto:
BTW license cache had to be updated because the format has changed. In order to avoid having to maintain the branch `dynamic-license-cache`, I would suggest using `askalono-cli` to generate it when necessary.

o2sh:
Yes, it's indeed quite unfortunate they didn't release a new version for more than a year especially if they fixed bugs since then...

Thanks a lot for your fix @ebroto

o2sh:
Well, turns out I can't `cargo publish` with an un-released dependancy...

image

ebroto:
Damn, I was not aware of this limitation!

I've opened an issue in amzn/askalono#47 and it seems we will have a new version soon :tada:

jpeddicord:
Sorry for the huuuuge delta between releases -- 0.4.0 is up now, and I'm going to try to pay attention to making smaller, faster releases.

0.4.0 has some minor breaking changes that include dropping Results where unnecessary and removing a deprecated field. Shouldn't be anything difficult to update to, but I'm happy to help if you need it.

PR #154, opened 2019-11-08
Add sixel backend

### LINKED ISSUE #121: Extend Image backend support to SIXEL
As of today we can only display images using the kitty terminal graphics protocol #104 

The next step would be to add support to [SIXEL](https://github.com/saitoha/libsixel)

A Sixel encoder/decoder aleady exist for Rust --> https://github.com/meh/rust-sixel which can be used as a starting point.

----
For the brave only 💯 🥇

CephalonRho:
Closes #121 
This adds decent support for sixel graphics. I only tested it on xterm, but it should work on other terminal emulators too. There seems to be a 16 color limit, so dithering is used to make them look somewhat okay. The kitty backend also needed a fix because the thread workaround caused it to read one more byte from stdin. It wasn't fatal, but would've caused weird behaviour later on, so I replaced it with posix calls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -734,7 +734,7 @@ dependencies = [
 
 [[package]]
 name = "onefetch"
-version = "2.0.0"
+version = "2.0.1"
 dependencies = [
  "ansi_term 0.12.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "askalono 0.3.0 (git+https://github.com/amzn/askalono.git?rev=39f6e228775e17266bcba46f48b26b01bbfb7698)",
--- a/src/image_backends/kitty.rs
+++ b/src/image_backends/kitty.rs
@@ -1,11 +1,9 @@
 use image::{imageops::FilterType, DynamicImage, GenericImageView};
 use libc::{
-    ioctl, tcgetattr, tcsetattr, termios, winsize, ECHO, ICANON, STDIN_FILENO, STDOUT_FILENO,
-    TCSANOW, TIOCGWINSZ,
+    c_void, ioctl, poll, pollfd, read, tcgetattr, tcsetattr, termios, winsize, ECHO, ICANON,
+    POLLIN, STDIN_FILENO, STDOUT_FILENO, TCSANOW, TIOCGWINSZ,
 };
-use std::io::Read;
-use std::sync::mpsc::{self, TryRecvError};
-use std::time::Duration;
+use std::time::Instant;
 
 pub struct KittyBackend {}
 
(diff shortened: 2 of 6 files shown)


o2sh:
@CephalonRho Don't you think, it would be better to let the user choose which `image_backend` to use instead of having a fallback to the first one that is supported by the terminal?

---
Still haven't tested it, but I'm sure it works great!!

o2sh:
mlterm (version 3.8.6) >>
image

xterm(330) >>
image

o2sh:
Still doesn't work on my xterm, which version do you use?

Beside, I don't think that the two step process of first setting the image backend and then being able to use it is the best way to do it.

Let's make it simpler:

```
onefetch --sixel ./my-picture.png
onefetch --kitty ./my-picture.png
```
and no fallback (`get_best_backend`), if the specified backend isn't supported, just show the regular output (ascii_logo + git_infos), or just the git_info (right column) and no ascii_logo.

CephalonRho:
I'm using xterm(350). It should be launched with "xterm -ti vt340" to enable sixel support, but if it's not enabled then it should've panicked unless --image-backend was specified.

Not sure if requiring the user to know which backend their terminal emulator supports is a great idea. We can detect the backend that will definitely work best on the used emulator automatically, so I don't see a reason to not do that. Right now there shouldn't be a reason to use the --image-backend argument other than the detection failing (for example on yaft), so it wouldn't be a two step process most of the time.

o2sh:
> It should be launched with "xterm -ti vt340" to enable sixel support

That dit it, thx!
> it should've panicked unless --image-backend was specified.

No, it didn't panicked, just a funky rendering (see screenshot above)

---

Ok, for the automatic `image_backend` detection. 

---

By the way, I don't think the `image-backend` flag works, when I'm on a Kitty terminal and set the image backend to sixel, it stills render the image using kitty... I'm pretty sure the value is lost during two execution.

I feel like this is the kind of setting that should be put into a static file and read at compile time/Run time, not a command line flag.

If we stick to the automatic detection, you can rollback to the first version, unless I'm missing something

CephalonRho:
I forgot to include the part that actually detects sixel support when removing the thread. That's hopefully fixed now.

---

The `image-backend` flag works fine for me.
image
It prints lots of garbage to the screen, but that's expected when using the sixel backend on a terminal emulator that doesn't support sixel.
The flag does not persist across executions, and unless I missed something there's no config file to implement that with. It would probably be weird for scripts to use anyway.

o2sh:
LGTM :+1: 

Thx a lot,

Looking forward to your next contribution @CephalonRho

PR #155, opened 2019-11-09
Added org-mode support

Phundrak:
I modified `src/language.rs` in order to support Org-mode language, which is already supported by tokei. I also added `resources/org.ascii` for the Org-mode ascii-art.

I tested these modifications against [this org project](https://labs.phundrak.com/phundrak/langue-phundrak-com), and it seems to work. `cargo test` doesn’t return any error either. However, it seems there is a license detection issue, but it was already there before my commits (I will open an issue for that).


ebroto:
Hey, thanks for the PR!

I'll take a look later but about the license problem, I think it will only work if you include the body of the licenses in dedicated files instead of links to them.

PR #158, opened 2019-11-10
Added support for CMake, Emacs Lisp, Fish and Nix

Phundrak:
These changes are basically the same as my previous PR (#155), but this time I added language support for CMake, Emacs Lisp, Fish and Nix.

I also fixed the display for Org, which was displayed as “Org-mode” instead. Each of these additions were tested against git repos mainly relying on one of these languages.
The only issue I have is with the Nix logo, it seems my terminal emulator ([Luke Smith’s build of st](https://github.com/LukeSmithxyz/st)) won’t display any differently `Color::Blue` and `Color::BrightBlue`, but my guess is this is an issue with colored.


Phundrak:
Oh, right, I completely forgot I added this line, it is now removed.
Oh, really? Happy to help! 👍

PR #160, opened 2019-11-10
Automate Homebrew release

erikgaal:
This PR aims to add a GitHub Action that automatically creates a PR in the `homebrew/homebrew-core` repository that updates the Formula. I don't really know if there is a _great_ way to test this without publishing releases manually now. Also we should figure out which `HOMEBREW_GITHUB_API_TOKEN` we should provide, as I don't think the `github-actions-bot` has enough permissions to create a PR in another repository.

- [x] Figure out the right API token.
- [x] Test the workflow.

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -1,0 +1,19 @@
+name: homebrew-update-formula
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  deploy:
+
+    runs-on: macos-latest
+
+    env:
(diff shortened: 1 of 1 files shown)


o2sh:
Hi @erikgaal, I hope you're doing great :raised_hands:

Sorry for the very long delay :pray::pray: I just created HOMEBREW_GITHUB_API_TOKEN with repo scope.

Is this what you were expecting ?

erikgaal:
Hi @o2sh,

Sorry for taking so long to respond on this again. I've updated the workflow to make it work again. I think we're set now!

o2sh (approved):

o2sh:
LGTM :+1: 

Can't wait for the next release to test it in real conditions.

Thanks @erikgaal

PR #162, opened 2019-11-12
Update cpp.ascii

Sh3mm:
used the c.ascii and modified it because the cpp ascii was lacking.


o2sh:
Looks great, thanks @Sh3mm

Issue #165, opened 2019-11-16
Add Pending changes info line


o2sh:
As an additional info line, we wish to inform the user about the pending changes (`git status`) on the current Repo, something like:

![DeepinScreenshot_select-area_20191116141648](https://user-images.githubusercontent.com/13710835/68993758-c3f78680-087b-11ea-8de5-cc0d0d6a1fa4.png)

which reads as --> `6 modified files, 2 new files, 7 deleted files`

This line should be shown only if pending changes exist.

:100: :1st_place_medal:

PR #166, opened 2019-11-17
Add changes line

axdiamond:
This addresses #165 

Onefetch will now show a changes line when there are uncommitted changes in the repo!

Basically, it takes the output of `git diff --shortstat HEAD`, and replace the text with appropriate symbols. 

    files changed into +-
    insertions(+) into +
    deletions(-) into -


o2sh:
Thanks a lot for your contribution @axdiamond 

Few remarks:
* I think it would be better to call it `Pending` instead of `Changes`, to let the user know that this are only uncommited changes + it prevents having to treat singular/plural cases --> `Change(s)`
* About the positioning, maybe place it just below `HEAD`
* You are mixing information at line level (`insertion(s)(+)`, `deletion(s)(-)`) with information at file level (`file(s) changed`), maybe `git diff ---shortstat` is not the right command for the job...We only want **file level infos** for now, `git status` shows it pretty well.

Looking forward to merge your PR :+1:

axdiamond:
@o2sh I did the other changes, but I'm not sure how to get lines changed from git.

git status only shows files modified, not lines. 

Do you have any suggestions on how I should get the lines changed?

o2sh:
> git status only shows files modified, not lines.

That's right, we are only interested in file level changes.

axdiamond:
Ooooohhh I'm a silly goose. 

How does it look now?

axdiamond:
I'm new to rust btw, so any nitpicking is welcome!

o2sh:
I detected one missing checks --> `"A" | "AM" |"??" => added += 1`, (AM)

---

if one of these 
```
let mut deleted = 0;
let mut added = 0;
let mut modified = 0;
```
remains equal to zero, it shouldn't be shown in the result.

---

> I'm new to rust btw, so any nitpicking is welcome!

LGTM, @spenserblack did I miss smth?

o2sh:
if (modified == 0 || modified && added ==  0) , you get a leading whitespace --> I'll trim the result.

---
Thanks a lot for your help @axdiamond , your contribution is very much appreciated  :1st_place_medal:

PR #170, opened 2019-11-27
Delete default.nix

filalex77:
After https://github.com/NixOS/nixpkgs/pull/74191 gets merged, these files may no longer be necessary.

cc @Kloenk

---
This change is [Reviewable](https://reviewable.io/reviews/o2sh/onefetch/170)


jonringer:
it may still be valuable to ensure that you work with nixpkgs-unstable, personally, I would keep the default.nix

filalex77:
@jonringer Didn't know about that. How's that done?

jonringer:
you can do `nix build -f .` and it should try to evaluate+build your default.nix. Since `src =./.;`, it will differ from what's on nixpkgs

filalex77:
Thanks for the input. I'll close the PR, and the discussion will remain for posterity.

jonringer:
I wouldn't worry about it too much, the main thing to remember is that `<nixpkgs>` will pull from the channels, and this will likely be slight (or majorly) different for everyone.

If you want a lot of reproducibility, i would freeze nixpkgs at a commit, if you want to ensure that it roughly works with nixpkgs master, then I would leave as-is.

PR #182, opened 2020-08-02
Add info for cargo-deb

SuperSandro2000:
This just adds some fields used by cargo deb to avoid having a markdown extended description and sets the correct section. The other two default to those values but it does not hurt to set them explizit.


o2sh:
Thanks for your contribution @SuperSandro2000

PR #199, opened 2020-08-29
tokio command for non blocking sys call

o2sh:
Following the discussion in #186 
Turns out tokio provides a non blocking `Command` function which returns a Future.

The code is now effectively asynchronous :

```
get_language_stats --> 2.768313294s
get_configuration --> 21.916µs
get_current_commit_info --> 774.652µs
get_project_license --> 263.539183ms
get_version --> 292.895558ms
get_pending_pending --> 293.050209ms
get_last_change --> 294.630701ms
get_git_info --> 298.447978ms
get_packed_size --> 388.507071ms
get_commits --> 10.346720716s
get_creation_time --> 15.720670077s
get_authors --> 15.944136695s

```
The code is unfortunately less clean as I had to replace the async block with a separate function `get_info_lines` in order to mark it with the #[tokio::main], @spenserblack maybe it can be improved ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "0.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "arrayref"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


o2sh:
> Could Info::new itself be given the #[tokio::main] attribute instead of a new function?

Done.
By doing so, I add to make the Info::new function async which then allowed me to get rid of the nested async block.

PR #203, opened 2020-09-02
Exempt "help wanted" from stale bot

spenserblack:
Since `help wanted` implies that an issue is valid for onefetch, I thought it should be exempted.

What do you think?


o2sh:
Yes, definitely!

Issue #205, opened 2020-09-03
Refactor language.rs


o2sh:
Currently, adding support for a new language requires modifying the language.rs at multiple locations.
This process can be quite cumbersome and error prone (especially for new contributors).

Some shortcuts can be taken:

- Assuming the ascii file to be in "../resources/language.ascii" => match case in `get_ascii_art()` no more needed
- Using `tokei::LanguageType` instead of our own Enum, else (if not possible) assume that the name matches => match case in fn `from(language: tokei::LanguageType)` no more needed.
- ...

spenserblack:
We could probably use a macro for most language definition (the [paste crate](https://crates.io/crates/paste) is good for this). This macro could also define tests.

PR #206, opened 2020-09-03
Define languages with macro

spenserblack:
For #205 

~~This macro works mostly, except for a few mismatches with some tokei language names. These mismatches could probably be resolved by making the macro rule a bit more complex to allow onefetch => tokei language mappings.~~ 

This macro now works.

The implementation of the macro itself can be changed if we use types from tokei instead of defining our own type. But this is the basic idea.

It can (and probably should) be adjusted to allow color implementations as well.


o2sh:
This looks great @spenserblack, macro_rules! indead :smile: 

Maybe you could include get_all_language_types() inside the macro:

```
fn get_all_language_types() -> Vec<tokei::LanguageType> {
    vec![
        $( tokei::LanguageType::$name, )*
    ]
}
```
So that we can remove the big Vec! at the end of the file.

> It can (and probably should) be adjusted to allow color implementations as well.

Sure! tell me if you need a hand :+1:

spenserblack:
OK, this macro should be good to go.
I've included tests built by the macro to enforce width and height, but they are ignored by default for now so the CI won't fail. They can be explicitly tested with `cargo test -- --ignored`.

o2sh:
Great work @spenserblack 
I think, we should move to  a more reasonable max-width of 45. By doing so, the number of ascii logos to fix drops from 17 to 11:

```
test language::ascii_size::clojure_width ... FAILED
test language::ascii_size::csharp_width ... FAILED
test language::ascii_size::cpp_width ... FAILED
test language::ascii_size::forth_width ... FAILED
test language::ascii_size::groovy_width ... FAILED
test language::ascii_size::jupyter_height ... FAILED
test language::ascii_size::jupyter_width ... FAILED
test language::ascii_size::lua_height ... FAILED
test language::ascii_size::lua_width ... FAILED
test language::ascii_size::nim_width ... FAILED
test language::ascii_size::objectivec_width ... FAILED
test language::ascii_size::swift_width ... FAILED
test language::ascii_size::zig_width ... FAILED
```
I'll create one issue per language

o2sh (approved):

Issue #198, opened 2020-08-29
Use git log once to derive get_commits/get_creation_time/get_authors 


o2sh:
Following a Suggestion by @spenserblack in #197 

-------

Considering the significant impact on the response time of the `git log` command. It seems more reasonable to call it only once and use its output to derive:

- Number of Commits: get_commits()
- Creation Time: get_creation_time()
- Authors: get_authors()

spenserblack:
`git log --pretty` could probably be used to get info in a way we can parse. The `.lines()` function returns a double-ended iterator, so you could call `.next_back()` to get stats of first commit, then iterate the rest normally.

PR #211, opened 2020-09-20
Use git log once for get_commits/get_creation_time/get_authors and get_last change

o2sh:
Fixes  #198 

Hopefully the code isn't too messy...:shit::shit:

Response time on the linux repo:

```
get_language_stats --> 2.496199699s
get_configuration --> 556ns
get_current_commit_info --> 586.312µs
get_project_license --> 149.397478ms
get_version --> 168.437398ms
get_pending_pending --> 167.720066ms
get_git_info --> 171.087506ms
get_packed_size --> 198.697039ms
get_git_history --> 12.227256108s
get_creation_date --> 2.054µs
get_authors --> 177.582094ms
get_last_change --> 1.343µs
```

The new entry `get_git_history` is the new function responsible for executing the `git log`, the results are then passed on to the other functions.

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -367,7 +367,7 @@ impl Info {
             args.push("--no-merges");
         }
 
-        args.push("--pretty=format:%cr%x09%an");
+        args.push("--pretty=\"format:%cr\t%an\"");
 
         let output = Command::new("git")
             .args(args)
@@ -443,15 +443,14 @@ impl Info {
         let mut authors = std::collections::HashMap::new();
         let mut total_commits = 0;
         for line in git_history {
-            let commit_author = line.split('\u{09}').collect::<Vec<_>>()[1].to_string();
+            let commit_author = line.split('\t').collect::<Vec<_>>()[1].to_string();
             let commit_count = authors.entry(commit_author.to_string()).or_insert(0);
             *commit_count += 1;
             total_commits += 1;
         }
 
         let mut authors: Vec<(String, usize)> = authors.into_iter().collect();
-        authors.sort_by_key(|(_, c)| *c);
-        authors.reverse();
+        authors.sort_by(|(_, a_count), (_, b_count)| b_count.cmp(a_count));
 
         authors.truncate(n);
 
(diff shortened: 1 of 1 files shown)


src/info.rs
```diff
@@ -362,6 +361,24 @@ impl Info {
         })
     }
 
+    async fn get_git_history(dir: &str, no_merges: bool) -> Vec<String> {
+        let mut args = vec!["-C", dir, "log"];
+        if no_merges {
+            args.push("--no-merges");
+        }
+
+        args.push("--pretty=format:%cr%x09%an");
```
spenserblack, line 370: Just for slightly more readability (and keeps the character a literal instead of git having to parse it). The tab character doesn't separate commands, so the extra quotes probably aren't necessary, but I added them to the suggestion just in case there's different behavior in other environments.
```suggestion
        args.push("--pretty=\"format:%cr\t%an\"");
```

src/info.rs
```diff
@@ -630,44 +613,25 @@ impl Info {
...
-            .expect("Failed to execute git.");
-
-        let output = String::from_utf8_lossy(&output.stdout);
+    fn get_creation_date(git_history: &[String]) -> Result<String> {
+        let first_commit = git_history.last();
 
-        let output = match output.lines().next() {
-            Some(creation_time) => creation_time.replace('"', ""),
+        let output = match first_commit {
+            Some(creation_time) => creation_time.split('\u{09}').collect::<Vec<_>>()[0].to_string(),
```
spenserblack, line 631:
```suggestion
            Some(creation_time) => creation_time.split('\t').collect::<Vec<_>>()[0].to_string(),
```

src/info.rs
```diff
@@ -422,39 +439,22 @@ impl Info {
...
+            let commit_author = line.split('\u{09}').collect::<Vec<_>>()[1].to_string();
+            let commit_count = authors.entry(commit_author.to_string()).or_insert(0);
             *commit_count += 1;
             total_commits += 1;
         }
 
-        // sort authors by commit count where the one with most commit count is first
         let mut authors: Vec<(String, usize)> = authors.into_iter().collect();
         authors.sort_by_key(|(_, c)| *c);
         authors.reverse();
```
spenserblack, line 454: While we're adjusting these functions, this should probably be simplified.
```suggestion
        authors.sort_by(|(_, a_count), (_, b_count)| b_count.cmp(a_count));
```

spenserblack (changes requested):
Looks great!
The suggestions involving tabs can be ignored, because they're fairly inconsequential, but I left 1 suggestion on how authors are sorted that you probably would want to look at.

spenserblack (approved):
:tada:

PR #218, opened 2020-10-04
Changed to a smaller Lua logo

maash3r:
Made a smaller ASCII Lua logo to fit the size. Also changed the outer circle to represent a dotted line so as to follow the original logo. Hope you like it.


o2sh:
Great work @maash3r !
To make it look even better, you could maybe add another color. For example, make the outer circle white.

And also if you could get rid of the character "2" (or replace it with "@") in the ASCII logo, it looks a bit out of place.

maash3r:
I'll work on the color and that '2' was an oversight, sorry, will replace it and update soon

o2sh:
This is perfect, thanks a lot @maash3r 

As you mentioned in one of your previous comments, this is not the only logo that exceeds the max size. So, I opened new issues in that regard --> #219 #220 #222 #223 

I hope to see more contributions from your part soon :100:

PR #229, opened 2020-10-05
222 new objective c logo

atluft:
Trouble locating an official logo for objective c, xcode tools label the files with an ".m".
Looked at string to ASCII versions of "[Obj-C]", but they are way too wide.

The "j" seems a bit weak, any thoughts on making that one letter pop?


o2sh:
Design wise, I think this is already a great improvement.

However, it exceeds the max-width.

atluft:
@o2sh , thanks for mentioning the width. Updated now to be correct width.

PR #237, opened 2020-10-06
feat: show number of tags and branches

### LINKED ISSUE #202: show number of tags & branches
Hello,
It would be great if onefetch show the number tags & branches.

thx

reime005:
This simply counts and displays the number of branches and tags that the repository has locally.

Closes #202

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -32,6 +32,8 @@ pub struct Info {
     pending: String,
     repo_size: String,
     number_of_lines: usize,
+    number_of_tags: usize,
+    number_of_branches: usize,
     license: String,
     custom_logo: Language,
     custom_colors: Vec<String>,
@@ -74,10 +76,38 @@ impl std::fmt::Display for Info {
             )?;
         }
         if !self.disable_fields.project {
-            write_buf(
-                &mut buf,
-                &self.get_formatted_info_label("Project: ", color),
-                &self.project_name,
+            let branches_str = if self.number_of_branches == 1 {
+                String::from("1 branch")
+            } else if self.number_of_branches > 1 {
+                format!("{} branches", self.number_of_branches)
+            } else {
+                String::new()
+            };
+
+            let tags_str = if self.number_of_tags == 1 {
+                String::from("1 tag")
+            } else if self.number_of_tags > 1 {
+                format!("{} tags", self.number_of_tags)
+            } else {
+                String::new()
+            };
+
+            let branches_tags_str = if tags_str.is_empty() && !branches_str.is_empty() {
+                format!("( {} )", branches_str)
+            } else if branches_str.is_empty() && !tags_str.is_empty() {
+                format!("( {} )", tags_str)
+            } else if !branches_str.is_empty() {
+                format!("( {}, {} )", tags_str, branches_str)
+            } else {
+                String::new()
+            };
+
+            let project_str = &self.get_formatted_info_label("Project: ", color);
+
+            writeln!(
+                buf,
+                "{}{} {}",
+                project_str, self.project_name, branches_tags_str
             )?;
         }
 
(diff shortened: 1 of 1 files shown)


spenserblack (changes requested):
Thanks for your interest and your PR!
Left a comment on `async` usage.

o2sh:
Also, I don't think it's necessary to have two new lines just for that. Maybe, you can add them to the "Project" info_line.

Something like, `Project: onefetch (2 branches, 17 tags)` 

What do you think?

-------
Looking forward to merge your PR 👍

reime005:
@o2sh @spenserblack thanks for the feedback. I changed the code

spenserblack (approved):

src/info.rs
```diff
@@ -74,10 +76,17 @@ impl std::fmt::Display for Info {
             )?;
         }
         if !self.disable_fields.project {
-            write_buf(
-                &mut buf,
-                &self.get_formatted_info_label("Project: ", color),
-                &self.project_name,
+            let branches_tags_str = &format!(
```
o2sh, line 79: You should build branches_tags_str based on what's not equal to zero:
```

if self.number_of_branches == 0 &&  self.number_of_tags == 0 => branches_tags_str = ""
else {
if (self.number_of_branches != 0) number_of_branches_str = "{self.number_of_branches} branches " // maybe check if > 1 for plural
if (self.number_of_tags != 0) number_of_tags_str = "{self.number_of_tags} tags"
branches_tags_str = "( {number_of_tags_str}, {number_of_tags_str})" // don't show the comma if no tags
}
```

src/info.rs
```diff
@@ -378,6 +391,34 @@ impl Info {
...
+                .expect("Failed to execute git.");
+
+            let tags = String::from_utf8_lossy(&output.stdout);
+
+            tags.lines().count()
+        };
+
+        let branches = async {
+            let output = Command::new("git")
+                .args(vec!["-C", dir, "branch"])
```
o2sh, line 409: Add `-r` flag to `git branch` otherwise you're only showing local branches. I think we're more interested with remote branches. And maybe try to filter out `origin/HEAD -> origin/master`

o2sh (commented):
Left a few comments

reime005:
@o2sh changed again. filter does not make sense to me

src/info.rs
```diff
@@ -74,10 +76,36 @@ impl std::fmt::Display for Info {
...
-                &self.project_name,
+            let mut branches_tags_str: String = String::new();
+            let mut tags_str: String = String::new();
+            let mut branches_str: String = String::new();
+
+            if self.number_of_branches == 1 {
+                branches_str = String::from("1 branch");
+            } else if self.number_of_branches > 1 {
+                branches_str = format!("{} branches", self.number_of_branches);
+            }
```
spenserblack, line 87: This allocates memory for a `String`, then potentially creates a new `String`.
if-else statements can be expressions. I think the more "rusty" way of doing this is
```rust
let branches_str = if self.number_of_branches == 1 {
    String::from("1 branch")
} else if self.number_of_branches > 1 {
    format!("{} branches")
} else {
    String::new()
}
// same change for other values
```
This will result in less memory allocations.

spenserblack (changes requested):
Thanks for making those changes :smiley:
Left another comment

src/info.rs
reime005: true, that makes sense

reime005: adjusted

spenserblack: Thanks!

spenserblack (approved):
Thanks again for the changes! I'm approving, but I'll defer to approval from @o2sh for merging.

o2sh:
LGTM :+1:

PR #241, opened 2020-10-07
Fix grammar error in README.md

CodeLongAndProsper90:
Fix a slight grammar error in the README.


o2sh:
You're right, thanks @CodeLongAndProsper90

PR #243, opened 2020-10-08
Identify author by email

yoichi:
before
スクリーンショット 2020-10-08 19 07 38

after
スクリーンショット 2020-10-08 19 08 36

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -394,7 +394,7 @@ impl Info {
             args.push("--no-merges");
         }
 
-        args.push("--pretty=%cr\t%an");
+        args.push("--pretty=%cr\t%ae\t%an");
 
         let output = Command::new("git")
             .args(args)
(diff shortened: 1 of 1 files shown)


src/info.rs
```diff
@@ -489,10 +489,13 @@ impl Info {
 
     fn get_authors(git_history: &[String], n: usize) -> Vec<(String, usize, usize)> {
         let mut authors = std::collections::HashMap::new();
+        let mut names = std::collections::HashMap::new();
         let mut total_commits = 0;
         for line in git_history {
             let commit_author = line.split('\t').collect::<Vec<_>>()[1].to_string();
```
o2sh, line 495: This should probably be called `commit_author_email`

src/info.rs
```diff
@@ -489,10 +489,13 @@ impl Info {
...
+            let author_name = line.split('\t').collect::<Vec<_>>()[2].to_string();
```
o2sh, line 496: This should probably be called `commit_author_name`

src/info.rs
```diff
@@ -489,10 +489,13 @@ impl Info {
 
     fn get_authors(git_history: &[String], n: usize) -> Vec<(String, usize, usize)> {
         let mut authors = std::collections::HashMap::new();
+        let mut names = std::collections::HashMap::new();
```
o2sh, line 492: This should probably be called `author_name_by_email`

o2sh (commented):
Using the email as a UID is definetely the right way to do it. Thanks @yoichi 

I suggested some renaming to make it clearer.

src/info.rs
yoichi: Thanks. I think the following variable names are good because of the relationship with the format string of git log.

-  %an = author_name
- %ae = author_email

I've renamed commit_author to author_email in c68e0210

PR #245, opened 2020-10-08
Regex matching for remote.url

o2sh:
`get_repo_name_and_url()` didn't work on repositories having remotes not called "origin".

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -440,7 +440,7 @@ impl Info {
         let mut remote_url = String::new();
         let mut repository_name = String::new();
 
-        let remote_regex = Regex::new(r"remote.[a-zA-Z0-9]+.url").unwrap();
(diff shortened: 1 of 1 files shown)


src/info.rs
```diff
@@ -439,9 +440,11 @@ impl Info {
         let mut remote_url = String::new();
         let mut repository_name = String::new();
 
+        let remote_regex = Regex::new(r"remote.[a-zA-Z0-9]+.url").unwrap();
```
spenserblack, line 443:
```suggestion
        let remote_regex = Regex::new(r"remote\.[a-zA-Z0-9]+\.url").unwrap();
```
`.` means any character in regex. Did you mean to use `\.`?

o2sh: You're absolutely right! the `.` is not meant to be a regex pattern

PR #246, opened 2020-10-08
fix number of branches to ignore HEAD

o2sh:
Exclude `origin/HEAD -> origin/master` from total of remote branches

### DIFF
--- a/src/info.rs
+++ b/src/info.rs
@@ -428,7 +428,11 @@ impl Info {
 
             let branches = String::from_utf8_lossy(&output.stdout);
 
-            branches.lines().count() - 1 //Exclude origin/HEAD -> origin/master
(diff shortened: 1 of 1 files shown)


src/info.rs
```diff
@@ -428,7 +428,7 @@ impl Info {
 
             let branches = String::from_utf8_lossy(&output.stdout);
 
-            branches.lines().count()
+            branches.lines().count() - 1 //Exclude origin/HEAD -> origin/master
```
spenserblack, line 431: If onefetch is run in a repository with no branches, such as a repo that was just created but with no commits, this may result in an unsigned integer underflow error

spenserblack (approved):
LGTM :+1:

PR #247, opened 2020-10-09
Updated Tex's ASCII logo

maash3r:
Changed Tex's ASCII logo to follow the defined standards. Also I would call it a _rich_ logo 😉


o2sh:
> Also I would call it a rich logo :wink: 

:laughing: Indeed, that's lot of :moneybag::moneybag:

Great job @maash3r

Issue #235, opened 2020-10-06
Support COLORTERM for TrueColor ASCII


atluft:
After working on the swift logo I was thinking it might be useful to be able to specify true colors for a better result:

![image](https://user-images.githubusercontent.com/2379260/95237152-d5b88400-07d5-11eb-9c56-03a332646e45.png)

Terminals that don't support COLORTERM environment variable would look like this:
![image](https://user-images.githubusercontent.com/2379260/95237267-fa146080-07d5-11eb-83b5-c711e21e6788.png)

o2sh:
This is very good idea @atluft 
For the final implementation, you should probably modify the `define_languages!` macro to have a second Vec! for TrueColor. By Doing that, you will be able to get rid of _the offset of 10_. Also, It's preferable to read env::var("COLORTERM") only once and pass it along.

atluft:
Will do, new to Rust programming and having your advice helps. All comments are welcome!

spenserblack:
@atluft Just curious: did you test how the `colored` crate renders TrueColor on a terminal that doesn't support it? I would assume that it would pick the closest available color if it can't render the TrueColor.

spenserblack:
Nevermind, I checked the source code and it appears that `colored` *doesn't* check if COLORTERM is set or not.

I think the best way would probably be to have the `define_languages!` macro call *another* macro, like `define_colors!`, and also define a struct for colors, like
```rust
struct Colors {
    default_colors: Vec<Color>,
    true_colors: Option<Vec<Color>>,
}
```

`define_colors!` can take 2 patterns:
- `vec![ $( color:expr ),+ ]`, the current usage with no true colors, would expand to
  ```rust
  Colors {
      default_colors: vec![ $( $color ,)+ ],
      true_colors: None,
  }
  ```
- `$color_struct:expr`, which would take a definition of the `Colors` struct, and wouldn't need any special expansion rules.

Does that make sense? :smiley_cat:

o2sh:
> did you test how the colored crate renders TrueColor on a terminal that doesn't support it

Just tested it on rxvt, it prints it in plain black:

![rxvt](https://user-images.githubusercontent.com/13710835/95335414-a9713600-08af-11eb-80df-43fe57d703bb.png)

spenserblack:
Thanks, @o2sh!
[...]

PR #252, opened 2020-10-09
235 truecolor define color

atluft:
Please consider this change to add true colors mentioned in issue #235 

The additional braces are needed in define_languages to change the capture of $colors:expr (a single thing) to a stream of tokens $( $colors:tt )*. The stream of tokens is required for the nested call to define_colors which utilizes the vec! ... token stream as a means of choosing the vector based expansion instead of the struct Colors (a single thing) expansion.

See [this](https://users.rust-lang.org/t/macro-not-expanding-when-nested/10342/2) excellent explanation of how the parent macro globs an expr into a single thing while tt preserves the token stream.


spenserblack:
:thinking: no CI checks. @o2sh, should the `on: [pull_request]` rule be added to the GitHub action?

src/language.rs
```diff
@@ -4,10 +4,22 @@ use {
...
 };
 
+#[derive(Clone)]
+pub struct Colors {
+    basic_colors: Vec<Color>,
+    true_colors: Option<Vec<Color>>,
+}
+
+macro_rules! define_colors {
+  ( vec![ $( $color:expr ),+ ] ) => { Colors { basic_colors: vec![$( $color ),+ ], true_colors: None } };
```
spenserblack, line 17:
```suggestion
  ( vec![ $( $color:expr ),+ $(,)? ] ) => { Colors { basic_colors: vec![$( $color ),+ ], true_colors: None } };
```
This allows an optional trailing comma, so that the patterns `vec![foo, bar, ]` and `vec![foo, bar]` are both accepted.

src/language.rs
```diff
@@ -4,10 +4,22 @@ use {
...
+  ( $color:expr ) => { $color };
+}
+
 macro_rules! define_languages {
-    ($( { $name:ident, $ascii:literal, $display:literal, $colors:expr $(, $serialize:literal )? } ),* ,) => {
+    ($( { $name:ident, $ascii:literal, $display:literal, {$($colors:tt)*} $(, $serialize:literal )? } ),* ,) => {
```
spenserblack, line 22: Is it necessary for the pattern to be `:tt`? This is the least specific pattern possible, and tends to break pattern matching. Generally, I've only seen `$( $tokens:tt )*` used at the *end* of macro rule definitions, to mark optional extra tokens. In fact, I'm a bit surprised this isn't breaking pattern matching, since `}` is a token. I would assume that `{$($colors:tt)*}` would create ambiguity on whether the closing `}` is *part* of the stream of tokens or *after* the stream of tokens.
I would expect `$colors:expr` to be sufficient, since both a `vec!` declaration and a struct declaration should match an expression.

o2sh:
Absolutely, thanks for the heads up @spenserblack

src/language.rs
```diff
@@ -46,11 +58,15 @@ macro_rules! define_languages {
...
+                let colors = match *self {
+                    $( Language::$name => define_colors ! ( $($colors)* ), )*
+                    Language::Unknown => define_colors ! ( vec![Color::White] ),
+                };
+                let use_true_color = env::var("COLORTERM");
+                if use_true_color.is_ok() && use_true_color.unwrap().to_lowercase() == "truecolor" && colors.true_colors.is_some() {
+                    colors.true_colors.unwrap()
+                } else {
+                    colors.basic_colors
                 }
```
spenserblack, line 70: See if this will work. Checking `is_ok` and then unwrapping should probably be replaced with pattern matching, with either `if let` or `match`.
```suggestion
                if let (Ok("truecolor"), Some(colors)) = (use_true_color.map(|s| s.to_lowercase()), colors.true_colors) {
                    colors
                } else {
                    colors.basic_colors
                }
```

src/language.rs
```diff
@@ -101,62 +117,94 @@ macro_rules! define_languages {
 }
 
 define_languages! {
-    { Assembly, "assembly.ascii", "Assembly", vec![Color::Cyan] },
```
spenserblack, line 104: I was afraid of this additional complexity.
Ideally, these two should be valid.
```
{ Assembly, "assembly.ascii", "Assembly", { Colors { basic_colors: vec![Color::Cyan], true_colors: None } } },
{ C, "c.ascii", "C", vec![Color::Cyan, Color::Blue] },
```
That is, the inclusion of the braces should be optional IMO.

spenserblack (commented):
Thanks for the PR!
Left some notes.

src/language.rs
atluft: @spenserblack thank you for your review, I greatly value any feedback on this. I simply don't know what a good answer looks like for this. What would you advise?

You are correct.  ```:tt``` is not required and other approaches exist. I wasn't sure what a good solution looks like for this and selected to keep everything inside macro expansion because that is where I think the problem begins.

AFAIK, the problem is when ```$colors``` is ```expr``` in the parent macro both ```vec![ Color::Red]``` and ```Color { ... }``` are matches for ```expr```. Unfortunately, the parent match provides a singular match for nested macros like ```define_colors```. The nested macro won't match ```vec!``` because the literal token ```vec!``` isn't available (its been flattened), instead ```vec! ...``` matches ```expr``` in ```define_colors``` and is processed like ```Colors``` which is wrong.

An alternative I found online uses ```downcast_ref``` to do type checking after macro expansion. This approach has less impact on the original table, but delays separating ```vec``` and ```Colors``` until type checking is possible. I am not sure if that is compile time or runtime checking in this case.

``` rust
macro_rules! define_colors {
    ( $color:expr ) => {
        if let Some(vector) = (&$color as &dyn Any).downcast_ref::<Vec<Color>>() {
            Colors { basic_colors: vector.clone(), true_colors: None }
        }
        else if let Some(colors) = (&$color as &dyn Any).downcast_ref::<Colors>() {
            colors.clone()
        }
        else {
            Colors { basic_colors: vec![ Color::White ], true_colors: None }
        }
    };
}
```

src/language.rs
```diff
@@ -68,19 +58,16 @@ macro_rules! define_languages {
...
-                match color_struct.true_colors {
-                    Some(true_colors) => match env::var("COLORTERM") {
-                        Ok(val) => if val.to_lowercase() == "truecolor" { colors = true_colors; },
-                        Err(_e) => { },
-                    },
-                    None => { },
-                };
-                colors
+                let use_true_color = env::var("COLORTERM");
+                if use_true_color.is_ok() && use_true_color.unwrap().to_lowercase() == "truecolor" && colors.true_colors.is_some() {
```
o2sh, line 66: Maybe extract it into a separate function and include the "24bit" test
```
fn is_truecolor_terminal() -> bool {
    env::var("COLORTERM")
        .map(|colorterm| colorterm == "truecolor" || colorterm == "24bit")
        .unwrap_or(false)
}
```

src/language.rs
spenserblack: I see.
An alternative is to use `define_colors!` inside the `define_languages!` macro invocation. Here's a very simple example.

```rust
macro_rules! define_colors {
    ( [ $( color:expr ),+ ] ) => { Colors { basic_colors: vec![$( $color ),+ ], true_colors: None } };
    ( $color:expr ) => { $color };
}

macro_rules! define_languages {                                               
     ($( { $name:ident, $ascii:literal, $display:literal, $colors:expr $(, $serialize:literal )? } ),* ,) => {
        /* macro definition */
    }
}

define_languages! {
    { C, "c.ascii", "C", define_colors!( [Color::Cyan, Color::Blue] ) },
}
```
I slightly changed the first rule of `define_colors!` so that it doesn't require `vec!`. Since it will always be added, might as well let the macro deal with it.

atluft: @spenserblack , thank you and I agree moving define_colors into the invocation is an improvement.

src/language.rs
atluft: reworked define_colors to not utilize vec.

spenserblack (changes requested):
It would be great if you could add some tests to the macro that
- enforce that the number of `true_colors`, if defined, equal the number of `basic_colors`
- enforce that there are no TrueColors in the `basic_colors` field

src/language.rs
atluft: agree, implemented in latest attempt.

atluft:
> It would be great if you could add some tests to the macro that
> 
> * enforce that the number of `true_colors`, if defined, equal the number of `basic_colors`
> * enforce that there are no TrueColors in the `basic_colors` field

Agree, I can try, what would the appropriate action be? Is ```panic!``` correct?

spenserblack:
>Agree, I can try, what would the appropriate action be?

Just like how the macro defines a `mod ascii_size` that's conditionally compiled when testing, you can add a mod for testing true colors.

The tests can `assert_eq!` the lengths. For testing no TrueColors in `basic_colors`, a `panic!` might be sufficient.

atluft:
A basic colors failure looks like below. In this simulated case the C language is using a TrueColor instead of Blue.

```
---- onefetch::language::true_colors::c_basic_color_test stdout ----
thread 'onefetch::language::true_colors::c_basic_color_test' panicked at ' language C has true color Color::TrueColor { r: 0, g: 0, b: 255 } in basic colors at index 1 please change to a basic color: Black, Red, Green, Yellow, Blue, Magenta, Cyan, White, BrightBlack, BrightRed, BrightGreen, BrightYellow, BrightBlue, BrightMagenta, BrightCyan, BrightWhite', src/onefetch/language.rs:182:1

```

atluft:
A mismatch in array counts for colors looks like below. In this case the swift languag has 10 true colors but only 9 basic colors.

```
---- onefetch::language::true_colors::swift_color_vector_length_test stdout ----
thread 'onefetch::language::true_colors::swift_color_vector_length_test' panicked at 'assertion failed: `(left == right)`
  left: `9`,
 right: `10`:  left (basic) color length do not match right (true) color length.
Colors {
    basic_colors: vec![
        Color::Red, // 0
        Color::Red, // 1
        Color::Red, // 2
        Color::Red, // 3
        Color::Red, // 4
        Color::Red, // 5
        Color::Red, // 6
        Color::Red, // 7
        Color::Red, // 8
    ],
    true_colors: vec![
        Color::TrueColor { r: 248, g: 129, b: 52 }, // 0
        Color::TrueColor { r: 249, g: 119, b: 50 }, // 1
        Color::TrueColor { r: 249, g: 109, b: 48 }, // 2
        Color::TrueColor { r: 250, g: 99, b: 46 }, // 3
        Color::TrueColor { r: 250, g: 89, b: 44 }, // 4
        Color::TrueColor { r: 251, g: 80, b: 42 }, // 5
        Color::TrueColor { r: 251, g: 70, b: 40 }, // 6
        Color::TrueColor { r: 252, g: 60, b: 38 }, // 7
        Color::TrueColor { r: 252, g: 50, b: 36 }, // 8
        Color::TrueColor { r: 253, g: 40, b: 34 }, // 9
    ],
}
', src/onefetch/language.rs:182:1
```

o2sh:
@atluft I just moved the `is_true_color_terminal()` func in cli.rs to make it a config parameter.

src/onefetch/language.rs
```diff
@@ -59,6 +73,66 @@ macro_rules! define_languages {
...
+                        let mut color_index = 0;
+                        for bc in colors.basic_colors.iter() {
+                            let color_str = &format!( "Color::{:?}", bc );
+                            assert!( !color_str.contains( "TrueColor" ), " language {} has true color {} in basic colors at index {} please change to a basic color: Black, Red, Green, Yellow, Blue, Magenta, Cyan, White, BrightBlack, BrightRed, BrightGreen, BrightYellow, BrightBlue, BrightMagenta, BrightCyan, BrightWhite", $display, color_str, color_index );
+                            color_index = color_index + 1;
+                        }
+                    }
+
+                    #[test]
+                    #[ignore]
```
spenserblack, line 125:
```suggestion
```
Tests like this probably shouldn't be ignored. When people make a PR that might contain invalid color settings, we would probably want the CI to catch that.

src/onefetch/language.rs
```diff
@@ -59,6 +73,66 @@ macro_rules! define_languages {
...
+                paste! {
+                    #[test]
+                    #[ignore]
+                    fn [<$name:lower _basic_color_test>] () {
+                        let colors = $colors;
+                        let mut color_index = 0;
+                        for bc in colors.basic_colors.iter() {
+                            let color_str = &format!( "Color::{:?}", bc );
+                            assert!( !color_str.contains( "TrueColor" ), " language {} has true color {} in basic colors at index {} please change to a basic color: Black, Red, Green, Yellow, Blue, Magenta, Cyan, White, BrightBlack, BrightRed, BrightGreen, BrightYellow, BrightBlue, BrightMagenta, BrightCyan, BrightWhite", $display, color_str, color_index );
+                            color_index = color_index + 1;
```
spenserblack, line 120:
```suggestion
                        for (index, bc) in colors.basic_colors.iter().enumerate() {
                            let color_str = &format!( "Color::{:?}", bc );
                            assert!( !color_str.contains( "TrueColor" ), " language {} has true color {} in basic colors at index {} please change to a basic color: Black, Red, Green, Yellow, Blue, Magenta, Cyan, White, BrightBlack, BrightRed, BrightGreen, BrightYellow, BrightBlue, BrightMagenta, BrightCyan, BrightWhite", $display, color_str, color_index );
```
Just a bit more readable IMO, and prevents you from having to declare a value outside the scope of the loop.

src/onefetch/language.rs
```diff
@@ -59,6 +73,66 @@ macro_rules! define_languages {
...
+            $(
+                paste! {
+                    #[test]
+                    #[ignore]
+                    fn [<$name:lower _basic_color_test>] () {
+                        let colors = $colors;
+                        let mut color_index = 0;
+                        for bc in colors.basic_colors.iter() {
+                            let color_str = &format!( "Color::{:?}", bc );
+                            assert!( !color_str.contains( "TrueColor" ), " language {} has true color {} in basic colors at index {} please change to a basic color: Black, Red, Green, Yellow, Blue, Magenta, Cyan, White, BrightBlack, BrightRed, BrightGreen, BrightYellow, BrightBlue, BrightMagenta, BrightCyan, BrightWhite", $display, color_str, color_index );
```
spenserblack, line 119: Is asserting on `String::contains` the best way? Can I suggest making use of Rust's pattern matching with `match` or `if let`?
For example,
```rust
if let Color::TrueColor { .. } = bc {
    panic!("your panic message here");
}
```
I'd also recommend against wording the failed test message as if it's addressed to a person, with instructions on how to pass the test. A failed test should simply state why it failed IMO (e.g. "TrueColor found in basic_colors").

spenserblack, line 119: `$display` isn't the name of the variable, but how the language is displayed on the terminal. The majority of the time they will be very similar, but they *could* be different. You can use `stringify!( $name )` to get a string representation of the language's identity.

src/onefetch/language.rs
```diff
@@ -59,6 +73,66 @@ macro_rules! define_languages {
...
+                        }
+                    }
+
+                    #[test]
+                    #[ignore]
+                    fn [<$name:lower _color_vector_length_test>] () {
+                        let colors = $colors;
+                        let bc_count = colors.basic_colors.len();
+                        let tc_count = if let Some(tcs) = &colors.true_colors { tcs.len() } else { bc_count };
+                        assert_eq!( bc_count, tc_count, " left (basic) color length do not match right (true) color length.\n{}", colors );
```
spenserblack, line 130:
```suggestion
                        if let Some(tcs) = &colors.true_colors { assert_eq!( bc_count, tcs.len(), " left (basic) color length do not match right (true) color length.\n{}", colors ); }
                        
```

src/onefetch/language.rs
atluft: Agree, see 36a9e9c

src/onefetch/language.rs
atluft: Agree, see 36a9e9c

src/onefetch/language.rs
atluft: Agree, see 36a9e9c

atluft: Agree, see 36a9e9c

src/onefetch/language.rs
atluft: Agree, see 36a9e9c

src/onefetch/language.rs
```diff
@@ -59,6 +73,62 @@ macro_rules! define_languages {
...
+                    output += "    ], \n";
+                    output += "}\n";
+                    write!( f, "{}", output )
+                }
+            }
+
+            $(
+                paste! {
+                    #[test]
+                    #[ignore]
```
spenserblack, line 109:
```suggestion
```
This PR is looking good! This will probably be my last suggested change.

atluft:
@spenserblack thanks for all your help!

spenserblack (approved):
Looks good to me :+1:
Last step is to fix the merge conflicts :slightly_smiling_face:

o2sh:
Absolutely love this feature by the way :heart_eyes: 

I couldn't wait for it to be merged to start playing with it 




Thanks a lot to both of you :+1:

Issue #180, opened 2020-06-27
[Breaking change Tokei] Handle Markdown and Jupyter Notebooks


o2sh:
As you may already know, Onefetch relies on [tokei](https://github.com/XAMPPRocky/tokei) for language detection.

Unfortunately, tokei shows Markdows and Jupyter Notebooks stats in separate sections apart from other languages:

![image](https://user-images.githubusercontent.com/13710835/85922726-cbf1bd00-b885-11ea-9ad0-a73f3b9df567.png)

And onefetch uses the "Code" column to account for each language's weight in the total distribution
Here we have 0 for Markdown and Jnb instead of 65 and 28 (Total). 

As a result, onefetch considers those languages to  represent 0% of the total distribution:

![image](https://user-images.githubusercontent.com/13710835/85922797-50dcd680-b886-11ea-911b-034fd89cb6e7.png)

We should find a way to use the "Total" value instead

--------
#151

PR #265, opened 2020-10-14
Handle Markdown and Jupyter Notebooks

o2sh:
Fix for #180 

New code distribution of the repo shown in the issue:

lacmus

### DIFF
--- a/src/onefetch/language.rs
+++ b/src/onefetch/language.rs
@@ -185,7 +185,7 @@ impl Language {
 
         let total: f64 = stats.iter().map(|(_, v)| v).sum();
 
-        if (total - 0 as f64).abs() < f64::EPSILON {
+        if total.abs() < f64::EPSILON {
             None
         } else {
             for (_, val) in stats.iter_mut() {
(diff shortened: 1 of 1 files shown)


src/onefetch/language.rs
```diff
@@ -163,18 +163,36 @@ impl Language {
...
+            if code == 0 {
+                continue;
+            }
+
+            stats.insert(Language::from(*language_type), code as f64);
+        }
+
+        let total: f64 = stats.iter().map(|(_, v)| v).sum();
+
+        if (total - 0 as f64).abs() < f64::EPSILON {
```
spenserblack, line 188: Is it necessary to subtract by 0? :thinking:

spenserblack (approved):
LGTM :+1:
Just had 1 question.

src/onefetch/language.rs
o2sh: It looks weird, I aggree; it's a clippy suggestion.

At first, I had written `total == 0 as f64`

spenserblack: Got it, that makes sense. I assume clippy wouldn't complain if it was this
```suggestion
        if total.abs() < f64::EPSILON {
```

o2sh: I should really stop relying blindly on clippy suggestions :weary:
Thanks @spenserblack

PR #266, opened 2020-10-15
Enable image backends on macOS

yoichi:
Enable image backends on macOS because we have terminals with graphics support on macOS.

mlterm supports sixel:
スクリーンショット 2020-10-15 19 14 52

kitty:
スクリーンショット 2020-10-15 19 12 55

NOTE: iTerm2 also supports sixel but it shows broken output (I think it is a iTerm2 issue):
スクリーンショット 2020-10-15 19 19 25

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ error-chain = "0.12"
 [target.'cfg(windows)'.dependencies]
 ansi_term = "0.12"
 
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(not(target_os = "windows"))'.dependencies]
 libc = "0.2.79"
 base64 = "0.13.0"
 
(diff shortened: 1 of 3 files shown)


o2sh:
> Enable image backends on macOS because we have terminals with graphics support on macOS.

Great! I don't have a machine running macOS so I couldn't test it for myself.

Your PR looks good. What would you think about replacing `#[cfg(not(target_os = "windows"))]` with `#[cfg(not(windows))]` ?

yoichi:
> Your PR looks good. What would you think about replacing `#[cfg(not(target_os = "windows"))]` with `#[cfg(not(windows))]` ?

Applied in 0cf7398b.

> > iTerm2 also supports sixel but it shows broken output
> 
> Yes, that's weird. Have you tried `onefetch -i /path/to/image --image-backend sixel`

I've already tried. It is not a problem of image backend selection.

yoichi:
FYI: I've reported a issue on iTerm2 https://gitlab.com/gnachman/iterm2/-/issues/9204, which causes the broken output

PR #270, opened 2020-10-17
Fix layout with sixel image backend

yoichi:
# Problem

With sixel image backend, unnecessary +2 on cursor up CSI sequence causes misalignment between image and text:

original

while +2 keeps alignment on terminal in fullscreen mode:

orig_full

# How to fix

* remove unnecessary +2 for terminal not in fullscreen mode
* fix row number calculation for terminal in fullscreen mode

# Screenshots of the fixed version

## not in fullscreen mode

fixed

## in fullscreen mode

fixed_full

# Environment

* onefetch runs on Ubuntu 20.04
* mlterm runs on Windows 10

### DIFF
--- a/src/onefetch/image_backends/sixel.rs
+++ b/src/onefetch/image_backends/sixel.rs
@@ -77,8 +77,10 @@ impl super::ImageBackend for SixelBackend {
             ioctl(STDOUT_FILENO, TIOCGWINSZ, &tty_size);
             tty_size
         };
-        let width_ratio = tty_size.ws_col as f64 / tty_size.ws_xpixel as f64;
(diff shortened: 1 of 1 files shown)


CephalonRho:
Both the previous version and your fixed version have the same (correct) alignment on my end (mlterm on NixOS).
I don't remember the exact reason I added 2 lines to the vertical movement, I tried a lot of offsets and those looked correct everywhere I tried.
You're right that removing it makes more sense mathematically, and it still works, so that seems fine.

I'm not sure what was wrong with the ratio calculations, I can only guess that there were rounding errors that I haven't encountered.

CephalonRho (approved):
Previous comment. Wrong button. Sorry.

PR #271, opened 2020-10-17
Use more color than 16 in sixel graphics

yoichi:
# Problem

In sixel backend, the colors are limited to a very small number 16.
If the image specified by the image option uses more colors, the vividness will be lost.

# Fix

Increase color number from 16 to 128.

The color of the beak has been reproduced.

# Side Effect

Image rendering time become longer.
On my environment, it takes ~0.24sec with 16 colors -> ~0.4sec with 128 colors

# Enironment

* onefetch runs on Ubuntu20.04
* mlterm runs on Windows 10


CephalonRho (changes requested):
The problem with this is that at least xterm isn't required to support more than 16 colors. This isn't because of a flaw with sixel graphics itself, it's because [the VT340 doesn't support more than 16 color map entries](https://www.vt100.net/docs/vt3xx-gp/chapter14.html#S14.3.3.1), so the right behavior from xterm would be to fail above that because it emulates the VT340.
It's possible to increase the amount of colors for emulators that support them, but 16 colors is a safe default that should work on everything emulating the VT340.

The best way to do this would be to query the terminal emulator for the amount of color map entries it supports, but I'm not sure if that's possible.
Alternatively there could be a command line argument to specify that value, that way there shouldn't be any case where sixel images can't be displayed because of a hardcoded value.

spenserblack:
I'm not familiar mlterm. Does it set an environment variable to signify the number of colors it supports? At the very least, I would expect it to set the `$TERM` environment variable to `mlterm` (though I'm not sure if something like
`if color_128_terms.any(|t| t == env::var("TERM")) { /* set to 128 colors */ }` would be the best way to do things).

yoichi:
Thanks @CephalonRho for clarifying the reason for the 16 colors.
I understand the existence of terminals that support sixel but not 256 colors and the need for support for them (although I haven't used them myself).

Thanks @spenserblack for suggestion to use environment variable. However, there are situations where that method cannot be used (ssh to target host from the terminal, or docker run from the terminal). We should query the terminal for those cases, but I don't know how to determine color support.

I'll consider a command line argument as @CephalonRho suggested:

> Alternatively there could be a command line argument to specify that value, that way there shouldn't be any case where sixel images can't be displayed because of a hardcoded value.

yoichi:
In 83459a8 I've added a new option --image-colors to specifiy number of colors to be used.
If the option is not given, use 16 colors as before.

image

src/onefetch/cli.rs
```diff
@@ -115,6 +116,15 @@ impl Cli {
...
                 .help("Which image BACKEND to use."),
         )
+        .arg(
+            Arg::with_name("image-colors")
+                .long("image-colors")
+                .value_name("NUM")
+                .takes_value(true)
+                .max_values(1)
+                .default_value("16")
+                .help("NUM of colors (16-256) used in image backend."),
```
spenserblack, line 126: You might want to use [`Arg::possible_values`](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.possible_values) to limit what numbers can be used.

src/onefetch/cli.rs
```diff
@@ -173,6 +183,17 @@ impl Cli {
             None
         };
 
+        let image_colors = if let Some(value) = matches.value_of("image-colors") {
```
spenserblack, line 186: If you have a `default_value` defined, then the arg should always have *some* value, so it would probably be safe to unwrap `matches.value_of("image-colors")`.

src/onefetch/cli.rs
```diff
@@ -173,6 +183,17 @@ impl Cli {
...
+            let colors = usize::from_str(value).unwrap();
```
spenserblack, line 187: If `Arg::possible_values` is not used, then it could be possible for a user to enter a non-number value, so unwrapping here may not be safe.

src/onefetch/cli.rs
```diff
@@ -115,6 +116,16 @@ impl Cli {
                 .possible_values(&possible_backends)
                 .help("Which image BACKEND to use."),
         )
+        .arg(
+            Arg::with_name("image-colors")
+                .long("image-colors")
+                .value_name("NUM")
+                .takes_value(true)
+                .max_values(1)
+                .possible_values(&["16", "32", "64", "128", "256"])
```
spenserblack, line 125: @CephalonRho Can you confirm that these are the possible values that should be used?

src/onefetch/cli.rs
```diff
@@ -115,6 +116,16 @@ impl Cli {
...
+                .default_value("16")
+                .help("NUM of colors [16, 32, 64, 128, 256] used in image backend."),
```
spenserblack, line 127:
```suggestion
                .help("NUM of colors [16, 32, 64, 128, 256] to use in image backend."),
```
Minor nitpick: this makes the grammar among the option arg help more consistent.

spenserblack (changes requested):
As far as CLI setup goes, looks good to me! :+1: 
Just one request to change the grammar in the help setting.

I'm not too familiar with sixel, so I'll be asking @CephalonRho to look at the `possible_values` setting.

src/onefetch/cli.rs
CephalonRho: Any number above 0 should work. These seem to be chosen arbitrarily, but they are values that should provide a reasonable looking output.
A solution using a [custom validator](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.validator) would be perfect, but it's probably not worth the effort.

CephalonRho (approved):

spenserblack (approved):

Issue #275, opened 2020-10-18
Replace ASCII logo with shell command output


o2sh:
Add a new command-line flag `--ascii `that takes a shell command as input value and renders its output where the ASCII logo is normally displayed, something like:

`onefetch --ascii "$(uname-a)"` or `onefetch --ascii "$(cat main.py)"`

PR #279, opened 2020-10-20
Add ascii flag

KaindlJulian:
Implements the flag described in #275

Example:
image

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -79,7 +79,19 @@ impl Cli {
                 .long("ascii")
                 .value_name("STRING")
                 .takes_value(true)
-                .help("Takes a STRING as input to replace the ASCII logo")
+                .max_values(1)
+                .help("Takes a non-empty STRING as input to replace the ASCII logo.")
+                .long_help("Takes a non-empty STRING as input to replace the ASCII logo. \
+                It is possible to pass a generated STRING by command substitution. \
+                Example: onefetch --ascii \"$(fortune | cowsay -W 25)\"")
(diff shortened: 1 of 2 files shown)


src/onefetch/cli.rs
```diff
@@ -73,6 +74,13 @@ impl Cli {
                         .collect::<Vec<&str>>()
                 ),
         )
+        .arg(
+            Arg::with_name("ascii-input")
+                .long("ascii")
+                .value_name("STRING")
+                .takes_value(true)
```
o2sh, line 81: We probably need to add a `.max_values(1)` and also provide a `.validator` to make sure the ascii-input isn't empty. Indeed, `onefetch --ascii ""` results in:

```
thread 'main' panicked at 'assertion failed: self.end >= self.start', src/onefetch/ascii_art.rs:41:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
And als maybe a `.long_help` with an example: `onefetch --ascii "$(fortune | cowsay -W 25)"`

o2sh (commented):
Thanks a lot for your PR @KaindlJulian 
I left a few suggestions :+1:

src/onefetch/cli.rs
KaindlJulian: Thanks for taking a look, I totally forgot to try out an empty string ^^

The new `.validator` wont allow empty strings. I also added a `.long_help`, but I didn't know if it should have more detail or not since its the only `.long_help` at the moment and I had nothing to compare to.

PR #288, opened 2020-10-22
Add CLI option to disable image

### LINKED ISSUE #287: Add CLI flag to disable ASCII art (or image) and only display info lines
We wish to add a flag `--off` that will result in only showing the info lines (no ASCII logo/image):

`$ > onefetch --off`

off

akrantz01:
Closes #287 

Adds the CLI option `--off` which prevents the image from being displayed.

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -23,6 +23,7 @@ pub struct Cli {
     pub excluded: Vec<String>,
     pub print_languages: bool,
     pub true_color: bool,
+    pub image_off: bool,
 }
 
 impl Cli {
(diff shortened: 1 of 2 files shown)


src/onefetch/cli.rs
```diff
@@ -23,6 +23,7 @@ pub struct Cli {
     pub excluded: Vec<String>,
     pub print_languages: bool,
     pub true_color: bool,
+    pub image_off: bool,
```
o2sh, line 26: As it's not only for images, maybe it would be better to name it `art_off`

src/onefetch/info.rs
```diff
@@ -256,7 +256,11 @@ impl std::fmt::Display for Info {
         let center_pad = "   ";
         let mut info_lines = buf.lines();
 
-        if let Some(custom_image) = &self.config.image {
+        if self.config.image_off {
+            while let Some(line) = info_lines.next() {
+                writeln!(f, "{}", line)?;
+            }
```
o2sh, line 262: I don't think you need to iterate over info_lines, `writeln!(f, "{}\n", buf)?` should be enough

o2sh (commented):
Thanks for your PR @akrantz01,
I left a few suggestions :+1:

src/onefetch/cli.rs
```diff
@@ -172,12 +173,18 @@ impl Cli {
                 .multiple(true)
                 .takes_value(true)
                 .help("Ignore all files & directories matching EXCLUDE."),
+            )
+        .arg(
+            Arg::with_name("off")
+                .long("off")
```
o2sh, line 179: Maybe we should add `.conflicts_with_all(&["image", "ascii-language", "ascii-input"])`

PR #291, opened 2020-10-23
info.rs refactoring

o2sh:
I extracted from info.fmt() the part responsible for merging together the `info_lines` and `logo_lines` into a separate struct: `Printer`

### DIFF
--- a/src/onefetch/cli_utils.rs
+++ b/src/onefetch/cli_utils.rs
@@ -1,0 +1,94 @@
+use crate::onefetch::{ascii_art::AsciiArt, error::*, info::Info, language::Language};
+use colored::Color;
+use std::env;
+use std::io::Write;
+use strum::IntoEnumIterator;
+
+pub struct Printer<W> {
+    writer: W,
+    info: Info,
+}
+
+impl<W: Write> Printer<W> {
+    pub fn new(writer: W, info: Info) -> Self {
+        Self { writer, info }
+    }
+
+    pub fn print(&mut self) -> Result<()> {
+        let center_pad = "   ";
+        let info_str = format!("{}", &self.info);
+        let mut info_lines = info_str.lines();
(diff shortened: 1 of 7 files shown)


src/onefetch/cli_utils.rs
```diff
@@ -0,0 +1,94 @@
...
+                    (Some(logo_line), None) => buf.push_str(&format!("{}\n", logo_line)),
+                    (None, Some(info_line)) => buf.push_str(&format!(
+                        "{:<width$}{}{:^}\n",
+                        "",
+                        center_pad,
+                        info_line,
+                        width = logo_lines.width()
+                    )),
+                    (None, None) => {
+                        buf.push_str("\n");
```
spenserblack, line 57: This might result in a clippy error.
```suggestion
                        buf.push('\n');
```

src/onefetch/cli_utils.rs
```diff
@@ -0,0 +1,94 @@
...
+        let colors: Vec<Color> = Vec::new();
+        let mut buf = String::new();
+        if let Some(custom_image) = &self.info.config.image {
+            if let Some(image_backend) = &self.info.config.image_backend {
+                buf.push_str(&image_backend.add_image(
+                    info_lines.map(|s| format!("{}{}", center_pad, s)).collect(),
+                    custom_image,
+                ));
+            } else {
+                panic!("No image backend found")
```
spenserblack, line 30: Since `Printer::print` returns a `Result`, it might make sense for this `panic!` to be replaced with an `Err`.

spenserblack (approved):
LGTM :+1:
Left some minor notes

src/onefetch/cli_utils.rs
o2sh: I actually managed to remove this else stmt altogether in #289

PR #304, opened 2020-11-01
add dependency feature

Luke-zhang-04:
Add dependency field


Luke-zhang-04:
I've gotten started but I'm not sure if my code will even work. I haven't had the time to test any of it.

My idea is that we can store all the package manager info in a hashmap or something similar (kind of like the `define_languages` macro). I've stolen some code from `license.rs` too.

o2sh:
@Luke-zhang-04, I pushed your implementation a little further; it now compiles.

I think your design works well and can be extended to more package managers without difficulties. A thorough review will be done once the feature is usable (for npm at least).

Luke-zhang-04:
To clarify, an `impl` is the same as a `class` right?

o2sh:
From [here](https://learning-rust.github.io/docs/b5.impls_and_traits.html)

> 💡 When we discussed about C-like structs, I mentioned that those are similar to classes in OOP languages but without their methods. impls are used to define methods for Rust structs and enums.

Luke-zhang-04:
I'll try and get cargo too

o2sh:
The `cargo fmt -- --check` is failing because a `.rustfmt.toml` was added to [`o2sh::master`](https://github.com/o2sh/onefetch) branch; if you merge this branch with yours and `cargo fmt`, you should get a diff.

Luke-zhang-04:
image

o2sh:
Great work @Luke-zhang-04 
With [`369506c`](https://github.com/o2sh/onefetch/pull/304/commits/369506c03c3989962f635a1c6c731cafb40991d5) I tried to improve it a bit with better error handling and file separation.

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    pub fn npm(contents: &str) -> Option<i32> {
+        let parsed = json::parse(contents).unwrap();
+
+        Some(parsed["dependencies"].len() as i32)
+    }
+
+    pub fn pip(contents: &str) -> Option<i32> {
+        let count = Regex::new(r"(^[A-z]+)|(\n[A-z]+)").unwrap().find_iter(contents).count();
```
spenserblack: This looks OK to me, but you might want to iterate over `contents.lines()`, filter to lines matching a regex, then return the count. This allows for a simpler regex, since you only have to check 1 line.

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+        let package_files = fs::read_dir(dir)
+            .chain_err(|| "Could not read directory")?
+            .filter_map(std::result::Result::ok)
+            .map(|entry| entry.path())
+            .filter(|entry| {
+                entry.is_file()
+                    && entry
+                        .file_name()
+                        .map(OsStr::to_string_lossy)
+                        .map(|s| is_package_file(&self, s.as_ref()))
```
spenserblack: Since `is_package_file` ends up taking `&self` as an argument, why not `impl` it directly on `Detector` instead of defining it inside another `fn`?

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+                        .unwrap_or_default()
+            })
+            .filter_map(|entry| {
+                let (parser, package_manager) =
+                    &self.package_managers[entry.file_name().unwrap().to_str().unwrap()];
```
spenserblack: In `is_package_file`, it looks like you're filtering out filenames that don't match any of the keys in the HashMap.
Instead of indexing into `package_managers` directly, you can use `package_managers.get`, which will return `None` if there is no matching key. This should work well with `filter_map`, which filters out `None` values, and also make it so that you don't need to define `is_package_file`.

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+                let contents = fs::read_to_string(entry).unwrap_or_default();
+
+                match parser(&contents) {
+                    Some(number_of_deps) => {
+                        Some(format!("{} ({})", number_of_deps, package_manager))
+                    }
+                    None => None,
+                }
```
spenserblack: Look into [`Option::map`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map).

src/onefetch/info.rs
```diff
@@ -222,6 +232,7 @@ impl Info {
         let last_change = Info::get_date_of_last_commit(&git_history);
         let project_license = Detector::new()?.get_project_license(workdir_str);
         let dominant_language = Language::get_dominant_language(&languages_stats);
+        let dependencies = deps::Detector::new().get_deps_info(workdir_str).unwrap();
```
spenserblack: This should probably be caught instead of unwrapped.

spenserblack (changes requested):
Thanks for your PR! This looks like it could be a pretty cool feature!

Left a few notes.

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+pub struct Detector {
+    package_managers: HashMap<String, (DependencyParser, PackageManager)>,
+}
+
+// Package parsers go here. Parsers should take stirng contents and output a i32
+mod package_parsers {
+    use {regex::Regex, toml::Value};
+
+    pub fn cargo(contents: &str) -> Option<i32> {
+        let parsed = contents.parse::<Value>().unwrap();
```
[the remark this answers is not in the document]
o2sh: Fixed with [`369506c`](https://github.com/o2sh/onefetch/pull/304/commits/369506c03c3989962f635a1c6c731cafb40991d5).
We can maybe add an error message with `chain_err`

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn gomodules(contents: &str) -> Option<i32> {
+        let count = Regex::new(r"v[0-9]+").unwrap().find_iter(contents).count();
+
+        Some(count as i32)
+    }
+
+    pub fn npm(contents: &str) -> Option<i32> {
+        let parsed = json::parse(contents).unwrap();
+
+        Some(parsed["dependencies"].len() as i32)
```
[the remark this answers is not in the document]
o2sh: Also fixed with With [`369506c`](https://github.com/o2sh/onefetch/pull/304/commits/369506c03c3989962f635a1c6c731cafb40991d5)

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+    Pip,
+}
+
+impl std::fmt::Display for PackageManager {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match *self {
+            PackageManager::Cargo => write!(f, "Cargo"),
+            PackageManager::GoModules => write!(f, "Go Modules"),
+            PackageManager::Npm => {
+                if std::path::Path::new("./yarn.lock").exists() {
```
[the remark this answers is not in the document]
Luke-zhang-04: Good point. I'm not sure how I fix this without adding more code. Maybe I should just count it as NPM?

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+                } else {
+                    write!(f, "Npm")
+                }
+            }
+            PackageManager::Pip => write!(f, "Pip"),
+        }
+    }
+}
+
+type DependencyParser = fn(&str) -> Option<i32>;
```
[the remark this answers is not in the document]
Luke-zhang-04: Is usize just an unsigned int? Cause then that would make sense

src/onefetch/deps.rs
```diff
@@ -0,0 +1,122 @@
...
+    use {regex::Regex, toml::Value};
+
+    pub fn cargo(contents: &str) -> Option<i32> {
+        let parsed = contents.parse::<Value>().unwrap();
+
+        Some(parsed["dependencies"].as_table().unwrap().len() as i32)
+    }
+
+    pub fn gomodules(contents: &str) -> Option<i32> {
+        let count = Regex::new(r"v[0-9]+").unwrap().find_iter(contents).count();
```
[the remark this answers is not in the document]
Luke-zhang-04: They usually look like this
```
name

go 1.15

require (
    pkgName v1.0.0
)

File: src/onefetch/deps.rs (hunk 1 above)
Comment (Luke-zhang-04) [reply src/onefetch/deps.rs]: Yeah I'll do that. I didn't know that existed

**Comment by Luke-zhang-04** [comment]
Is there a reason all the requested changes are marked as outdated?

**Comment by o2sh** [comment]
> 
> 
> Is there a reason all the requested changes are marked as outdated?

@spenserblack started his review before my commit in which I moved most of the files where he added his comments.
Except for the two comments regarding catching errors when parsing files, the rest of his remarks still apply.

File: src/onefetch/deps.rs (hunk 9 above)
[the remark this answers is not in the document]
Comment (Luke-zhang-04) [reply src/onefetch/deps.rs]: afc7ef086bcf24687735969d69afa12fa9ef090d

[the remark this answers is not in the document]
Comment (Luke-zhang-04) [reply src/onefetch/deps.rs]: afc7ef0

File: src/onefetch/deps.rs (hunk 1 above)
Comment (Luke-zhang-04) [reply src/onefetch/deps.rs]: Could we just do `(^|\n)[A-z]+`?

File: src/onefetch/info.rs (hunk 5 above)
Comment (Luke-zhang-04) [reply src/onefetch/info.rs]: would that look like
```rust
let dependencies = deps::Detector::new().get_deps_info(workdir_str)?;
```

src/onefetch/deps.rs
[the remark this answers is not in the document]
Luke-zhang-04: 9311d49

src/onefetch/deps.rs
Luke-zhang-04: I changed it in a19ced8 to `(^|\n)[A-z]+`. It seems easier to just do it that way

src/onefetch/info.rs
Luke-zhang-04: d60ecfe

src/onefetch/deps.rs
Luke-zhang-04: I'm not sure I understand. `is_package_file` doesn't take `self` as an argument? If I were to make it am associated function (static method?) I would have to call it with `onefetch::deps::DependencyDetector::is_package_file`, right?

Luke-zhang-04: It's actually a class method :facepalm: so 6fdb61d

src/onefetch/deps.rs
Luke-zhang-04: I think @o2sh fixed this

src/onefetch/deps.rs
Luke-zhang-04: 0b8caa0

Luke-zhang-04:
I think I fixed everything in the review :+1:

Luke-zhang-04:
The new things I added in CONTRIBUTING.md seem like they belong in the wiki. Or I can remove the examples.

src/onefetch/deps.rs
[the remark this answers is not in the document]
spenserblack: >Is usize just an unsigned int?

Yes. More specifically, it is an unsigned int that has a size that depends on the system it is running on. For example, `usize` is 4 bytes on a 32-bit system. It's often used as the return type for sizes of arrays and array-like values.

src/onefetch/deps.rs
spenserblack: That works. [You can also build it as a multi-line regex](https://docs.rs/regex/1.4.2/regex/struct.RegexBuilder.html#method.multi_line) if you want.

src/onefetch/deps/package_parser.rs
```diff
@@ -0,0 +1,26 @@
+use crate::onefetch::error::*;
+use {regex::Regex, toml::Value};
+
+pub fn cargo(contents: &str) -> Result<usize> {
+    let parsed = contents.parse::<Value>()?;
+
+    Ok(parsed["dependencies"].as_table().unwrap().len())
```
spenserblack, line 7: A `Cargo.toml` file is valid even if it doesn't have a `[dependencies]` key, so could you use [`toml::Value::get`](https://docs.rs/toml/0.5.7/toml/value/enum.Value.html#method.get), returning 0 for the count if `None` is returned?

src/onefetch/deps/package_parser.rs
```diff
@@ -0,0 +1,26 @@
...
+pub fn go_modules(contents: &str) -> Result<usize> {
+    let count = Regex::new(r"v[0-9]+")?.find_iter(contents).count();
+
+    Ok(count)
+}
+
+pub fn npm(contents: &str) -> Result<usize> {
+    let parsed = json::parse(contents)?;
+
+    Ok(parsed["dependencies"].len())
```
spenserblack, line 19: See comment on `fn cargo`.

spenserblack (approved):
Thanks for making all those changes! And thanks for updating the contributing guidelines!

I'm approving this overall, but I left a couple notes to catch edge-cases when the `dependencies` key isn't defined.

o2sh (approved):

src/onefetch/deps/package_parser.rs
Luke-zhang-04: Oh yeah, good point. I didn't think of that.

src/onefetch/deps/package_parser.rs
Luke-zhang-04: I just tried. At least for NPM, it gives a 0 by default.

src/onefetch/deps/package_parser.rs
Luke-zhang-04: 1d6d95b

Luke-zhang-04:
I fixed the no dependency thing in 1d6d95b but I don't think I did it in the best way. I'm not sure though.

Luke-zhang-04:
Great. I just looked at https://github.com/kubernetes/kubernetes/blob/master/go.mod and there's a `replace` directive. I guess I'll have to redo the go modules one.

Luke-zhang-04:
I think we're good to merge

o2sh:
Thanks a lot for fixing the `go module` parser :pray:

o2sh:
The go module parser was still a little bugged, this seems to work:

```rust
pub fn go_modules(contents: &str) -> Result<usize> {
    let mut count = 0;
    let mut start = false;
    for line in contents.lines() {
        if line.contains("require") {
            start = true;
            continue;
        }

        if start && line.contains(')') {
            break;
        }

        if start {
            count += 1;
        }
    }

    Ok(count)
}
```

Issue #283, opened 2020-10-21
Extend image backend support to ITerm2 Inline Images Protocol.


o2sh:
As mentioned by @yoichi in #266, even though iterm2 seems to support SIXEL; it shows a broken output:

![image](https://user-images.githubusercontent.com/13710835/96747802-077e3d00-13c9-11eb-9148-44b2f98abbc0.png)

Apparently, iterm2 has its own [rendering protocol](https://iterm2.com/documentation-images.html) which _extends the xterm protocol with a set of proprietary escape sequences_. This xterm protocol must be the SIXEL protocol - Implemented by @CephalonRho - or smth close to it.

mborejdo:
Looking forward to this. https://github.com/wez/wezterm is an awesome terminal-emulator with great Windows support. It also supports the iTerm2 inline image protocol and it would be great to have some windows love :-)

o2sh:
From what I understood, @yoichi opened an [issue](https://gitlab.com/gnachman/iterm2/-/issues/9204) addressing this problem which was fixed int this [commit](https://gitlab.com/gnachman/iterm2/-/commit/119da0c28245d797dc124990676acfdd67789cfb)

Unfortunately, I don't have a machine running macOS so I can't test it.

mborejdo:
Hey @o2sh. thanks for your reply. I am not sure I understand. Does this mean the iterm2 inline image support (which works also on windows on wezterm) will not be added, once the sixel-issue is resolved?

o2sh:
@mborejdo, wezterm also supports Sixel

![wezterm](https://user-images.githubusercontent.com/13710835/97725766-d390cf00-1ace-11eb-8b09-9ff3a444fbe4.png)

However, this won't work on Windows.

mborejdo:
hey @o2sh  . Yes, thats really unfortunate. Thats why I was hoping to have the iterm2-inline-image working on windows as this works as expected (almost)
Also unfortunately onefetch's --ascii does not help here, as the commandline is limited in length. Would it be possible to provide something like --ascii-file or the like so I could pass the base64 encoded inline-image to onefetch to be displayed instead of the default?

o2sh:
> Also unfortunately onefetch's --ascii does not help here, as the commandline is limited in length.

Try this `onefetch —ascii-input "$(cat image_base64.txt)"`

PR #305, opened 2020-11-01
Add iTerm2 image protocol support

yoichi:
Fix https://github.com/o2sh/onefetch/issues/283

スクリーンショット 2020-11-01 22 57 58

### DIFF
--- a/src/onefetch/image_backends/iterm2.rs
+++ b/src/onefetch/image_backends/iterm2.rs
@@ -1,8 +1,6 @@
 use {
     image::{imageops::FilterType, DynamicImage, GenericImageView},
-    libc::{
-        ioctl, winsize, STDOUT_FILENO, TIOCGWINSZ,
-    },
+    libc::{ioctl, winsize, STDOUT_FILENO, TIOCGWINSZ},
     std::env,
 };
 
@@ -39,7 +37,9 @@ impl super::ImageBackend for ITerm2Backend {
         let image_rows = height_ratio * f64::from(image.height());
 
         let mut bytes: Vec<u8> = Vec::new();
-        image.write_to(&mut bytes, image::ImageOutputFormat::Png).unwrap();
+        image
+            .write_to(&mut bytes, image::ImageOutputFormat::Png)
+            .unwrap();
         let encoded_image = base64::encode(bytes);
         let mut image_data = Vec::<u8>::new();
 
--- a/src/onefetch/image_backends/iterm2.rs
+++ b/src/onefetch/image_backends/iterm2.rs
@@ -1,0 +1,61 @@
+use {
+    image::{imageops::FilterType, DynamicImage, GenericImageView},
+    libc::{
+        ioctl, winsize, STDOUT_FILENO, TIOCGWINSZ,
+    },
+    std::env,
+};
+
+pub struct ITerm2Backend {}
+
+impl ITerm2Backend {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn supported() -> bool {
+        let term_program = env::var("TERM_PROGRAM").unwrap_or("".to_string());
+        term_program == "iTerm.app"
+    }
+}
+
+impl super::ImageBackend for ITerm2Backend {
+    fn add_image(&self, lines: Vec<String>, image: &DynamicImage, _colors: usize) -> String {
+        let tty_size = unsafe {
+            let tty_size: winsize = std::mem::zeroed();
+            ioctl(STDOUT_FILENO, TIOCGWINSZ, &tty_size);
+            tty_size
+        };
+        let width_ratio = f64::from(tty_size.ws_col) / f64::from(tty_size.ws_xpixel);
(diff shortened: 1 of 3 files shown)


src/onefetch/image_backends/iterm2.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct ITerm2Backend {}
+
+impl ITerm2Backend {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    pub fn supported() -> bool {
+        let term_program = env::var("TERM_PROGRAM").unwrap_or("".to_string());
+        term_program == "iTerm.app"
```
o2sh, line 18: What about other terminals that supports Iterm2 image protocol, like WezTerm ?

src/onefetch/image_backends/mod.rs
```diff
@@ -32,6 +36,11 @@ pub fn check_if_supported(backend_name: &str) -> Result<()> {
                 return Err("Kitty image backend is not supported".into());
             }
         }
+        "iterm2" => {
+            if !iterm2::ITerm2Backend::supported() {
+                return Err("Sixel image backend is not supported".into());
```
o2sh, line 41: typo: Sixel instead of Iterm2

src/onefetch/image_backends/iterm2.rs
yoichi: WezTerm sets TERM_PROGRAM=WezTerm
It supports iTerm2 image protocol, but it behaves a bit differently from iTerm2:
スクリーンショット 2020-11-02 20 54 30

o2sh: if you do this
```rust
image_data.extend(format!("\x1B[{}A", image_rows as u32).as_bytes()); // move cursor to top-left corner
image_data.extend(format!("\x1B[{}C", image_columns as u32 + 1).as_bytes()); // move cursor to top-right corner of image
```
instead of 
```rust
image_data.extend(format!("\x1B[{}A", image_rows as u32 - 1).as_bytes()); // move cursor to start of image`
```
you get the correct input:

iterm

Hopefully, it doesn't break the output on Iterm2 :worried:

yoichi: > Still, my first remark remains relevant: we shouldn't assume that Iterm2 will be the only terminal supporting this protocol.

I think we cannot detect all terminals supporting iTerm2 image protocol. 
So I think we should not prohibit them in image_backends::check_if_supported().
Fixed by d821bc35

yoichi: > Logically, this should break the output on Iterm2, in that case - @yoichi plz test it -, I'll open an issue on WezTerm's Github.

I've tested with the change

```
diff --git a/src/onefetch/image_backends/iterm2.rs b/src/onefetch/image_backends/iterm2.rs
index 0c2f3c7..86dc2cf 100644
--- a/src/onefetch/image_backends/iterm2.rs
+++ b/src/onefetch/image_backends/iterm2.rs
@@ -33,7 +33,7 @@ impl super::ImageBackend for ITerm2Backend {
             (lines.len() as f64 / height_ratio) as u32,
             FilterType::Lanczos3,
         );
-        let _image_columns = width_ratio * f64::from(image.width());
+        let image_columns = width_ratio * f64::from(image.width());
         let image_rows = height_ratio * f64::from(image.height());
 
         let mut bytes: Vec<u8> = Vec::new();
@@ -47,7 +47,8 @@ impl super::ImageBackend for ITerm2Backend {
         image_data.extend(encoded_image.bytes());
         image_data.extend(b"\x07");
 
-        image_data.extend(format!("\x1B[{}A", image_rows as u32 - 1).as_bytes()); // move cursor to start of image
+        image_data.extend(format!("\x1B[{}A", image_rows as u32).as_bytes()); // move cursor to top-left corner
+        image_data.extend(format!("\x1B[{}C", image_columns as u32 + 1).as_bytes()); // move cursor to top-right corner of image
         let mut i = 0;
         for line in &lines {
             image_data.extend(format!("\x1B[s{}\x1B[u\x1B[1B", line).as_bytes());
```

fixes the problem on WezTerm, and cause a problem on iTerm2:

スクリーンショット 2020-11-03 0 43 45

o2sh: Thanks @yoichi, I created the issue --> https://github.com/wez/wezterm/issues/317

CephalonRho (approved):
I can't test this due to not having a macOS machine available and being unable to get wezterm to start, but from what little I know about the iTerm2 image protocol, this seems fine to me.

It only detects the terminal being used by reading `TERM_PROGRAM`, not the support for the protocol itself, but there doesn't appear to be a way around that right now.

o2sh (approved):

wez:
> I can't test this due to not having a macOS machine available and being unable to get wezterm to start,

I'd be happy to troubleshoot this with you if you have the time!

PR #309, opened 2020-11-03
Remove async/await

o2sh:
## With async/await on the linux repo:

```bash
#Response times for async calls
o2sh ~/dev/onefetch (telemetry) > onefetch ../linux/
get_number_of_tags_branches --> 40.105318ms
get_git_version_and_username --> 39.971554ms
get_packed_size --> 89.731579ms
get_version --> 189.732919ms
get_pending_changes --> 199.959989ms
get_git_history --> 18.836888247s
```

`get_git_history` is overwhelmingly slower than the other functions;even with async/await the execution is still very synchronous-like.

```bash
time onefetch ../linux

real	0m22,442s
user	0m48,883s
sys	0m6,863s
```
## Without async/await on the linux repo:

```bash
time onefetch ../linux

real	0m15,631s
user	0m22,745s
sys	0m4,017s
```
-------
It may have been a false good idea to add tokio runtime - with its huge added complexity - to onefetch. Especially considering the use case - only one very slow function `get_git_history` -. Without async/await, onefetch appears to compile and run faster and use less cpu resources (cf. `user` time).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,12 +58,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arc-swap"
-version = "0.4.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d25d88fd6b8041580a654f9d0c581a047baee2b3efee13275f2fc392fc75034"
-
-[[package]]
 name = "arrayref"
 version = "0.3.6"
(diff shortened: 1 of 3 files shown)


spenserblack (approved):
LGTM :+1:

o2sh:
Thanks @spenserblack

PR #312, opened 2020-11-07
Change Nix ASCII icon

balsoft:
Old NixOS icon is a bit messy and not very recognizeable.

Take the ASCII icon for Nix from https://github.com/dylanaraps/neofetch/blob/56c8d7e0c81108163a3f2970940ea2287648916f/neofetch#L8519

I don't know if that icon can be considered a "substantial" portion of
neofetch, so just to be safe I also included the copyright notice (license
is MIT).


balsoft:
Ah, somehow I expected the 45 character limit to be applied to the rendered version and not the raw version.

spenserblack:
>I expected the 45 character limit to be applied to the rendered version and not the raw version.

In this case, it is testing the rendered version (which has a length of 43 at line 9), but the issue is that [the max width is actually 40](https://github.com/o2sh/onefetch/blame/v2.7.1/src/onefetch/language.rs#L134), even though CONTRIBUTING.md says it's 45.

Would you like to submit a PR to fix this? 😺

o2sh:
> but the issue is that the max width is actually 40, even though CONTRIBUTING.md says it's 45.

I'm so sorry :frowning:, fixed in 585a51d

PR #315, opened 2020-11-10
Replace git sys calls with git2-rs (libgit2) 

### LINKED ISSUE #310: Replace git sys calls with git2-rs
Currently, we manage to retrieve Git repository information via two means:

* Git sys calls --> `std::process::Command::new("git").`..
* [`git2-rs`](https://github.com/rust-lang/git2-rs) -->  `git2::Repository::::discover(&config.path)`,

In order to homogenize the code and not rely on the Git binary to be present on the user's machine, we wish to update the following methods:

- get_packed_size
- get_git_history

to switch from Git sys calls to git2-rs function calls instead.

o2sh:
Functions migrated so far:

* `get_pending_changes`
* `get_version`
* `get_git_username`
* `get_number_of_tags_branches`

-----
#310


src/onefetch/repo.rs
```diff
@@ -0,0 +1,168 @@
...
+        for e in statuses.iter() {
+            let s: Status = e.status();
+            if s.is_index_new() || s.is_wt_new() {
+                added += 1;
+            } else if s.is_index_deleted() || s.is_wt_deleted() {
+                deleted += 1;
+            } else {
+                modified += 1;
+            }
+        }
```
spenserblack, line 87: This isn't necessarily an *improvement*, just a different way to do the same thing. Totally up to your preference for which gets implemented :slightly_smiling_face:
```suggestion
        let (added, deleted, modified) = statuses.iter().fold((0, 0, 0), |(added, deleted, modified), e| {
            let s: Status = e.status();
            if s.is_index_new() || s.is_wt_new() {
                (added + 1, deleted, modified)
            } else if s.is_index_deleted() || s.is_wt_deleted() {
                (added, deleted + 1, modified)
            } else {
                (added, deleted, modified + 1)
            }
        });
```

src/onefetch/repo.rs
```diff
@@ -0,0 +1,168 @@
...
+            result = format!("{}+-", modified)
+        }
+
+        if added > 0 {
+            result = format!("{} {}+", result, added);
+        }
+
+        if deleted > 0 {
+            result = format!("{} {}-", result, deleted);
+        }
```
spenserblack, line 100: Is it possible for at least two of modifed, added, and deleted to have a value > 0? If `added > 0 && deleted > 0`, won't this cause the result to only include the deleted lines?

spenserblack (approved):
I probably need to familiarize myself with git2 more, but looks good to me! :+1:
Just had some minor questions and suggestions.

Also, did you notice any significant performance difference?

o2sh:
> Also, did you notice any significant performance difference?

I'll compare response times after all the functions relying on git sys calls are updated. As of now, I still have to work on:

-  get_git_version: This one might be tricky, even impossible.
-  get_packed_size
-  get_git_history

src/onefetch/repo.rs
o2sh: I love it :heart_eyes_cat:

src/onefetch/repo.rs
o2sh: I'm not sure to get it...at each step we **append** to the previous value of `result` either added or deleted.

src/onefetch/repo.rs
```diff
@@ -0,0 +1,168 @@
...
+            Err("invalid workdir".into())
+        }
+    }
+
+    pub fn get_number_of_tags(&self) -> Result<usize> {
+        Ok(self.repo.tag_names(None)?.len())
+    }
+
+    pub fn get_number_of_branches(&self) -> Result<usize> {
+        Ok(self.repo.branches(Some(BranchType::Remote))?.count() - 1)
```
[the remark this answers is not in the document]
o2sh: I am pretty certain you already made me this remark in the past :sweat:

src/onefetch/repo.rs
spenserblack: This might be behavior that's currently in `master`, too.

```bash
# in onefetch repo @ master
mkdir test-repo
cd test-repo
git init
echo print 1 > test.py
echo print 2 >> test.py
echo print 3 >> test.py
echo print 4 >> test.py
echo print 5 >> test.py
git add --all
git commit -m "Init"
echo print 0 > test.py
echo print 1 >> test.py
echo print 3  >> test.py
echo print 4 >> test.py
echo print 51 >> test.py
cd -
cargo run -- test-repo
```

Onefetch will report **Pending**: 1+-
(one modification)

however,
```bash
cd test-repo
git diff
```
Reports
```diff
+print 0
 print 1
-print 2
 print 3
 print 4
-print 5
+print 51
```
Which is 1 addition, 1 deletion, and 1 modification.

Does that make sense? Am I interpreting how Onefetch is supposed to display pending changes properly?

spenserblack: Oh, nevermind, I see that it's a count of pending *file* changes, not line changes :facepalm:
Please ignore :laughing:

o2sh:
...I give up :disappointed: I moved the remaining function relying on Git sys calls into a separate file `git_utils.rs`.

The performance remains identical as I didn't touch `get_git_history` which is the slowest function.

To be continued

Issue #321, opened 2020-11-21
Support for vimscript


geeseven:
Hey @o2sh,

I am toying with adding support for [vimscript][0].  It is supported by tokei, but they don't really have a proper logo.  The website has some styled text, but the size does not fit with this project.

```

       .                        .            
       :                        -            
       :                        -        --  
       :                        -        --  
                                         --  
                                         --  
 :   ::: :::.::  ----  ---- -----  ---- -----
 :   ::: ::::::: ---- ----- ----- ----- -----
 :   :.: ::::::: ---- ----------- ----- -----
 :   : : ::::.::_---- ----------- ----- -----
 :   : : :  : .::---_ -    --   - -_  -. --  
 :  :: : :  : .: ---- -    --   - -_  -. --  
 :  :: : :  : .: ---- -    --   - -_  -. --  
 :  :. : :  : .: ---- -    --   - -_  -. --  
 : ::  : :  : .:    - -    --   - -_  -. --  
 ::::  : :  : .:.---- -------   - -_---  --  
 :::   : :  : .:.---- -------   - -_---  --  
 :::   : :  : .:.---- -------   - -_---  --  
 ::    : :  : .:.---.  ------   - -_--:  --  
                                  -_         
                                  -_         
                                  -_         
                                  -_         
```

I tried stacking the words and it still does not look that great.

```
                    ::                       
                    ::                       
                                             
           ::    ::.:: ::::::::::            
           ::    :: :: :::::::::::           
           ::   ::: :: ::   ::  ::           
           ::  :::  :: ::   ::  ::           
           :: :::   :: ::   ::  ::           
           :::::    :: ::   ::  ::           
           :::      :: ::   ::  ::           
                                             
                         ..                  
                         --            ---   
                                       ---   
  -----   ------_  ----- --  :-----  --------
 ------  -------_ ------ -- _------- --------
 -----:  --      ---     -- ---   ---  ---   
 ------- --      ---     -- --_   ---  ---   
      -- --      ---     -- --_   ---  ---   
 ------- -------_---     -- --_------  ---   
 ------:  ------_---     -- --_-----   ---   
                            --_              
                            --_     
```

Should we just use Vim's logo?

```
                                .                     
     ______________  ...:   -_____________-  
     ______________: ...... ._____________:  
       -________:   .......   -_________::   
       -________:  ........ --_________:     
       -________:  ...... _-_________:       
       -________:  ....  -_________:         
       -________:  ..  - ________.:          
[...]

PR #322, opened 2020-11-22
add support for VimScript #321

geeseven:
Let's add some VimScript support.

VimScript


o2sh:
Beautiful 😍 
Thanks @geeseven

PR #324, opened 2020-11-23
add support for Vala

geeseven:
Vala anyone?

Vala


o2sh:
Very nice design @geeseven 
I tried to clean it a bit:

vala

PR #330, opened 2020-11-27
feat: hide logo if terminal size is too wide

Luke-zhang-04:
a new CLI option to configure whether the logo should be shown
if configued to auto, Onefetch will detect the terminal size and
hide/show the logo based off that.

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -222,7 +222,7 @@ impl Cli {
         .arg(
             Arg::with_name("off")
                 .long("off")
-                .help("Only shows the info lines. DEPRECATED: use \"--hide-logo\" true instead.")
+                .help("Only shows the info lines. DEPRECATED: use \"--hide-logo true\" instead.")
                 .conflicts_with_all(&["image", "ascii-language", "ascii-input", "hide-logo"]),
             )
         .arg(
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -7,9 +7,11 @@ use {
         language::Language,
     },
     clap::{crate_description, crate_name, crate_version, App, AppSettings, Arg},
+    colored::Colorize,
     image::DynamicImage,
     std::{convert::From, env, str::FromStr},
     strum::IntoEnumIterator,
+    term_size,
 };
 
 pub struct Cli {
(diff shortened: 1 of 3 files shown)


Luke-zhang-04:
Ok it works but the code is pretty garbage

Luke-zhang-04:
image

src/onefetch/cli.rs
```diff
@@ -220,17 +222,37 @@ impl Cli {
...
         let print_languages = matches.is_present("languages");
         let print_package_managers = matches.is_present("package-managers");
-        let art_off = matches.is_present("off");
+        let mut art_off = matches.is_present("off");
         let true_color = cli_utils::is_truecolor_terminal();
+        let max_term_size = 95;
+
+        // The --off flag was passed in
+        if art_off {
+            std::println!(
```
spenserblack, line 251: Just curious: were you having an issue with `println!` (no `std::`)?

src/onefetch/cli.rs
```diff
@@ -255,6 +277,25 @@ impl Cli {
             None
         };
 
+        if let Some(should_hide_logo) = matches.value_of("hide-logo").or(Some("auto")) {
+            if should_hide_logo == "true" {
```
spenserblack, line 281: Not a necessary change, just a fun shortcut I thought of 🙂 
```rust
if let Ok(b) = should_hide_logo.parse() {
    art_off = b; // parses should_hide_logo to boolean, which handles both "true" and "false" cases.
} else if !art_off {
    // etc.
}
```

src/onefetch/cli.rs
```diff
@@ -255,6 +277,25 @@ impl Cli {
...
+                art_off = true;
+            } else if should_hide_logo == "false" {
+                art_off = false;
+            } else if !art_off {
+                if let Some((width, _)) = term_size::dimensions_stdout() {
+                    art_off = width <= max_term_size;
+                } else {
+                    std::println!(
```
spenserblack, line 289:
```suggestion
                    std::eprintln!(
```
Prints to STDERR instead of STDOUT. Might be prefer if a user wants to do something like `onefetch 2> error.log` to hide error messages in the terminal.
Also, see question above about using `std::`.

o2sh:
I really don't see the point of having the three options (true|false|auto), if the user provides the `--hide-logo` it should be enough to decide whether to hide the ascii logo (or image). Only the `auto` option is important IMO.

Luke-zhang-04:
Alright. I was worried about backwards compatibility but if I can get rid of it then that's fine too.

src/onefetch/cli.rs
Luke-zhang-04: I don't know any other way to do it so..

src/onefetch/cli.rs
Luke-zhang-04: Alright

Luke-zhang-04:
Ok so I tried something a bit different. Hopefully you like it. But it seems more intuitive for me to have the auto option on by default :man_shrugging:

o2sh:
I see you went from one extreme to the other :sweat_smile: I think we should still allow the user to force `art_off `to true no matter his/her terminal's width and have the `auto` option to automatically hide it.

Luke-zhang-04:
G r e a t.

Luke-zhang-04:
Oh, I see what you mean. So without flags, it should hide the logo depending on width. But if the flag is passed, it will always hide the logo.

o2sh:
With 51a7a4b and 2979a77, I may have backtracked a bit. We're back to two options for `--hide-logo` (always and auto) with a default to always.

Luke-zhang-04:
OK I guess that's fine too

o2sh:
I'm very sorry for the confusion :disappointed:
Again, thanks a lot for your contribution @Luke-zhang-04

Issue #302, opened 2020-10-28
Add JSON output support


joseluisq:
It would be great if `onefetch` could share its stats like `Project`, `Authors`, etc in JSON format. Probably a `--json` flag for that.
I also think that [serde](https://crates.io/crates/serde) can help here.

nguyenquannnn:
Hey @o2sh can I take a dig at this?

HallerPatrick:
Hey, after there is not much progress on this issue, I just dived into this feature. 
So far I have a running implementation which gives following output to stdout:

```json
{
  "git_username": "Patrick Haller",
  "git_version": "git version 2.28.0",
  "repo_name": "onefetch",
  "number_of_tags": 23,
  "number_of_branches": 2,
  "head_refs": {
    "refs": [
      "master",
      "origin/master"
    ]
  },
  "pending_changes": "10+- 1+",
  "version": "v2.7.3",
  "creation_date": "2 years, 3 months ago",
  "languages": [
    [
      "Rust",
      96.36507309363887
    ],
    [
      "Python",
      2.8447254049782695
    ],
    [
      "Markdown",
      0.7902015013828527
    ]
  ],
  "dependencies": "16 (cargo)",
  "authors": [
    [
      "Ossama Hjaji",
      589,
      57
    ],
    [
      "dependabot[bot]",
      66,
      6
    ],
    [
      "Spenser Black",
      49,
      4
    ]
  ],
  "last_change": "5 hours ago",
  "repo_url": "https://github.com/o2sh/onefetch.git",
  "number_of_commits": "1021",
  "lines_of_code": 2511,
  "repo_size": "14.48 MiB (110 files)",
  "license": "MIT",
  "dominant_language": "Rust",
  "text_colors": {
    "title": {
      "TrueColor": {
        "r": 228,
        "g": 55,
        "b": 23
      }
    },
    "tilde": "White",
    "underline": "White",
    "subtitle": {
      "TrueColor": {
        "r": 228,
        "g": 55,
        "b": 23
      }
    },
    "colon": "White",
    "info": "White"
  }
}
```

I used the mentioned serde lib. This output represents the Info struct, but is omitting the used ascii colours and the cli struct.

What do you think?

spenserblack:
Looks awesome! :smiley:

But it would probably be best if the JSON isn't a 1:1 of the Info struct.

***

[...]

PR #337, opened 2020-12-08
Add JSON Output Support

HallerPatrick:
Issue: #302

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -31,7 +31,7 @@ pub struct Cli {
     pub excluded: Vec<String>,
     pub print_languages: bool,
     pub print_package_managers: bool,
-    pub print_json: bool,
+    pub format: String,
     pub true_color: bool,
     pub art_off: bool,
     pub text_colors: Vec<String>,
@@ -230,10 +230,19 @@ impl Cli {
                 .help("Ignore all files & directories matching EXCLUDE."),
             )
         .arg(
-            Arg::with_name("json")
-                .short("j")
-                .long("json")
-                .help("Print out a json representation of the normal onefetch output")
+            Arg::with_name("format")
+                .short("f")
+                .long("format")
+                .help("Select a output format.")
+                .takes_value(true)
+                .default_value("human")
+                .validator(
+                    |t| {
+                        let opts = vec!["human", "json"];
+                        if opts.contains(&t.as_str()) {return Ok(()); }
+                        Err(String::from(format!("Output type must be one of the following: {:?}", opts)))
+                    }
+                )
             )
 .get_matches();
 
@@ -243,7 +252,11 @@ impl Cli {
         let no_color_palette = matches.is_present("no-color-palette");
         let print_languages = matches.is_present("languages");
         let print_package_managers = matches.is_present("package-managers");
-        let print_json = matches.is_present("json");
+        let format = if let Some(format) = matches.value_of("format") {
+            String::from(format)
+        } else {
+            String::from("human")
+        };
 
         let fields_to_hide: Vec<String> = if let Some(values) = matches.values_of("disable-fields")
         {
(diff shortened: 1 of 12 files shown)


src/onefetch/cli.rs
```diff
@@ -228,6 +229,12 @@ impl Cli {
                 .takes_value(true)
                 .help("Ignore all files & directories matching EXCLUDE."),
             )
+        .arg(
+            Arg::with_name("json")
+                .short("j")
+                .long("json")
+                .help("Print out a json representation of the normal onefetch output")
+            )
```
spenserblack, line 237: It might be better as `--format=json` instead of `--json`. The `--format=` option could then be reused if other output formats are supported later. If the `format` option isn't used, it could default to "human" output.

src/onefetch/info.rs
```diff
@@ -23,6 +25,8 @@ pub struct Info {
     repo_url: String,
     number_of_commits: String,
     lines_of_code: usize,
+    packed_repo_size: String,
+    files_count: Option<u128>,
```
spenserblack, line 29: I'm curious why you decided to use `u128`. Will `u64::MAX` or `usize::MAX` to be insufficient?

src/onefetch/git_utils.rs
```diff
@@ -74,7 +74,20 @@ pub fn get_number_of_commits(git_history: &[String]) -> String {
     number_of_commits.to_string()
 }
 
-pub fn get_packed_size(dir: &str) -> Result<String> {
+pub fn get_packed_size(repo_size: String, files_count: Option<u128>) -> Result<String> {
+    match files_count {
+        Some(files_count) => {
+            let res = repo_size + (" (") + &(files_count.to_string()) + (" files)");
```
spenserblack, line 80: Instead of string addition, could you use the `format!` macro? I think it's more readable.

spenserblack (commented):
Thanks for your work on this!

Left a few notes.

src/onefetch/info.rs
HallerPatrick: Because `u128` was already used for line counts. But you are right that u64 should be sufficient (any code base that would cause an u64 overflow deserves it ;) ).

src/onefetch/git_utils.rs
HallerPatrick: I can change that!

src/onefetch/cli.rs
HallerPatrick: Sounds like a reasonable proposal, will change that.

src/onefetch/cli.rs
```diff
@@ -228,6 +229,21 @@ impl Cli {
...
+                .help("Select a output format.")
+                .takes_value(true)
+                .default_value("human")
+                .validator(
+                    |t| {
+                        let opts = vec!["human", "json"];
+                        if opts.contains(&t.as_str()) {return Ok(()); }
+                        Err(String::from(format!("Output type must be one of the following: {:?}", opts)))
+                    }
+                )
```
spenserblack, line 245: Instead of a validator, you might want to use `.possible_values`, which should accomplish roughly the same thing.

src/onefetch/cli.rs
```diff
@@ -236,6 +252,11 @@ impl Cli {
         let no_color_palette = matches.is_present("no-color-palette");
         let print_languages = matches.is_present("languages");
         let print_package_managers = matches.is_present("package-managers");
+        let format = if let Some(format) = matches.value_of("format") {
+            String::from(format)
+        } else {
+            String::from("human")
+        };
```
spenserblack, line 259: Since "format" has a default value, I think `matches.value_of("format")` will *always* be `Some`.
```suggestion
        let format = matches.value_of("format").map(String::from).unwrap();
```

HallerPatrick:
This Review feature is amazing! Thank you for your time to look over my code.

spenserblack (approved):
Thank you for your hard work, @HallerPatrick!

joseluisq:
Awesome! :+1:

Issue #338, opened 2020-12-09
Misleading Size Label


HallerPatrick:
While working on #310 I stumbled on `get_repo_size` which uses a the os git command  `git -C <dir> count-objects -vH` to determine the *current* size of the git repo.

Il explain why this is misleading:

|                        | onefetch w/o `.git` |  `.git`             | Total |
| ------------- | -------------           |------------- | ------------- |
| Size                |  1.8 MiB                    |  15M              | 17M  |
| File Count     | 111                            | 25                  | 136  |

The git command also includes the size of `.git`, while the file count excludes `.git`. 

So we should either fix the repo size or fix the file count or add one of the others as additional information.

o2sh:
Thanks for pointing that out @HallerPatrick.
Personally, I don't mind excluding `.git` folder when reading the `repo size`. Especially, if it allows to have more coherent values in regard to the `file count`.

HallerPatrick:
I am doing the migration to git2 and working on this. If it is okay I would change it on the fly with Issue #310 @o2sh ?

o2sh:
Sure, that would be great :+1:

subnut:
@HallerPatrick Is this the reason of #342 ?
FYI, there are 15 files, and the size excluding `.git` is definitely not 0 bytes

HallerPatrick:
> @HallerPatrick Is this the reason of #342 ?
> FYI, there are 15 files, and the size excluding `.git` is definitely not 0 bytes

Its probably not the reason.

PR #341, opened 2020-12-10
https://github.com/o2sh/onefetch/issues/310

HallerPatrick:
Resolve Issue #310 and #338 

Please review, because this includes a lot of changes. Sorry for that.


spenserblack (commented):
Looks awesome! Thanks for your hard work!

o2sh:
Very impressive @HallerPatrick,

I tried running your branch on a few repos, everything seems to work except for the [linux](https://github.com/torvalds/linux) repo where  I got  a:

```
hread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/onefetch/git_utils.rs:40:45
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
if failed here: `let author_name = author.name().unwrap().to_string();` I don't see how a commit can exist without an author name...anyway, I fixed it locally and it worked just fine.

However, the response time has increased quite a bit, still on the linux repo, we went from:

```sh
$ time onefetch_master >

real	0m15,742s
user	0m21,599s
sys	0m3,985s
```

to:
```sh
$ time onefetch >

real	0m28,633s
user	0m33,951s
sys	0m1,097s
```
----------------
By the way, the `git_utils.rs` file was created to isolate functions that were still relying on git sys calls, now everything can be moved to the `repo.rs` file.

spenserblack:
>if failed here: `let author_name = author.name().unwrap().to_string();`

[`git2::Signature::name`](https://docs.rs/git2/0.13.12/git2/struct.Signature.html#method.name) returns `None` if the name isn't valid UTF-8. Perhaps `name_bytes` should be used instead (and also `email` ➡️  `email_bytes`)?

src/onefetch/git_utils.rs
```diff
@@ -1,134 +1,132 @@
...
+    pub fn get_repo_size(&self) -> (String, u64) {
+        let (repo_size, file_count) = match self.repo.index() {
+            Ok(index) => {
+                let mut repo_size: u128 = 0;
+                let mut file_count: u64 = 0;
+                for index_entry in index.iter() {
+                    file_count += 1;
+                    repo_size += index_entry.file_size as u128;
+                }
+                (repo_size, file_count)
```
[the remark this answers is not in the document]
HallerPatrick: Looks more idiomatic! Also `folds`are really cool. Had to fix it a little bit up. With type annotations and syntactically

HallerPatrick:
> Very impressive @HallerPatrick,
> 
> I tried running your branch on a few repos, everything seems to work fine except for the [linux](https://github.com/torvalds/linux) repo where I got a:
> 
> ```
> hread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/onefetch/git_utils.rs:40:45
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> if failed here: `let author_name = author.name().unwrap().to_string();` I don't see how a commit can exist without an author name and why it works when using git sys calls..but anyway, I fixed it locally and it worked just fine.

Alright also fixed the Signature utf-8 problem with using `from_utf8_lossy` which should return � for invalid utf-8 chars. 


> 
> However, the response time has increased quite a bit, still on the linux repo, we went from:
> 
> ```shell
> $ time onefetch_master >
> 
> real	0m15,742s
> user	0m21,599s
> sys	0m3,985s
> ```
> 
> to:
> 
> ```shell
> $ time onefetch >
> 
> real	0m28,633s
> user	0m33,951s
> sys	0m1,097s
> ```

I also tested this now with [Hyperfine](https://github.com/sharkdp/hyperfine) (which seems to be a good tool) and had the same results. Its interesting that using git sys calls seem to be much faster than using git2. Maybe commands like `object-count` are more efficiently implemented in the git cli tools, then I did.

I definitely wanted to dig deeper into this. 

> 
> By the way, the `git_utils.rs` file was created to isolate functions that were still relying on git sys calls, now everything can be moved into the `repo.rs` file.

I would probably move this to another pull requests to keep the diff small and separation of purpose.  If that's okay.

o2sh:
> Maybe commands like object-count are more efficiently implemented in the git cli tools.

It's the `get_git_history` function that takes most of the response time, on the linux repo:

before:

```
get_git_history --> 11.77949631s
get_files_count --> 18.192162ms
get_repo_size --> 1.709682ms
```

now:

```
get_git_history --> 25.771483865s
get_repo_size --> 49.508897ms
```

HallerPatrick:
@o2sh Yeah , I think we reached a little bit of a dead end with git2. With the sys call we just used the logs to parse the commits, but with git2 we can only use the actual refs/commit, which is much more expensive.

IMO, we are dealing with the linux repo with a edge case. With something like `tensorflow` the performance diff is not really noticeable, which is already a big repo with 100k commits and 2,6 M LOC. 

```bash
$ time onefetch
3.69s user 1.86s system 212% cpu 2.605 total

$ time ../../Projects/onefetch/target/release/onefetch  
3.84s user 1.34s system 200% cpu 2.586 total
```

But in the end it is your decision if that is enough or if performance is more important to you. We would have to wait for git2 bindings to logs.

o2sh:
what about this part:

> By the way, if you remove revwalk.set_sorting(git2::Sort::TIME)?; from get_git_history, its RT drops to:
> 
> get_git_history --> 15.868057149s

I think it can be removed without issue, hence getting us pretty close to git sys calls RT.

HallerPatrick:
> what about this part:
> 
> > By the way, if you remove revwalk.set_sorting(git2::Sort::TIME)?; from get_git_history, its RT drops to:
> > get_git_history --> 15.868057149s
> 
> I think it can be removed without issue: hence getting us pretty close to git sys calls RT.

It looks like the commit list is already sorted, at least are creation date and last commit the same with the removed line. So yeah I removed it

o2sh:
I took the liberty to make some minor changes e99c56f

HallerPatrick:
> I took the liberty to make some minor changes [e99c56f](https://github.com/o2sh/onefetch/commit/e99c56f0fe9b03c92e1f3234a322bba6599a4a31)

The cleaner the code the happier I am. 

Is anything now holding us back from merging?

o2sh:
We'll wait for @spenserblack to finish his review.

On my side, even if we loose in performance, I still believe that getting rid of the git sys calls - we still have one call in `get_git_version` - is the way to go as it makes for a more homogeneous code base and allows onefetch to be used even if the user doesn't have the Git binary installed on his machine.

HallerPatrick:
> We'll wait for @spenserblack to finish his review.
> 
> On my side, I'm very happy with this PR.
> 
> Indeed, even if we loose in performance, I still believe that getting rid of the git sys calls - we still have one call in `get_git_version` - is the way to go as it makes for a more homogeneous code base and allows onefetch to be used even if the user doesn't have the Git binary installed on his machine.

I am on your side. 

`get_git_versaion`/ `is_git_installed`are only there to get infos about the git binary, which cold never be replaced by `git2`.

src/onefetch/repo.rs
```diff
@@ -1,17 +1,122 @@
...
+                        if no_merges {
+                            let parents = commit.parents().len();
+                            if parents > 1 {
+                                return None;
+                            }
+                        }
+                        Some(commit)
+                    } else {
+                        None
+                    }
```
spenserblack, line 36: Since this is basically converting a `Result` to an `Option` that has been filtered, I think this can be simplified with `Result::ok` and `Option::filter`.
Something like
```rust
repo.find_commit(r).ok().filter(|commit| !(no_merges && commit.parents().len() > 1))
```
Once again, didn't test the above code, though :slightly_smiling_face:

spenserblack (approved):
>We'll wait for @spenserblack to finish his review.

Sorry, I didn't realize that you were still waiting for me.

This looks great, @HallerPatrick, thanks for all your help!
Once again, I left a suggested change that shouldn't change the logic at all, just the code style. Optional if you want to implement it.

Overall, LGTM :+1:

src/onefetch/repo.rs
HallerPatrick: Alright worked directly.

PR #347, opened 2020-12-16
Adding pyproject.toml and Pipfile for pip package manager

HallerPatrick:
I am not sure yet how to handle this case. For now `requirements.txt`, `Pipfile`and `pyproject.toml` are parsed and included for dependencies in output. I choose this because it is unlikely to include more then one of them in you python projects and I didnt want to change the logic of package retrieval. 

What are your thoughts? @o2sh @spenserblack

### DIFF
--- a/src/onefetch/deps/mod.rs
+++ b/src/onefetch/deps/mod.rs
@@ -34,7 +34,15 @@ impl DependencyDetector {
         );
         package_managers.insert(
             String::from("requirements.txt"),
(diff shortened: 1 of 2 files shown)


o2sh:
IMO the package_manager module (files in deps folder) needs to be refactored - in the same spirit of language.rs - to make it more easily extensible and easier to maintain.

I'll create an issue on that matter later today and assign it to me.

PR #350, opened 2020-12-17
 Refactoring of package manager module

o2sh:
With this PR, I introduced the `define_package_managers!` macro making it easier to add support for a new package manager.

However, it doesn't handle #347 where you have multiple parsers per package manager.


o2sh:
43206c4 seems to cover #347

HallerPatrick (approved):
Tested all changes on various repos and it yields the expected behaviour. Code wise there are no logical changes except the macro, which looks good to me.

HallerPatrick:
> [43206c4](https://github.com/o2sh/onefetch/commit/43206c42648db9f7879dde909a296446e3ca0268) seems to cover #347

I will close #347

spenserblack (approved):
LGTM :+1:

This looks much cleaner with the macro :smile:

PR #353, opened 2020-12-21
add support for GraphQL

geeseven:
Tweaks are welcomed.


o2sh:
Thanks a lot @geeseven,
with b69e84e, it tried to make it so that all the nodes look identical:

gg

geeseven:
either one works for me

PR #362, opened 2020-12-28
add qml support

mapau:
add qml support


mapau:
Hmm ok the qml.ascii is too wide .. I try to shrink it

o2sh:
Looks great @mapau 
I added some final touches in cd99dbd, 5bb12f9 and fd5bcac

qml

Thanks a lot for your PR :100:

PR #370, opened 2021-01-09
Add YAML output (-y) and ISO 8601 date format option (-z)

kernel-sanders:
This pull adds a -y flag which enables YAML output as well as a -z flag that outputs dates in ISO 8601 format. I'm using this to populate data on repos for a personal project so if these features don't interest you, no worries!

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -5,6 +5,7 @@ use {
         image_backends,
         info_field::{InfoField, InfoFieldOff},
         language::Language,
+        printer::SerializationFormat,
     },
     clap::{crate_description, crate_name, crate_version, App, AppSettings, Arg},
     image::DynamicImage,
@@ -31,8 +32,7 @@ pub struct Cli {
     pub excluded: Vec<String>,
     pub print_languages: bool,
     pub print_package_managers: bool,
-    pub print_in_json_format: bool,
-    pub print_in_yaml_format: bool,
+    pub output: Option<SerializationFormat>,
     pub true_color: bool,
     pub art_off: bool,
     pub text_colors: Vec<String>,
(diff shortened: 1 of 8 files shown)


src/onefetch/cli.rs
```diff
@@ -65,6 +67,18 @@ impl Cli {
             .long("json")
                 .help("Outputs Onefetch in JSON format.")
             )
+        .arg(
+            Arg::with_name("yaml")
+            .short("y")
+            .long("yaml")
+                .help("Outputs Onefetch in YAML format.")
```
spenserblack, line 74: Since this conflicts with the `--json` flag, I think that either
- [`conflicts_with("json")`](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.conflicts_with) should be added (and also a `conflicts_with("yaml")` for the json `Arg`), or
- The `--json` and `--yaml` flags should be replaced with a `--format` `Arg` with the [possible values](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.possible_values) `json` and `yaml`

Personally, I think the latter would be better to keep the CLI as simple as possible 🙂

spenserblack (changes requested):
Overall, LGTM! :heart: both YAML and ISO :smiley:

Left a suggestion on the CLI, let me know what you think 🙂

src/onefetch/cli.rs
o2sh: I'm working on the latter, I'll be pushing a commit soon.

o2sh:
Thanks for your PR @kernel-sanders.
With a871c07, I took the liberty - as suggested by @spenserblack - to merge the `--json` flag and `--yaml` into a single CLI flag called output.

Everything else LGTM 👍

o2sh (approved):

kernel-sanders (commented):
Looks great! Moving the output logic into printers.rs cleans up main a bit as well.

src/onefetch/cli.rs
```diff
@@ -243,7 +255,13 @@ impl Cli {
...
         let print_languages = matches.is_present("languages");
         let print_package_managers = matches.is_present("package-managers");
-        let print_in_json_format = matches.is_present("json");
+        let iso_time = matches.is_present("isotime");
+
+        let output = if let Some(output) = matches.value_of("output") {
+            Some(SerializationFormat::from_str(&output).unwrap())
+        } else {
+            None
+        };
```
spenserblack, line 264: Just another way to write the same thing 🙂 
```suggestion
        let output = matches.value_of("output").map(SerializationFormat::from_str).transpose().unwrap();
```

spenserblack (approved):

PR #373, opened 2021-01-12
Add CLI option to switch true colors on/off

spenserblack:
Allows user to control if onefetch should attempt to write true colors to the terminal.

***

I found myself wanting to turn off true colors when executing `onefetch` so that the output would match my preferred color scheme, and figured that other users might want to do the same :slightly_smiling_face:

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -175,6 +175,20 @@ impl Cli {
                 .help("Colors (X X X...) to print the ascii art."),
         )
         .arg(
+            Arg::with_name("true-colors")
+                .value_name("SETTING")
+                .long("true-color")
+                .possible_values(&["auto", "on", "off"])
+                .default_value("auto")
+                .help("control the display of true colors (auto, on, off)")
+                .long_help(
+                    "- auto: detect if true colors are supported
+- on: force true colors to be displayed
+- off: force only 8 standard colors to be displayed
+"
+                )
+        )
+        .arg(
             Arg::with_name("text-colors")
                 .short("t")
                 .long("text-colors")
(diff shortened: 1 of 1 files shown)


src/onefetch/cli.rs
```diff
@@ -174,6 +174,20 @@ impl Cli {
...
+                .long("true-color")
+                .possible_values(&["auto", "on", "off"])
+                .default_value("auto")
+                .help("control the display of true colors (auto, on, off)")
+                .long_help(
+                    "- auto: detect if true colors are supported
+- on: force true colors to be displayed
+- off: force only 8 standard colors to be displayed
+"
+                )
```
o2sh, line 189:
```suggestion
            Arg::with_name("true-color")
                .long("true-color")
                .value_name("WHEN")
                .takes_value(true)
                .possible_values(&["auto", "never", "always"])
                .default_value("auto")
                .hide_default_value(true)
                .help("Specify when to use true color (*auto*, never, always).")
                .long_help(
                    "Specify when to use true color (*auto*, never, always). \n\
                    If set to auto: true color will be enabled if supported by the terminal."
                )
```

src/onefetch/cli.rs
```diff
@@ -249,7 +263,12 @@ impl Cli {
             )
 .get_matches();
 
-        let true_color = cli_utils::is_truecolor_terminal();
+        let true_color = match matches.value_of("true-colors") {
+            Some("on") => true,
+            Some("off") => false,
+            Some("auto") => cli_utils::is_truecolor_terminal(),
+            _ => unreachable!(),
+        };
```
o2sh, line 271:
```suggestion
        let true_color = match matches.value_of("true-colors") {
            Some("always") => true,
            Some("never") => false,
            Some("auto") => cli_utils::is_truecolor_terminal(),
            _ => unreachable!(),
        };
```

o2sh (approved):
You're absolutely right, for those who like to play with their terminal's color scheme, this flag can come to be very handy.

My only suggestion would be to use `--show-logo` as a template for its declaration as the two flags seem to operate similarly.

spenserblack:
>My only suggestion would be to use `--show-logo` as a template

Nice catch! The args should definitely have consistent usage. Suggestions committed.

PR #380, opened 2021-01-22
Cache dependencies

spenserblack:
Hopefully, this should speed up the CI (after first cache is saved).

***

If you think this is a good idea, please squash merge, the commit history in this PR is a bit messy :cry:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+
+    - name: Restore cargo cache
     - uses: actions/cache@v2
       with:
         path: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
```
o2sh, line 24:
```suggestion

    - name: Restore cargo cache
    - uses: actions/cache@v2
      with:
        path: |
          ~/.cargo/registry
          ~/.cargo/git
          target
        key: ${{ runner.os }}-cargo-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
```

o2sh (approved):
Considering all the dependencies required to compile Onefetch, this is indeed a very good idea :+1:, thanks @spenserblack 

My only suggestion would be to add a job name in order to make it stand out in CI.

Feel free to squash it when you want.

PR #404, opened 2021-03-12
Run cargo update

spenserblack:
Fixes severe warnings for `generic-array` and `rand_core`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1157,9 +1157,9 @@ dependencies = [
 
 [[package]]
 name = "rand_core"
-version = "0.6.1"
+version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


o2sh:
Thanks for fixing the build @spenserblack, I see that you've updated more than just `generic-array` and `rand_core` in the Cargo.lock. Is there a reason for that ? Especially `image` and `libc` which would de-sync with the version specified in the Cargo.toml. We should maybe merge #401 and #399 before regenerating the Cargo.lock ?

spenserblack:
@o2sh Good catch! That was the result of running `cargo update` without any additional arguments, which updates *all* dependencies by default. Now that you've brought it up, I think it's probably best to only update those few dependencies. I'll push up a new commit.

spenserblack:
The remaining audit warnings seem to be upstream issues with `tokei`'s and `askalono`'s unmaintained dependencies, and it doesn't look like there is a new release yet of either that would fix those audit warnings.

PR #412, opened 2021-04-10
Add Scheme language

wigust:
---
name: Scheme
about: Request for a new language to be supported
title: "Support for Scheme"
labels: 'Type: Enhancement, good first issue'
assignees: ''

---

# Scheme
https://en.wikipedia.org/wiki/Scheme_(programming_language)

## Logo

https://en.wikipedia.org/wiki/Scheme_(programming_language)#/media/File:Lambda_lc.svg

## Upstream support

- [x] [tokei] supports this language

[tokei]: https://github.com/XAMPPRocky/tokei


o2sh:
Thanks a lot for your PR @wigust 
a360083, IMO `Color::Black` should be avoided as it doesn't show on most terminals - with black background. I switched it to DarkGrey - for terminals that support TrueColors:

scheme

PR #420, opened 2021-05-04
feat: cloning remote repository

### LINKED ISSUE #419: Support for fetching remote repositories
Hello, what do you think about that? Command for that could look like that
```bash
onefetch --github o2sh/onefetch
```
Or instead --github, --remote
```bash
onefetch --remote o2sh/onefetch
```

gbaranski:
closes #419 

- [x] Cloning remote repository from HTTP URL
- [x] Caching repositories
- [x] Progress bars(git clone may take some time)
- [x] ~~Improve performance by fetching with depth = 1~~ I can't find option in git2 library


gbaranski:
https://user-images.githubusercontent.com/34353640/117003027-9ce97a80-ace4-11eb-9908-637612c8d170.mp4

That's how it looks like with progress bars

src/onefetch/cli.rs
```diff
@@ -253,6 +254,11 @@ impl Cli {
             .takes_value(true)
             .help("Ignore all files & directories matching EXCLUDE."),
         )
+        .arg(
+            Arg::with_name("remote")
+            .long("remote")
+            .help("Clone repository from remote"),
+        )
```
spenserblack, line 261: I think it would be useful if this arg either
- Had another arg that [required](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.requires) this arg
- Took a second value

So that the user could specify whether the cloned repo should be deleted or kept.

e.g. `onefetch --remote <my URL> --keep` to keep the repo after cloning it.

spenserblack:
:thinking: I'm a bit worried about clone depth not being set. Perhaps at least a warning could be documented somewhere, like in the arg's help?

src/onefetch/cli.rs
gbaranski: > I think it would be useful if this arg either
> 
>     * Had another arg that [required](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.requires) this arg
> 
>     * Took a second value
> 
> 
> So that the user could specify whether the cloned repo should be deleted or kept.
> 
> e.g. `onefetch --remote <my URL> --keep` to keep the repo after cloning it.

Where would you like to keep this repository?

spenserblack: You could probably just clone it to the current directory if it's being kept.

gbaranski: > You could probably just clone it to the current directory if it's being kept.

I think it does not make much sense, because you could just do `git clone` and then run onefetch there.

Purpose of this feature is to quickly look up information about repository while not caring at all about the repository itself, but just about statistics.

spenserblack: >Purpose of this feature is to quickly look up information about repository while not caring at all about the repository itself, but just about statistics.

Yes, that makes sense, but, since it is not doing a shallow clone, you're getting the whole repository anyway. For example, if this feature was used on [the Linux repo](https://github.com/torvalds/linux), it would take a long time. Worse if the user uses this feature and *then* decides to clone it to do some work on it. It is hardly "quickly looking up information."

For example, somebody could use this feature, and then decide it's a repository that they would want to work on based on the stats. If this feature doesn't support keeping the repository, they would have to clone it a second time.

Although I agree that it is unnecessary to add a `--keep` option *if* the repo is cloned with a shallow depth, since onefetch's clone would take less time, and a shallow clone isn't that useful for actual work. :slightly_smiling_face:

spenserblack:
:frowning_face: I'm having trouble using this feature right now. `[onefetch error]: there is no TLS stream available; class=Ssl (16)`
Seems like an issue with git2 failing to link with OpenSSL when I compile.

While I look into this, could someone run
- `time cargo run --release -- --remote https://github.com/torvalds/linux.git`
- `time git clone --depth=1 https://github.com/torvalds/linux.git` and `time cargo run -- ./linux/`

I'm curious what the time difference is between onefetch cloning a repository and the user manually performing a shallow clone.

o2sh:
Well done @gbaranski :+1: , I took the liberty with bd3abbf to remove the `default-feature = false` from the git2 dependency in order to fix the issue reported by @spenserblack

> [onefetch error]: there is no TLS stream available; class=Ssl (16)

As @spenserblack mentionned, the `--remote` CLI arg should have a required value and not rely on the `INPUT` which is optional.
What about some regex pattern matching to make sure the user has provided a valid URL? I'm not sure how good is the error handling done by the `RepoBuilder` APIs :thinking: .

Can we handle SSH? Or do we stick to HTTPS, in which case the the "ssh" and "ssh_key_from_memory" features in [git2-rs](https://github.com/rust-lang/git2-rs/blob/90a45bb106a723259e2a0f0c3ff5fb4f70723fda/Cargo.toml#L38) should be disabled.

gbaranski:
> What about some regex pattern matching to make sure the user has provided a valid URL? 

Yeah I also thought about that, we could check if it starts by `http(s)://`, but what about SSH? check if starts with `git@`?

Or maybe just use `INPUT` and check if folder exists, if it does not, then try to clone it from remote git repo. 

> Can we handle SSH? Or do we stick to HTTPS, in which case the the "ssh" and "ssh_key_from_memory" features in [git2-rs](https://github.com/rust-lang/git2-rs/blob/90a45bb106a723259e2a0f0c3ff5fb4f70723fda/Cargo.toml#L38) should be disabled.

I think we could add `--ssh-key <path>` argument, so people could also alias `onefetch` to `onefetch --ssh-key ~/.ssh/id_ed25519`

o2sh:
For the URL validation, you can make use of the [`url`](https://docs.rs/url/2.2.1/url/) crate and do smth in the line of: `Url::parse(&url)is_ok()`.

> Or maybe just use INPUT and check if folder exists, if it does not, then try to clone it from remote git repo.

I would prefer to keep the default behaviour of the tool as is, which is to run locally and be 100% offline.

> I think we could add --ssh-key <path> argument, so people could also alias onefetch to onefetch --ssh-key ~/.ssh/id_ed25519

Or we can simply try with `$HOME/.ssh/id_rsa` like in [this example](https://docs.rs/git2/0.13.17/git2/build/struct.RepoBuilder.html). But yeah, maybe handling SSH isn't so important for a first iteration...

gbaranski:
> For the URL validation, you can make use of the [`url`](https://docs.rs/url/2.2.1/url/) crate and do smth in the line of: `Url::parse(&url)is_ok()`.
> 
> > Or maybe just use INPUT and check if folder exists, if it does not, then try to clone it from remote git repo.
> 
> I would prefer to keep the default behaviour of the tool as is, which is to run locally and be 100% offline.
> 
> > I think we could add --ssh-key  argument, so people could also alias onefetch to onefetch --ssh-key ~/.ssh/id_ed25519
> 
> Or we can simply try with `$HOME/.ssh/id_rsa` like in [this example](https://docs.rs/git2/0.13.17/git2/build/struct.RepoBuilder.html). But yeah, maybe handling SSH isn't so important for a first iteration...

Yeah but some people might not want to use their SSH Key for that. And what if someone has multiple keys and wants to use specific one?

spenserblack:
>what if someone has multiple keys and wants to use specific one?

I've mostly used default behavior when using SSH to clone, so I may be mistaken, but wouldn't `~/.ssh/config` be used to control which key is used for the github.com (or wherever they're cloning from) domain? We could probably just expect the user to set up the SSH configuration so that they use the key they want to use.

Somewhat related:
- [`GIT_SSH`](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_miscellaneous)
- [`GIT_SSH` AND `GIT_SSH_COMMAND`](https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHcode)
- `core.sshCommand`

Git users can use these to override the SSH key used (e.g. `git config core.sshCommand "ssh -i my-key <other args>"`),
but it's more typical (at least according to Git's documentation) to use `~/.ssh/config`.

gbaranski:
> > what if someone has multiple keys and wants to use specific one?
> 
> I've mostly used default behavior when using SSH to clone, and I know little about SSH beyond that, so I may be mistaken. But wouldn't `~/.ssh/config` be used to control which key is used for the github.com (or wherever they're cloning from) domain? Would it be sufficient to expect the user to set up the SSH configuration so that they use the key they want to use?
> 
> Somewhat related:
> 
>     * [`GIT_SSH`](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_miscellaneous)
> 
>     * [`GIT_SSH` AND `GIT_SSH_COMMAND`](https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHcode)
> 
>     * `core.sshCommand`
> 
> 
> Git users can use these to override the SSH key used (e.g. `git config core.sshCommand "ssh -i my-key <other args>"`),
> but it's more typical (at least according to Git's documentation) to use `~/.ssh/config`.

I think most people don't even have ~/.ssh/config file, because that's not necessary

spenserblack:
>I think most people don't even have ~/.ssh/config file

In that case, an additional CLI arg for SSH key path would probably be best, as you suggested earlier. I think the arg should require the `--remote` arg, and use `<home dir>/.ssh/id_rsa` as a default. There are a few crates to help get the home directory of each OS: [dirs](https://docs.rs/dirs/3.0.2/dirs/), [dirs-next](https://docs.rs/dirs-next/2.0.0/dirs_next/), [directories](https://docs.rs/directories/3.0.2/directories/), etc.

This is assuming that SSH support should be part of this PR and not a separate PR :slightly_smiling_face:

spenserblack:
Still not sure about the necessity of this feature, though. It is very easy to accomplish this with a small script. See [this bash script](https://gist.github.com/spenserblack/4214c81d146a9fd93b71f6c28dc6075f), which is run with `./onefetch-remote <repo URL>`.

I appreciate the work you've done on this PR, but are we sure that this feature should be added?

o2sh:
So far I've been kinda neutral about this feature overall - as it doesn't really "harm" the project to have it -, but the bash script does seem to be the best option for this :confused:.

spenserblack:
:thinking: This PR has been in stasis for a while. Perhaps we should ask another collaborator or frequent contributor to offer their opinion?

o2sh:
@HallerPatrick @yoichi @ebroto @erikgaal @CephalonRho

Luke-zhang-04:
so... this is kinda like https://github.com/spenserblack/repofetch?

o2sh:
Not really, [repofetch](https://github.com/spenserblack/repofetch) relies on the GH APIs to fetch information; it doesn't clone the repository locally. See #419 for more context.

Luke-zhang-04:
Hm, ok interesting. I'm just gonna say, I personally don't see myself using this feature, but it's cool nonetheless.

PR #430, opened 2021-05-27
Make audit scheduled and manually runnable

spenserblack:
Moves the security audit into a separate workflow so that it can be run
on a schedule, and also dispatched manually.

***

Since a new security issue could be found and reported for an existing dependency even if that dependency doesn't change, I thought it could be helpful to regularly check the audit.
I scheduled it to be run once daily at 23:59 GMT, but it can be set for a different time/frequency :slightly_smiling_face:

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -11,7 +11,7 @@ on:
       - '**/Cargo.toml'
       - '**/Cargo.lock'
   schedule:
-    - cron: '59 23 * * *'
+    - cron: '0 0 * * 0'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,19 @@
+name: Security Audit
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '59 23 * * *'
+  workflow_dispatch:
+
+jobs:
+  sec:
+    name: Security audit
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/audit.yml
```diff
@@ -0,0 +1,19 @@
+name: Security Audit
+on:
+  push:
+    branches: [ master ]
```
o2sh, line 4:
```suggestion
    branches: [ master ]
    paths: 
      - '**/Cargo.toml'
      - '**/Cargo.lock'
```

No need to trigger the workflow if the commit doesn't affect the dependencies.

.github/workflows/audit.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+    branches: [ master ]
```
o2sh, line 6:
```suggestion
    branches: [ master ]
    paths: 
      - '**/Cargo.toml'
      - '**/Cargo.lock'
```

No need to trigger the workflow if the commit doesn't affect the dependencies.

.github/workflows/audit.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+    - cron: '59 23 * * *'
```
o2sh, line 8:
```suggestion
    - cron: '0 0 * * 0'
```

What about once a week ?

o2sh (commented):
It's indeed safer to have the audit checks be triggered periodically.

I left a few notes.

.github/workflows/audit.yml
spenserblack: Sounds good to me :+1:

spenserblack:
Thanks for the review! Applied the suggested changes.

PR #435, opened 2021-05-31
Added Ada support

Labdabeta:
Made the ASCII art for the logo using JavE and the logo from the FSF GNAT compiler as reference. Just adds basic Ada support. Tokei already has Ada support, so it was just a matter of adding it to the recognized languages. I think the logo is just 21x40, but it starts with an empty column for aesthetics that can easily be removed if I miscounted and its actually 21x41.


o2sh:
The ASCII art looks great! with 20a912e, I took the liberty to add `TrueColors` and to replace the `Color::Black` with `Color::White` as the primary color for the logo to be more visible on most terminals - with black backgrounds.
Screenshot 2021-05-31 at 18 23 06

PR #436, opened 2021-05-31
feat(languages): add Svelte support

Luke-zhang-04:
image


o2sh:
Perfect @Luke-zhang-04, just a small correction with a481c40. We use custom serialization name only if we want something that is different from the language name in lowercase.

PR #438, opened 2021-06-01
Add HCL Support

PsypherPunk:
Add support for HCL/Terraform:

- ASCII art generated from `image_to_ascii.py` and the Terraform
  [logo](https://www.terraform.io/assets/images/og-image-8b3e4f7d.png).
  - `TrueColor` values derived from the RGB values therein.


PsypherPunk:
As per the contribution guidelines, specifically that the ASCII logo's _"dimensions must fall below 25*40"_, the logo is 25x40 but the addition of `{0}` and `{1}` push the width to `46`—is that a concern?

spenserblack:
It shouldn't be a concern. It's the rendered width/height that we're trying to enforce. As long as the *rendered* characters stay within the dimensions, it should be OK :slightly_smiling_face:

o2sh:
Thanks a lot for your contribution @PsypherPunk, in d591e1b I cleaned the ASCII art to make it more symmetrical.
Screenshot 2021-06-02 at 01 39 54

PR #443, opened 2021-06-08
Fixed Fortran ASCII to be more symmetrical

jake-87:
I use onefetch a lot, and I work with a lot of FORTRAN code, and the off center rocket just kills me every single time. It was the 3rd result when looking up ASCII rocket, no changes made. This fixes up the rocket and makes it much more pleasing to look at.


o2sh:
I agree, symmetrical designs are indeed more pleasing to the eye. Thanks for your contribution @jake-87

o2sh:
I realised after merging that the rocket's burst wasn't properly centered (same with the `.f90` heading), I fixed it in my last commit. I hope you don't mind.

Screenshot 2021-06-08 at 22 53 15

jake-87:
All good, symmetry is worth it.

PR #448, opened 2021-06-24
ignore bots in commits

### LINKED ISSUE #446: Ignore [bot] commits 🤖🤖
With the democratization of dependency management bots and other non-human commit-generating tools, commit logs can get very noisy.

It would be nice to have an option for filtering these types of commits out via a CLI flag `--no-bots`. A simple approach would be to skip all commits whose author's name includes  the string `[bot]`.

o2sh:
closes #446

### DIFF
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -27,6 +27,7 @@ pub struct Cli {
     pub image_backend: Option<Box<dyn image_backends::ImageBackend>>,
     pub image_color_resolution: usize,
     pub no_merges: bool,
+    pub no_bots: bool,
     pub no_color_palette: bool,
     pub number_of_authors: usize,
     pub excluded: Vec<String>,
@@ -209,6 +210,11 @@ impl Cli {
             .help("Ignores merge commits."),
         )
         .arg(
+            Arg::with_name("no-bots")
+            .long("no-bots")
+            .help("Ignore bots."),
+        )
+        .arg(
             Arg::with_name("isotime")
             .short("z")
             .long("isotime")
@@ -263,6 +269,7 @@ impl Cli {
         };
         let no_bold = matches.is_present("no-bold");
         let no_merges = matches.is_present("no-merges");
+        let no_bots = matches.is_present("no-bots");
         let no_color_palette = matches.is_present("no-palette");
         let print_languages = matches.is_present("languages");
         let print_package_managers = matches.is_present("package-managers");
@@ -360,6 +367,7 @@ impl Cli {
             image_backend,
             image_color_resolution,
             no_merges,
+            no_bots,
             no_color_palette,
             number_of_authors,
             excluded,
--- a/src/onefetch/cli.rs
+++ b/src/onefetch/cli.rs
@@ -9,6 +9,7 @@ use {
     },
     clap::{crate_description, crate_name, crate_version, App, AppSettings, Arg},
     image::DynamicImage,
+    regex::Regex,
     std::{convert::From, env, str::FromStr},
     strum::IntoEnumIterator,
     term_size,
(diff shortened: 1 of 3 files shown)


src/onefetch/cli.rs
```diff
@@ -349,6 +358,16 @@ impl Cli {
...
 
+        let bot_exclude_pattern = if matches.is_present("no-bots") {
+            if let Some(pattern) = matches.value_of("no-bots") {
+                Some(String::from(pattern))
+            } else {
+                Some(String::from(r".*\[bot\].*"))
+            }
+        } else {
+            None
+        };
```
spenserblack, line 369:
```suggestion
        let bot_exclude_pattern = matches.is_present("no-bots").then(
            || matches.value_of("no-bots").map_or(r"\[bot\]".to_string(), String::from)
        )
```
I think this would work. Been a while since I've used Rust, so the borrow checker might not like the closer, though :shrug:

As a side note, for regex, I think `.*\[bot\].*` and `\[bot\]` would match the same things, since `.` matches everything but a newline AFAIK. Just that `.*\[bot\].*` would match the whole string, and `\[bot\]` would only match `[bot]` within the string.

This is totally *not* a necessary change, just comes down to preferred coding style :smiley:

src/onefetch/cli.rs
```diff
@@ -208,6 +209,14 @@ impl Cli {
             .long("no-merges")
             .help("Ignores merge commits."),
         )
+        .arg(
+            Arg::with_name("no-bots")
+            .long("no-bots")
+            .min_values(0)
+            .max_values(1)
+            .value_name("REGEX")
+            .help("Exclude [bot] commits. Use <REGEX> to override the default pattern."),
```
spenserblack, line 218: I just realized that maybe a validator should be included to prevent invalid regex, which could cause a panic [here](https://github.com/o2sh/onefetch/pull/448/files#diff-f6a4b0b4b4baa74961fd54b0f78627290079064c0c0c90281d0a9a789f85c362R288)

spenserblack (commented):
I had a slight concern about a possible panic, but overall LGTM! :tada:

src/onefetch/repo.rs
```diff
@@ -269,3 +282,9 @@ pub fn is_valid(repo_path: &str) -> Result<bool> {
     let repo = Repository::open_ext(repo_path, RepositoryOpenFlags::empty(), Vec::<&Path>::new());
     Ok(repo.is_ok() && !repo?.is_bare())
 }
+
+pub fn is_bot(author: Signature, bot_exclude_pattern: &Option<String>) -> bool {
+    let author_name = String::from_utf8_lossy(author.name_bytes()).into_owned();
+    let re = Regex::new(bot_exclude_pattern.as_ref().unwrap()).unwrap();
```
spenserblack, line 288: Oh, also you might want to make use of `lazy_static!`, since I believe that this will compile regex on each line that `is_bot` is called on, otherwise.

src/onefetch/cli.rs
o2sh: You're absolutely right, I thought the Regex had to match the entire string to return true :man_facepalming:

src/onefetch/repo.rs
o2sh: Great remark! I chose to use a `Option<Regex>` instead of a `Option<String>` for the bot_exclude_pattern so that I can provide an pre-compiled Regex pattern to the program.

src/onefetch/cli.rs
o2sh:
```
.validator(|p| {
    if Regex::from_str(&p).is_err() {
        return Err(String::from("must be a valid regex pattern"));
    } else {
        Ok(())
    }
}),
```

spenserblack (approved):
:tada:

PR #452, opened 2021-06-30
Add CLI option for displaying author emails

### LINKED ISSUE #451: Add CLI flag to display "username <email>" in the Author section ✉️
`git shortlog` provides an option `--email` to display the author's email address alongside its username:

```
234 John Smith <john@smith.org>
126 Olivier Dupont <olivier@me.fr>
```
It could be a good idea to implement the same option in Onefetch.

remopas:
feature request, closes #451

### DIFF
--- a/src/onefetch/info.rs
+++ b/src/onefetch/info.rs
@@ -21,7 +21,7 @@ pub struct Info {
     creation_date: String,
     languages: Vec<(Language, f64)>,
     dependencies: String,
-    authors: Vec<(String, usize, usize)>,
+    authors: Vec<(String, Option<String>, usize, usize)>,
     last_change: String,
     repo_url: String,
     number_of_commits: String,
@@ -221,7 +221,7 @@ impl Info {
         let number_of_branches = internal_repo.get_number_of_branches()?;
         let creation_date = internal_repo.get_creation_date(config.iso_time)?;
         let number_of_commits = internal_repo.get_number_of_commits();
-        let authors = internal_repo.get_authors(config.number_of_authors);
+        let authors = internal_repo.get_authors(config.number_of_authors, config.email);
         let last_change = internal_repo.get_date_of_last_commit(config.iso_time);
         let (repo_size, file_count) = internal_repo.get_repo_size();
         let workdir = internal_repo.get_work_dir()?;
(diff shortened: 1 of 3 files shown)


src/onefetch/info.rs
```diff
@@ -312,15 +312,21 @@ impl Info {
...
             self.authors.iter().enumerate()
         {
-            if i == 0 {
+			let author = if let Some(author_email) = author_email_opt {
+				format!("{} <{}>", author_name, author_email)
+			}else {
+				author_name.to_owned()
+			};
+			
+			if i == 0 {
```
spenserblack, line 324:
```suggestion
            let author = if let Some(author_email) = author_email_opt {
                format!("{} <{}>", author_name, author_email)
            } else {
                author_name.to_owned()
            };

            if i == 0 {
```
I think this would fail the format check, otherwise :slightly_smiling_face: 

This project uses 4 spaces for indentation (please check the `.editorconfig` file).
You may want to install an EditorConfig plugin for your preferred text editor so that
your style conforms with this project's style, or any other project that has an `.editorconfig`
file.

src/onefetch/repo.rs
```diff
@@ -83,11 +83,12 @@ impl<'a> Repo<'a> {
 
         authors.truncate(n);
 
-        let authors: Vec<(String, usize, usize)> = authors
+        let authors: Vec<(String, Option<String>, usize, usize)> = authors
             .into_iter()
             .map(|(author_email, author_nbr_of_commits)| {
                 (
                     author_name_by_email.get(&author_email).unwrap().trim_matches('\'').to_string(),
+					if include_email { Some(author_email) } else { None },
```
spenserblack, line 91:
```suggestion
                    include_email.then(|| author_email),
```
See if this works. [`then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then) is a bit more readable IMO
(although I would prefer the `then_some` which is currently nightly-only).
Please note that this is *not* a necessary change, because the if-else expression is also fine IMO.

However, this suggestion also changes tab indentation to spaces, which is a bit necessary to keep the code style
consistent.

spenserblack (changes requested):
Thanks for your PR!

I left some suggestions about code style, which I believe would be necessary to pass the format check in the CI.

spenserblack:
Also, looks like your commits aren't getting mapped to your GitHub account. Probably an email mismatch. I wouldn't want for you to not get credit for your work :slightly_smiling_face:

remopas:
> Thanks for your PR!
> 
> I left some suggestions about code style, which I believe would be necessary to pass the format check in the CI.

Thanks a bunch for the suggestions! Will be more careful regarding formatting and tab-size in the future.

> Also, looks like your commits aren't getting mapped to your GitHub account. Probably an email mismatch. Do you need to amend the commit author? Not a big deal, but I wouldn't want for you to not get credit for your work 🙂

Oops! I fixed my git settings, but no need to amend them for this PR.

src/onefetch/repo.rs
```diff
@@ -63,7 +63,7 @@ impl<'a> Repo<'a> {
         number_of_commits.to_string()
     }
 
-    pub fn get_authors(&self, n: usize) -> Vec<(String, usize, usize)> {
+    pub fn get_authors(&self, n: usize, include_email: bool) -> Vec<(String, Option<String>, usize, usize)> {
```
spenserblack, line 66: Looks like the format check still failed. I think it's this line being too long. Try listing the function arguments in multiple lines.

spenserblack (changes requested):
Thanks for the changes! Looks like there's one more change to pass the formatting check.

spenserblack (approved):
Thanks! LGTM

Issue #447, opened 2021-06-18
Merge duplicate contributors 👤👤👤👤


o2sh:
Since @yoichi's PR #243: we identify each contributor by their email address instead of their username which is definitely the right solution. However in some cases, a user may have changed his GH email along the way and therefore appear twice - or more - in the `Authors` section: 

<img width="738" alt="Screenshot 2021-06-18 at 21 48 25" src="https://user-images.githubusercontent.com/13710835/122609847-f0d6d580-d07e-11eb-9716-3207d538a7f9.png">

It may be a good idea to have a second pass in [`get_authors`](https://github.com/o2sh/onefetch/blob/aa80dc96cd9a94940ccbba5f1160ca3adc69523b/src/onefetch/repo.rs#L53) and merge the lines that share the same `username.to_lowercase()`

HallerPatrick:
Sounds like a good idea to me.I would be interested in doing this.  @o2sh is it still open?

o2sh:
It's all yours :thumbsup:

HallerPatrick:
@o2sh While currently working on this, I am asking myself, why it is the better way to identify by email in the first place and not the username, when only the email can change and not the username? Or am I missing something?

o2sh:
You can actually also change your GitHub username, even though it's less common. 

The username/email that appears in the `git log` (and parsed by onefetch) is taken from the `.gitconfig`, so if your SSH key is properly configured: you can actually put whatever you want in their and still be able to push to your repository, - or in HTTPS, if you login with username/password. On GitHub, each commit is linked to an account based solely on its email adress(es). 

I agree that merging based on the email seems arbitrary but so it is if we used the username, both give unsatisfactory results. Hence the second pass.

HallerPatrick:
Alrighty, that seems reasonable 👍

spenserblack:
To add on to this, most tools that I'm familiar with use the email instead of the name to identify the committer. GitHub does this, for example. Check out [this test commit](https://github.com/spenserblack/test-repo/commit/5324628a2f7fe93f1bbefefbac6bcde931b6026c), where I used the same email but a different name. GitHub still identifies me as the committer.
So I think that identifying by email is more consistent with other tools that people use.

HallerPatrick:
Made a pull request. It is still not perfect, because there are also commits with a different username _and_ email. We would have to make guesses, if they are the same person, where I would refrain from...

[...]

PR #453, opened 2021-06-30
Use mailmap bindings

o2sh:
Following the discussion #447, I added the `.mailmap`author matching which came out to be quite straightforward thanks to the git2-rs bindings - 9ba9533.

I also changed the merging logic from "email" to "username + email" to comply with `git shortlog`.


HallerPatrick (commented):
This feels way cleaner! If there is no `.mailmap`, then this will behave like before right?

o2sh:
> This feels way cleaner! If there is no `.mailmap`, then this will behave like before right?

`git_mailmap_from_repository` seems to [purposely never errors](https://github.com/libgit2/libgit2/blob/e16747660f0d21928d2abaa49ac0657c6664f157/src/mailmap.c#L378). However, I've just noticed that `commit.author_with_mailmap(&mailmap)` fails if a commit's signature doesn't have an email.

I fixed it with e031a88.

spenserblack (approved):
LGTM!

HallerPatrick (approved):

PR #454, opened 2021-07-01
add support for powershell

dxking:
This pull request adds support for PowerShell. 

Logo was derived and modified from the existing `zsh.ascii` file. I'm not totally in love with the ASCII art but can't quite pin down what could improve it.


spenserblack:
:thinking: Perhaps you could use white text for the `>_` to make it a bit more distinct from the zsh logo and closer to the PowerShell logo. What do you think?

dxking:
@spenserblack I made some changes to `powershell.ascii` to add white for the `>_` and switched up the characters being used (nothing really seemed to go well with `#` that was being used for the background). I think it looks a bit better now.
image

o2sh:
To be even closer to the original design, we may want to "lean" toward smth like that:



```
{0}         ########################
{0}        ####{1}###{0}#################
{0}       ######{1}###{0}###############
{0}      ########{1}###{0}#############
{0}     ##########{1}###{0}###########
{0}    #########{1}###{0}############
{0}   ########{1}###{0}#############
{0}  #######{1}###{0}##############
{0} ######{1}###{0}#####{1}######{0}####
{0}########################
```

What do you think @dxking ?

o2sh:
To be even closer to the original design, we may want to "lean" toward smth like that:



```
{0}         ########################
{0}        ####{1}####{0}################
{0}       ######{1}####{0}##############
{0}      ########{1}####{0}############
{0}     ##########{1}####{0}##########
{0}    #########{1}####{0}###########
{0}   ########{1}####{0}############
{0}  #######{1}####{0}#############
{0} ######{1}####{0}#####{1}######{0}###
{0}########################
```
What do you think @dxking ?

dxking:
@o2sh Yeah, that looks great! I can get `powershell.ascii` updated with that design, thanks!

dxking:
Updated to use your suggested ASCII art.

PR #465, opened 2021-07-10
Count hidden files and directories

### LINKED ISSUE #464: Discover hidden files and directories
Would it be possible to make onefetch look through hidden directories inside a repository?

I have a directory in one of my repositories that is hidden (starts with a dot `.system`) and onefetch doesn't find any code in the repository, however, if I unhide that directory `mv .system system` it finds all the code inside. 

Which is why I'm assuming the issue is with hidden items.

0x20F:
Closes #464

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -37,6 +37,7 @@ pub struct Config {
     pub text_colors: Vec<String>,
     pub iso_time: bool,
     pub show_email: bool,
+    pub include_hidden: bool,
 }
 
 impl Config {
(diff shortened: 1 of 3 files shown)


o2sh:
Thanks for your PR @0x20F 
IMO, it would be preferable to follow tokei's example and have a CLI flag `--hidden` that defaults to false.

0x20F:
I agree, it's a lot nicer to allow users to choose.
I'm not sure if I did it the best way it could be done, feedback is welcome 😅

o2sh:
LGTM :tada: 

Thanks a lot for your contribution @0x20F

PR #494, opened 2021-09-30
Automate "languages supported" badge

### LINKED ISSUE #493: Automate documentation of language count
Currently, the READMEs report "over 50 languages" are supported. While this is true, it's a bit on the low side to be honest. Here is the current language count:
```console
$ onefetch --version
onefetch 2.10.2
$ onefetch --languages | wc -l
76
```
It might be possible to create an Action to take the output of `cargo run -- --languages | wc -l` and parse the result into the READMEs, to keep the count up-to-date.
Although, if the language count is kept up-to-date in the READMEs, all of them will have to be updated to no longer have "more than", "以上", etc.

Alternatively, instead of updating the text of the README, perhaps a badge with automated updates can be added to show the most up-to-date supported languages count.

spenserblack:
This creates a badge for the number of languages supported.
The badge is re-created either when a release is published or the workflow is manually dispatched.

### To Do

- [x] ~~Skip CI when commit message contains `[Skip CI]`~~ (looks like it already does this :laughing: )
- [x] Figure out what to do if the commit is empty (no changes to languages supported count)
  - Allow empty commits, fail silently?

***

Resolves #493


o2sh (approved):
LGTM :+1: I commited some minor changes - mostly renaming -, hopefully I didn't break anything... 

Don't forget to trigger the workflow before merging to generate a new badge.

spenserblack:
Oh, right, I forgot to link to the wiki 😆 
Workflow triggered

Issue #490, opened 2021-09-18
Improve language support of `onefetch`


o2sh:
Listed here are  _"all"_ the languages that could be supported by onefetch - the checked lines are the ones already supported. Beside giving us a better idea of our coverage, this list can serve as a basis for anyone looking to find a language for which they want to add support for.

- [x] ABNF
- [x] ABAP
- [ ] ActionScript
- [x] Ada
- [x] Agda
- [ ] Alex
- [ ] Alloy
- [x] Arduino
- [x] Assembly
- [x] AutoHotKey
- [ ] Autoconf
- [ ] Automake
- [x] BASH
- [ ] Batch
- [ ] Bean
- [ ] BrightScript
- [x] C
- [x] CMake
- [x] C#
- [ ] C Shell
- [ ] Cabal
- [ ] Cassius
- [x] Ceylon
- [x] Clojure
- [ ] COBOL
- [ ] CodeQL
- [x] CoffeeScript
- [ ] Cogent
- [x] ColdFusion
- [x] Coq
- [x] C++
- [x] Crystal
- [x] CSS
- [x] D
- [ ] DAML
- [x] Dart
- [ ] Device Tree
- [ ] Dhall
- [x] Dockerfile
- [ ] Dream Maker
- [ ] Dust.js
- [ ] Edn
- [x] Emacs Lisp
- [x] Elixir
- [x] Elm
- [ ] Elvish
- [x] Emojicode
- [x] Erlang
- [ ] FEN
- [x] F#
- [x] Fish
- [x] Forth
- [x] FORTRAN Legacy
- [x] FORTRAN Modern
- [ ] FreeMarker
- [ ] F*
- [ ] Futhark
- [ ] GDB Script
- [x] GDScript
- [ ] Gherkin (Cucumber)
- [ ] Gleam
- [x] GLSL
- [x] Go
- [ ] Go HTML
- [x] GraphQL
- [x] Groovy
- [ ] Gwion
- [ ] Hamlet
- [ ] Handlebars
- [ ] Happy
- [x] Haskell
- [x] Haxe
- [x] HCL
- [ ] Headache
- [ ] HEX
- [x] HLSL
- [x] HolyC
- [x] HTML
- [x] Idris
- [ ] INI
- [ ] Intel HEX
- [ ] Isabelle
- [ ] JAI
- [x] Java
- [x] JavaScript
- [x] JSON
- [x] Jsonnet
- [x] JSX
- [x] Julia
- [ ] Julius
[...]

PR #495, opened 2021-10-01
Add WebAssembly language support

dnaka91:
This adds support for WebAssembly as it's currently missing according to #490.
I'm not the greatest ASCII artist but tried my best to re-create the original logo.


o2sh:
You did great :muscle: 
Thanks a lot for your contribution @dnaka91

PR #496, opened 2021-10-01
Add TOML language support

dnaka91:
Adding support for the TOML configuration language.


o2sh:
The ASCII art looks great! with d1a37d1, I took the liberty to  replace the Color::Black with Color::White in order to make it more visible on most terminals - with black backgrounds.

dnaka91:
Sure no problem. Was wondering the same about the black color but sticked with the original logo colors.
I'll try making a few more missing ASCII logos over the weekend, maybe sending them all in a single PR instead of making one for each logo if that's fine with you.

PR #497, opened 2021-10-01
removed json dependency

meguminloli:
removed json dependencies and updated some other dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -719,12 +719,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "json"
-version = "0.12.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "078e285eafdfb6c4b434e0d31e8cfcb5115b651496faca5749b88fafd4f23bfd"
-
-[[package]]
 name = "lazy_static"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -944,7 +938,6 @@ dependencies = [
  "colored",
  "git2",
  "image",
- "json",
  "libc",
  "more-asserts",
  "paste",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ color_quant = "1.1.0"
 colored = "2.0.0"
 git2 = {version = "0.13.22", default-features = false}
 image = "0.23.14"
-json = "0.12.4"
 regex = "1.4.6"
 serde = "1.0.130"
 serde_json = "1.0.68"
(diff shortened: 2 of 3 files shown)


spenserblack (commented):
>and updated some other dependencies

It's good to keep dependencies up-to-date, but we use dependabot (e.g. #488), so we don't really need to manually update dependencies 🙂 

*:warning: __Warning:__ long rant ahead*

I believe it is generally best to keep unrelated changes out of a commit/PR. One of the uses of commits is to find *which change introduced a bug* (I trust that this PR doesn't introduce any bugs, though :smile: ). There are several ways to find the commit that introduced a bug, like `git bisect` (this is a very cool tool that sadly flies under the radar of many git users, I highly recommend looking it up). But when unrelated changes are included in a commit, it becomes more difficult to find *exactly* what's wrong.
Let's say, for example, that this PR *did* introduce a bug, and it was caused by updating regex. If we did a `git bisect` and found commit eed3ccd, this is the diff that we would see.
```diff
- git2 = {version = "0.13.22", default-features = false}
+ git2 = {version = "0.13.23", default-features = false}
image = "0.23.14"
- json = "0.12.4"
- regex = "1.4.6"
+ regex = "1.5.4"
```
Looking at that diff, it isn't immediately obvious which change introduced a bug. But if you split this into 3 commits, `remove json`, `update regex`, and `update git2`, this is the diff that we would see.
```diff
- regex = "1.4.6"
+ regex = "1.5.4"
```

As a general rule that I use personally, if I have to use "and" to describe a commit (e.g. "removed json and updated other dependencies"), it might be worth separating into multiple commits. Don't take this rule too strictly, though.

tl;dr I would recommend just removing the `json` dependency, and leaving `git2` and `regex` alone.

meguminloli:
> I believe it is generally best to keep unrelated changes out of a commit/PR. One of the uses of commits is to find _which change introduced a bug_ (I trust that this PR doesn't introduce any bugs, though 😄 ). There are several ways to find the commit that introduced a bug, like `git bisect` (this is a very cool tool that sadly flies under the radar of many git users, I highly recommend looking it up). But when unrelated changes are included in a commit, it becomes more difficult to find _exactly_ what's wrong. Let's say, for example, that this PR _did_ introduce a bug, and it was caused by updating regex. If we did a `git bisect` and found commit [eed3ccd](https://github.com/o2sh/onefetch/commit/eed3ccddee7af1a866fafac707b58870cb9bfa86), this is the diff that we would see.
Sorry for taking up your time and thanks for the advice, I'll keep it in mind in the future. I reversed the dependency update and left only the removal of the `json` dependency.

spenserblack:
>Sorry for taking up your time

Don't worry about it, I chose to write that long rant when I could've just written the tl;dr. Sorry for taking up your time by making you read all that :rofl:

PR #498, opened 2021-10-02
Add Jsonnet language support

remram44:
This is a tough logo but I did my best:

original logo

ASCII art logo


o2sh (approved):
Yeah, working with perspective in ASCII is tricky especially with the size constraint.
Still, I think it looks pretty good.
Thanks @remram44

PR #499, opened 2021-10-02
Add YAML language support

remram44:
Tried a few options but went with number 3:

ASCII art


o2sh (approved):
Option 3 looks great :+1: 
Thanks @remram44

PR #505, opened 2021-10-14
Add support for json

gahag:
Based on this logo:
image

Sample output:
image


o2sh:
Thanks a lot for your contribution @gahag :tada: 
with 69e2861, I took the liberty to invert the colors to make the logo more visible on black terminals - which is the de facto standard.

PR #506, opened 2021-10-15
Add SASS language support

alessandroasm:
Adds support to SASS as it's listed as missing on #490.


alessandroasm:
I'm glad to help :+1:

PR #527, opened 2021-11-08
Add workflow for Minimum Rust Version badge

### LINKED ISSUE #525: Document minimum Rust version?
There have been a few issues caused by incompatible Rust versions (#183 and #128). It might be helpful if the minimum required Rust version was documented somewhere.

I wrote [this action](https://github.com/spenserblack/actions-msrv) recently (it's just a wrapper around [cargo-msrv](https://github.com/foresterre/cargo-msrv)), which might be helpful if we decide to try this.

spenserblack:
Resolves #525

This CI is very slow, taking about 15 minutes to complete. To help reduced conflicts, I added a `git pull` to the script that commits the badge.
However, this slowness might be something to consider for the workflow triggers. Perhaps the trigger should be on release instead?


o2sh:
> Perhaps the trigger should be on release instead?

Agreed :+1:  
I took care of it in e42dbb4 among other minor changes.

If the PR is good for you I'll let you trigger the workflow  to generate a new badge before merging.

spenserblack:
Changes look good! My only change after those was changing the literal `+` in the shields.io URL to be percent-encoded.
Badge generated.

Issue #526, opened 2021-11-07
Use time crate instead of chrono


o2sh:
#516 #515 

The `chrono` crate does not seem to be [actively maintained](https://www.reddit.com/r/rust/comments/qamgyh/is_the_chrono_crate_unmaintained/hh6h36h/).

We may have to try and replace the `chrono` crate with `time` where possible.
As a result, we will have to port some of [chrono-humanize](https://gitlab.com/imp/chrono-humanize-rs)'s logic and integrate it into onefetch inside a `time.rs` module.

HallerPatrick:
How about using [humantime](https://github.com/tailhook/humantime) as an alternative für `chrono-humanize`?  It states that it is based on `time`. 

I could look into it to see if it worsk how we want it / how `chrono-humanize` did it.

o2sh:
Actually, humantime is a _"A parser and formatter for **std::time**::{Duration, SystemTime}"_ which only relies on `time` (and `chrono`) for benchmarks and unit tests. This makes it a perfect candidate for what we intend to do.

So yeah, fingers crossed 🤞

HallerPatrick:
You are right. Its only good for parsing human time. I misread that.

@o2sh do you know why we apply the offset for timezones to the elapsed time till creation [here](https://github.com/o2sh/onefetch/blob/689a0ab8ed92fad7743a72775d709ae43bc84718/src/info/repo.rs#L339-L355)? And why it is `west`. As long as the repo to display is not created hours ago, those +-5 hours are not relevant. But e can imitate this by adding the required offset as hours if needed. 
What seems harder is to humanize the duration. I looked into the source of `chrono-humanize` and its not too complicated, but also not trivial enough to just be added to the `onefetch` IMO. 

Making a new crate: `humanize-time` would be relatively easy to port from `chrono-humanize`, but is still some work.

I think about making this a new repo/crate.

o2sh:
> @o2sh do you know why we apply the offset for timezones to the elapsed time till creation here? And why it is west. As long as the repo to display is not created hours ago, those +-5 hours are not relevant.

If I'm not mistaken, this logic was added in one of your PR #341

> What seems harder is to humanize the duration. I looked into the source of chrono-humanize and its not too complicated, but also not trivial enough to just be added to the onefetch IMO.

Either that or just add a `time.rs` module inside the project. Whatever works best for you :+1:

HallerPatrick:
[...]

PR #533, opened 2021-11-17
Using time crate instead of chrono

HallerPatrick:
This PR solves Issue #526.

I ported, more or less, the `chrono-humanize` library to only use the std lib time and as a optional feature the `time-rs` crate ( for display iso time). Because I wanted to keep most of tests of the original repo (which are a lot!) and also included some nice interface, I made this an [own crate](https://github.com/HallerPatrick/time-humanize).

The code now even looks way cleaner and include less logic. I tested the human readable output as well as the iso output on different repos. Feel free to test this  as well.


HallerPatrick:
Looks like the audit finds a soundness issue for `chrono` and `time`. `time` is an optional feature of `time-humanize`, that is used to display the time as the iso standard. I am not sure if this is affecting our use case.

Cargo.toml
```diff
@@ -30,8 +30,8 @@ anyhow = "1.0"
 askalono = "0.4.3"
 byte-unit = "4.0.12"
 bytecount = "0.6.2"
-chrono = "0.4"
-chrono-humanize = "0.2.1"
+time-humanize = {version="0.1.2", features = ["time"]}
+time = { version = "0.3.5", features = ["formatting"]}
```
spenserblack, line 34:
```suggestion
time-humanize = { version="0.1.2", features = ["time"] }
time = { version = "0.3.5", features = ["formatting"] }
```
Minor nitpick. I don't think we have a preferred style, but I've usually seen spaces inside braces.
For example, [the docs for dependencies in `Cargo.toml`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries).

src/info/repo.rs
```diff
@@ -337,23 +339,21 @@ fn bytes_to_human_readable(bytes: u128) -> String {
...
-            .to_rfc3339_opts(chrono::SecondsFormat::Secs, true)
+        to_rfc3339(HumanTime::from(time.seconds()))
     } else {
-        let ht = HumanTime::from(dt_with_tz);
+        let ht = HumanTime::from_duration_since_timestamp(time.seconds().unsigned_abs());
         format!("{}", ht)
     }
 }
 
+fn to_rfc3339<T>(dt: T) -> String
```
spenserblack, line 350: It might be cool if we added some tests for these functions. Unfortunately we didn't have any prior tests for duration formatting to confirm consistent behavior.

spenserblack (commented):
Looks like `chrono` is still in our dependency tree because it's a build dependency of `tokei`, but I think that's OK.
I didn't see the advisory for `time` that you mentioned, though.

Cargo.toml
spenserblack: Actually ignore the above nitpick, since my suggestion doesn't match the rest of `Cargo.toml` :sweat_smile: 
If anything, it would be more like
```suggestion
time-humanize = {version="0.1.2", features = ["time"]}
time = {version = "0.3.5", features = ["formatting"]}
```

src/info/repo.rs
HallerPatrick: Yes, can do that

spenserblack:
Sorry @HallerPatrick, I gave you a bad suggestion. See https://github.com/o2sh/onefetch/pull/533#discussion_r751300809

HallerPatrick:
> Looks like `chrono` is still in our dependency tree because it's a build dependency of `tokei`, but I think that's OK.
> I didn't see the advisory for `time` that you mentioned, though.

In the audit there is a reference link, which leads to the `time-rs` repo. `chrono` and `time` have the same unsoundness bug. [time-rs issue](https://github.com/time-rs/time/issues/293) , [chrono issue](https://github.com/chronotope/chrono/issues/499)

spenserblack:
Got it, thanks for linking the issue!
According to [this security advisory](https://github.com/time-rs/time/security/advisories/GHSA-wcg3-cvx6-7396) and https://github.com/time-rs/time/issues/293#issuecomment-731504605 it seems that `time >= 0.2.23` won't be affected. In this PR both `onefetch` and `time-humanize` depend on `0.3.5`, so it looks OK in my opinion.
The affected `time` version (`0.1`) comes from chrono, which through the chain of dependencies is a build dependency of tokei. I think that it's out of our scope to fix deeply nested build dependencies of one of our dependencies :slightly_smiling_face: 

tl;dr it doesn't seem like we have to worry about `time`'s security advisory IMO

src/info/repo.rs
HallerPatrick: @spenserblack You think those tests are sufficient?

spenserblack: Yeah, those look good to me :+1:
Thanks!

spenserblack (approved):
Thanks for adding the tests, LGTM!

HallerPatrick:
@o2sh does not seem to be active currently. Because I got your okay, I would merge this now

o2sh:
So sorry I couldn't find the time to review your PR @HallerPatrick :disappointed: 
Thanks @spenserblack for providing great feedback as always :muscle: 

Everything seems to work as expected. 

My only concern is the `edition = "2021"` used by [time-humanize](https://github.com/HallerPatrick/time-humanize/blob/ef3668f61cdbd327f6a9d360b786ab36c865d1e4/Cargo.toml#L8) which makes onefetch's msrv (minimum supported rust version) go from 1.50.0 to 1.56.1

Rust version 1.56.1 is [very recent](https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html) and not yet available for [launchpad](https://launchpad.net/ubuntu/+source/rustc)

HallerPatrick:
> So sorry I couldn't find the time to review your PR @HallerPatrick :disappointed: 
> 
> Thanks @spenserblack for providing great feedback as always :muscle: 
> 
> 
> 
> Everything seems to work as expected. 
> 
> 
> 
> My only concern is the `edition = "2021"` used by [time-humanize](https://github.com/HallerPatrick/time-humanize/blob/ef3668f61cdbd327f6a9d360b786ab36c865d1e4/Cargo.toml#L8) which makes onefetch's msrv (minimum supported rust version) go from 1.50.0 to 1.56.1
> 
> 
> 
> Rust version 1.56.1 is [very recent](https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html) and not yet available for [launchpad](https://launchpad.net/ubuntu/+source/rustc)

No problem at all. @spenserblack gave good feedback.

For `time-humanize`: I was not aware of the Edition difference. I will look to lower it to `onefetch`s version if possible.

HallerPatrick:
@o2sh I updated the rust version of `time-humanize` to `edition = "2018"`. Would you update the dependency to `time-humanize = { version="0.1.3", features = ["time"] }`? Or should I do this through a PR?

o2sh:
Thanks @HallerPatrick 👍 
I think we can wait for the dependabot to take care of it 🤖

PR #548, opened 2021-12-01
Update MSRV badge and set timeout

### LINKED ISSUE #544: 'Generate MSRV Badge ' workflow times out
The 'Generate MSRV Badge' has been [timing out](https://github.com/o2sh/onefetch/actions/workflows/msrv-badge.yml) systematically since the last release.

spenserblack:
With https://github.com/spenserblack/actions-msrv/commit/a14a47dac66b936655e9f7714723c08e0ae27f02, the MSRV search is now linear, which [completed in a reasonable amount of time (~15 minutes)](https://github.com/spenserblack/onefetch/runs/4386733386?check_suite_focus=true).

This also sets a timeout for MSRV-getting step, so that it will stop before it takes up a huge chunk of workflow run minutes.

(indirectly) closes #544


o2sh:
LGTM 👍 
Now, we just need to wait for Dependabot to pick up the new version, 
Once this is good, I will run locally and confirm that it fixes #544 fingers crossed 🤞

spenserblack:
Sounds good!

>I'll wait for Dependabot to pick up the new version.

Dependabot shouldn't actually update this, because I moved the `v0.2` tag to be the same commit as `v0.2.1` :smile:

PR #560, opened 2022-01-01
Add Ren'Py language support

TabulateJarl8:
I was bored so I decided to add support for Ren'Py since it didn't seem to already be in there. Feel free to make suggestions of possible symbol changes in the logo.


o2sh:
That's a difficult one, I tried to change some of the characters to make it stand out more while simplifying the design a little bit. Tell me what you think.

before:
renpy
after:
renpy3

TabulateJarl8:
Oh yeah I think that definitely stands out a lot more, I like it. Never thought to use that character somehow.

o2sh:
final form:
Screenshot 2022-01-01 at 23 13 37

Issue #584, opened 2022-02-15
Colorful language breakdown


spenserblack:
### Discussed in https://github.com/o2sh/onefetch/discussions/582

<div type='discussions-op-text'>

<sup>Originally posted by **spenserblack** February 14, 2022</sup>
Just an idea that popped into my head. Making discussion to see if its worth doing.

I'm thinking that, like how GitHub uses linguist to have a colored bar of languages, we could do the same. Each of the languages currently has a "primary" color (the 1st defined color), and we already break down the percentages, so most of the work is already done. Like the current color bar, it would be a series of colored blocks.

For example, if a project is 50% `JavaScript`, 40% `Html`, and 10% `Sh`, then there would be 5 yellow blocks, 4 red blocks, and 1 green block.

The biggest downside that I can think of is that there could be some confusion/clutter with both a language breakdown colored bar and the `neofetch`-like color bar.</div>

Per the linked discussion, the current idea is that a GitHub-like colored breakdown of the top languages, with a legend mapping the colored blocks to their languages, would be a fun feature to have. See discussion for more details.

PR #585, opened 2022-02-19
add language bar

o2sh:
#584 

Screenshot 2022-02-19 at 23 21 45

Remarks:

- The language bar is 26 characters long (arbitrary value). However, in some cases it can be longer. Indeed, if a language's weight in the distribution is too small to account for a whole character, I do a cmp::Max(val, 1) so that, it will still appear in the language bar. This may not be optimal 🤔 

- As discussed before, if the terminal doesn't support true color, we fallback to a default color palette:

Screenshot 2022-02-19 at 23 49 27

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -304,27 +304,61 @@ impl Info {
     }
 
     fn get_language_field(&self, title: &str) -> String {
-        let mut language_field = String::from("");
-
+        let progress_bar_length = 26;
         let pad = title.len() + 2;
+        let color_palette = vec![
+            Color::Red,
+            Color::Green,
+            Color::Yellow,
+            Color::Blue,
+            Color::Magenta,
+            Color::Cyan,
+        ];
 
-        let languages: Vec<(String, f64)> = {
-            let mut iter = self.languages.iter().map(|x| (format!("{}", x.0), x.1));
+        let mut language_field = String::from("");
+        let mut i = 0;
+        let languages: Vec<(String, f64, Color)> = {
+            let mut iter = self.languages.iter().map(|x| {
+                i += 1;
+                let color = if self.config.true_color {
+                    x.0.get_colors(true)[0]
+                } else {
+                    color_palette[i % color_palette.len()]
+                };
+                (format!("{}", x.0), x.1, color)
+            });
             if self.languages.len() > 6 {
                 let mut languages = iter.by_ref().take(6).collect::<Vec<_>>();
                 let other_sum = iter.fold(0.0, |acc, x| acc + x.1);
-                languages.push(("Other".to_owned(), other_sum));
+                languages.push(("Other".to_owned(), other_sum, Color::White));
                 languages
             } else {
                 iter.collect()
             }
         };
 
+        for (_, language) in languages.iter().enumerate() {
+            let bar_width = std::cmp::max(
+                (language.1 / 100. * progress_bar_length as f64).round() as usize,
+                1,
+            );
+            language_field.push_str(&format!(
+                "{:<width$}",
+                "".on_color(language.2),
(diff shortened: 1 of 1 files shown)


src/info/mod.rs
```diff
@@ -304,27 +304,58 @@ impl Info {
     }
 
     fn get_language_field(&self, title: &str) -> String {
-        let mut language_field = String::from("");
-
+        let language_bar_length = 26;
```
spenserblack, line 307: Based on the screenshot, this seems to line up nicely. If this was 29 characters, it might be more symmetrical with the ASCII art, which maxes out at 40 characters wide. (29 + `"Languages: ".len()`).

src/info/mod.rs
```diff
@@ -304,27 +304,58 @@ impl Info {
...
+            Color::Cyan,
+        ];
 
-        let languages: Vec<(String, f64)> = {
-            let mut iter = self.languages.iter().map(|x| (format!("{}", x.0), x.1));
+        let mut language_field = String::from("");
+        let mut i = 0;
+        let languages: Vec<(String, f64, Color)> = {
+            let mut iter = self.languages.iter().map(|x| {
+                i += 1;
```
spenserblack, line 322:
```suggestion
        let languages: Vec<(String, f64, Color)> = {
            let mut iter = self.languages.iter().enumerate().map(|(i, x)| {
```

src/info/mod.rs
```diff
@@ -304,27 +304,58 @@ impl Info {
...
                 let other_sum = iter.fold(0.0, |acc, x| acc + x.1);
-                languages.push(("Other".to_owned(), other_sum));
+                languages.push(("Other".to_owned(), other_sum, Color::White));
                 languages
             } else {
                 iter.collect()
             }
         };
 
+        for (_, language) in languages.iter().enumerate() {
```
spenserblack, line 340:
```suggestion
        for language in languages.iter() {
```
The index isn't used, so does `enumerate` have to be called?

src/info/mod.rs
```diff
@@ -304,27 +304,58 @@ impl Info {
...
+            let bar_width = std::cmp::max(
+                (language.1 / 100. * language_bar_length as f64).round() as usize,
+                1,
+            );
+            language_field.push_str(&format!(
+                "{:<width$}",
+                "".on_color(language.2),
+                width = bar_width
+            ));
```
spenserblack, line 349: Instead of pushing strings in a loop, you might want to map `languages.iter()` to an iter of `String`s and then join/collect them. Otherwise, we have a rough idea of how many characters are needed (`language_bar_length` + ANSI escapes + extra characters depending on `std::cmp::max`), so we might want to use `String::with_capacity` to prevent reallocations.

spenserblack (approved):
This looks awesome!

Minor nitpicks, but overall LGTM :+1:

src/info/mod.rs
o2sh: If the language bar length could be fixed to 29, this would indeed make sense. However with the possible extra characters, this value can go up to 35, and if we add `"Languages: ".len()` up to 46.

src/info/mod.rs
o2sh: cf. 72aa83c, is this what you had in mind ?

spenserblack: Sorry, I meant using `with_capacity` if *not* collecting. AFAIK Rust's compiler is pretty smart about collecting lazy iterators of strings, and will efficiently allocate the necessary memory. https://github.com/o2sh/onefetch/pull/585/commits/64929ef226173013a90bddf519e25c2ece2b9889 is an example of what I was thinking.

spenserblack (approved):
Thanks for addressing suggestions!

Issue #33, opened 2019-05-01
hard to read on light background


bugabinga:
In terminal emulators with light backgrounds, most of `onefetch`s output is unreadable.

o2sh:
Yeah, unfortunately as dark colors don't show well on black background, light colors are hard to read on white background.

I don't see how i could please everyone...

Thx for your feedback @bugabinga

bugabinga:
> I don't see how i could please everyone...
That is hard indeed :)

However, how about these alternatives:
Offer flags for dark/light colors

```sh
$ onefetch --color=auto|light|dark|none
```
Maybe there is a way to use the default colors of the terminal emulator, though I admit, I don't know.

skyfaller:
One possible solution is to have a theme that uses ansi colors, as bat does: https://github.com/sharkdp/bat/issues/490

Delta uses more or less the same solution because it uses bat themes: https://github.com/dandavison/delta/issues/447#issuecomment-808821016

Is creating an ansi theme or using bat's ansi theme possible for onefetch?

spenserblack:
Themes definitely sounds like a fun idea! I would consider it a new feature instead of a resolution to this issue. Mainly because I consider this issue resolved already due to the below points.

- onefetch supports defining your own colors via command-line options
- isn't this more of an issue of terminal configuration than onefetch output? Out of curiosity I tried `onefetch --true-color never` with Gruvbox Light, and it looked fine to me

Since `bat` and `delta`'s themes are for syntax highlighting, I think our themes would be slightly different. No ASCII has more than 9 colors currently (`swift.ascii` has 9, all shades of red though). I think it would be better if this could be accomplished *without* changing the language macro rule and calls, which are quite long.
https://github.com/o2sh/onefetch/blob/bc09db57a03bbe8c4947f52b870a54dc2e707ffe/src/info/langs/language.rs#L175-L323

So the way I imagine themes working is that colors would be defined as primary, secondary, tertiary, etc., and these would replace color `{0}`, color `{1}`, `{2}`, etc. We probably don't want to define 9 colors since most langs use less than half of that, so maybe 5 colors at maximum, and `{5}`, `{6}`, etc. would either use their original color or the primary color.

skyfaller:
> isn't this more of an issue of terminal configuration than onefetch output? Out of curiosity I tried `onefetch --true-color never` with Gruvbox Light, and it looked fine to me

[...]

PR #593, opened 2022-02-23
Add `_` as possible color value

spenserblack:
Allows user to disable the color and instead use the default terminal
foreground color, which can be different from the 16 available colors.

Related: https://github.com/o2sh/onefetch/issues/33#issuecomment-1048968541

Ideally, a `Color::None` variant could be added to `colored`, but IDK if the crate is still maintained (last commit was a while ago).

******

While I personally like `_` as the "don't care" value, I'm open to suggestions :slightly_smiling_face: 

Also open to suggestions on best way to document this.

Note that this only works with text colors, since `None` values are ignored for the ASCII art. So this PR probably needs some adjustment: either remove the possible value for `-c`, or add support for disabling colors for the ASCII art.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -46,7 +46,7 @@ impl Config {
         let possible_backends = ["kitty", "iterm", "sixel"];
 
         let color_values = &[
-            "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15",
+            "_", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15",
         ];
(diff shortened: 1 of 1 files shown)


o2sh:
Maybe we should update `TextColor` to this:

```
pub struct TextColor {
    pub title: Color,
    pub tilde: Option<Color>,
    pub underline: Option<Color>,
    pub subtitle: Color,
    pub colon: Option<Color>,
    pub info: Option<Color>,
}
```

and have tilde , underline, colon and info default to the terminal foreground color (None) and be Some(value) if the user decides to specify its own colors with `--text-colors`.

spenserblack:
Yeah, that's a good idea.

spenserblack:
Closed by #604

PR #604, opened 2022-03-05
Default to terminal foreground color for tilde, underline, colon and info

o2sh:
As suggested [here](https://github.com/o2sh/onefetch/pull/593#issuecomment-1049336777)

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -1,6 +1,6 @@
 use crate::cli::{self, Config};
 use crate::ui::get_ascii_colors;
-use crate::ui::text_color::{color_if_some, TextColor};
+use crate::ui::{text_color::TextColor, ColorizeOption};
 use anyhow::Result;
 use author::Author;
 use colored::{Color, ColoredString, Colorize};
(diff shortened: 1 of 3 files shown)


spenserblack (approved):
Looks good!

I think this is good to merge as-is, but you might want to check out https://github.com/spenserblack/onefetch/commit/42711cfe45490cad9880c5ac9ba32f74362173e1.

Maybe I over-engineered a bit, but the basic idea was to create our own `Colorize` trait that "inherits" from `colored::Colorize`, allowing us to pass options and keep the style of `s.fn` vs `fn(s)`.

[diff to `main`](https://github.com/o2sh/onefetch/compare/main...42711cfe45490cad9880c5ac9ba32f74362173e1)

spenserblack (approved):
That was probably a good idea to rename to `try_color` to avoid naming conflicts :+1:

o2sh:
Thanks @spenserblack, and you're right `s.fn` instead `fn(s)` is indeed cleaner. I took inspiration from what you did and added the ColorizeOption but I don't think we need to override the whole Colorize interface.

PR #609, opened 2022-03-11
Add Wolfram language support

rcon56:
The ascii image mainly consists of 'W' and 'O' because the logo of [Wolfram Language](https://www.wolfram.com/language/) is a WOLF and wolves howl like 'woooooooooo' I guess 👀. 

Here is the screenshot in dark background with true color. 
Wolfram-Screenshot


spenserblack:
Great logo! I'm a bit concerned about how this would look in light terminals (see https://github.com/o2sh/onefetch/issues/33#issuecomment-1048073499 and below comments for recent conversation). But I suppose the wolf would just become negative space on a light terminal, which is OK. But if you can think of a reasonable alternative to the white color, that would be appreciated. :wolf:
You also might want to experiment with filling in the outline with black text, so that the wolf would still have a black outline on light terminals.

(Overall I think we'll need to replace white with `None` colors in logos to use the terminal's default text color)

rcon56:
> Great logo! I'm a bit concerned about how this would look in light terminals (see [#33 (comment)](https://github.com/o2sh/onefetch/issues/33#issuecomment-1048073499) and below comments for recent conversation). But I suppose the wolf would just become negative space on a light terminal, which is OK. But if you can think of a reasonable alternative to the white color, that would be appreciated. 🐺 You also might want to experiment with filling in the outline with black text, so that the wolf would still have a black outline on light terminals.
> 
> (Overall I think we'll need to replace white with `None` colors in logos to use the terminal's default text color)

Thanks for your suggestion! I have tried what you mentioned above in terminal with different themes and found that:
- Using default text color for the main body of 🐺  might be the best solution for now. (AFAIK, the feature of `None` color has not been merged into the main branch or I just overlooked?)
- Using grey can be acceptable in most situations.
- However, I think adding black text to represent the border will break the integrity of the ascii image because it's hard to portray such a thin borderline by ascii characters.

spenserblack:
>AFAIK, the feature of `None` color has not been merged into the main branch or I just overlooked?

That's right, that was more of a note-to-self I guess. I should probably get started on that :slightly_smiling_face: 

>Using grey can be acceptable in most situations

That sounds good for TrueColor, but I don't think it's possible with the 8 standard colors.

>I think adding black text to represent the border will break the integrity of the ascii image

I see what you mean.

I just had an idea for a relatively painless way to implement `None` colors, so I think I'll hold off on this PR for a bit and see if I can get that sorted out.

rcon56:
> #625 has been merged, so please merge it into your branch and check out the logo 😃

Great! I have merged it and also changed the white color to the grey one in the _true color_ scheme.

Here's what I got in light terminal :)
image

spenserblack (approved):
Looks great! Thanks for your patience!

o2sh:
@spenserblack Maybe we should also map `Rgb(255, 255, 255)` to `Default`?

spenserblack:
I was thinking about that in #625, but here's why I hesitated:
- True color IMO should be the "real" color of the logo. If the user wants control over that (light theming, changing contrast), they can use `--true-color never`. 
- True color is where contributors get to express more artistic freedom, like creating gradients. I could anticipate a gradient with a `#FFFFFF` start or end, which would be interfered with by a `Default` remapping. That's just an example, but I'm concerned about interfering with a contributor's intent when true color kind of promises that they can do whatever they want.
- If we remap `#FFFFFF` (255, 255, 255), that opens the door to requests to remap `#FEFFFF` and down. Possibly any color brighter than `#CCCCCC` (roughly 75% brightness) would be on the table.

TBH I think a note in either the README of Wiki stating something like
>If you use a light terminal theme, struggle with high contrast colors, or for any other reason need to limit `onefetch` to the 8 ANSI colors, you can use `onefetch --true-color never`. For regular usage you can add this to your `.bashrc`: `alias onefetch='onefetch --true-color never'`

would be enough.

tl;dr I've been thinking of the basic colors as "compatibility mode" and true colors as "pretty mode".

PR #610, opened 2022-03-12
fix Lua logo so that it can be read even without color

yutkat:
I am calling onefetch from on Neovim and I am concerned that the Lua logo will be unreadable.
I modified it so that I could see it without the color.

before:

after:


o2sh:
hmm, I've never considered this use case, we may have other logos like this.

Anyway, I took the opportunity to do some minor retouching the Lua ASCII logo to make it more symmetrical:

lua

PR #611, opened 2022-03-14
Allow for optional colors

spenserblack:
This was originally going to be a quick and painless change using an `OptionalColor` type and `.into()`, but it quickly turned into way more changes than expected :sweat_smile: Hopefully nothing has broken.

The last commit is `TEMP` because it seems like `strum` is naively using `None` instead of `Option::None` in its macro definition, breaking that commit. https://github.com/spenserblack/onefetch/commit/5142c5fd2f89aad1fe394c4abce09d5026e000d8 is OK to review, though. I'll make a PR to `strum`, but, assuming https://github.com/spenserblack/onefetch/commit/5142c5fd2f89aad1fe394c4abce09d5026e000d8 is good, ~~it's ready to be merged if we don't want to wait~~.

Edit: Nevermind, tests fail on that commit.

Edit2: Force-pushed, 0ef1c64019ff52ff18cbb9263dd9259fbdf79556 should be OK.

Edit3: https://github.com/Peternator7/strum/pull/214 merged, on next release this whole PR should be ready to review.

Edit4: No need for a new strum release. I realized that `Fg` is a more descriptive variant than `None`.


o2sh:
Is this really a good idea to alter the logo's original colour palette? IMO, this may help a small portion of users on light theme terminals in terms of readability but all users will lose in terms of aesthetics.

spenserblack:
Well the main reason was to allow the logo's "white" color to just be the text color. I agree that this is a massive change, but we've gotten a few notes that onefetch doesn't look good on light terminals, and one user pointed out that

>neofetch has no problem with this

(Source: https://github.com/o2sh/onefetch/issues/33#issuecomment-1049037192)

It appears that Neofetch doesn't really use the white color. When the default color is white, or the selected color with `--ascii_colors` is white, it actually leaves the text uncolored, instead using the default foreground color.
https://github.com/dylanaraps/neofetch/blob/ccd5d9f52609bbdcd5d8fa78c4fdb0f12954125f/neofetch#L4725-L4728

I think we have 2 options:
1. Make a stance that onefetch is preferred on dark terminal themes
2. Alter the logo colors

spenserblack:
CI is failing from clippy test unrelated to this PR

spenserblack:
Or I suppose we could just make the `Fg` variant available (but currently unused), and just fix existing logos as light terminal users point them out.

src/info/langs/language.rs
```diff
@@ -2,14 +2,38 @@ use colored::Color;
...
+    true_colors: Option<Vec<OptionalColor>>,
 }
 
+macro_rules! define_optional_color {
+    ($( $color:ident ),*) => {
+        #[derive(Debug)]
+        pub enum OptionalColor {
+            $( $color, )*
+            TrueColor{ r: u8, g: u8, b: u8 },
+            Fg,
```
o2sh, line 18: We should maybe add a comment to explain what is `Fg` for.

src/info/langs/language.rs
```diff
@@ -2,14 +2,38 @@ use colored::Color;
...
+                    $( OptionalColor::$color => Some(::colored::Color::$color), )*
+                    OptionalColor::TrueColor{ r, g, b } => Some(::colored::Color::TrueColor{r, g, b}),
+                    OptionalColor::Fg => Option::None,
+                }
+            }
+        }
+    }
+}
+
+define_optional_color!(Black, Red, Green, Yellow, Blue, Magenta, Cyan, White);
```
o2sh, line 33: Maybe remove White from the list

src/info/langs/language.rs
```diff
@@ -2,14 +2,38 @@ use colored::Color;
...
-    basic_colors: Vec<Color>,
-    true_colors: Option<Vec<Color>>,
+    basic_colors: Vec<OptionalColor>,
+    true_colors: Option<Vec<OptionalColor>>,
 }
 
+macro_rules! define_optional_color {
+    ($( $color:ident ),*) => {
+        #[derive(Debug)]
+        pub enum OptionalColor {
```
o2sh, line 15: I'm not sur about the name, what about `AsciiColor` instead

o2sh:
You're right, I guess I was just reluctant to add more complexity to the code that would only be beneficial to a small fraction of users. But, what you did is actually pretty clean and as you pointed neofetch does exactly that.

src/info/langs/language.rs
spenserblack: Yeah, that would work (and be more similar to neofetch). Perhaps `White` and `Fg` can be effective aliases of each other?

src/info/langs/language.rs
spenserblack: Yeah, this was back when there was a `None` variant. `AsciiColor` sounds good to me.

spenserblack:
> I was just reluctant to add more complexity to the code that would only be beneficial to a small fraction of users

That makes sense. I wouldn't be opposed to delaying any decision on this PR until we get a good idea of how many users really need this (I'm of the opinion that a good light theme would reverse black and white TBH, but it looks like some popular ones don't do that).

By making `White` ignored as you suggested (https://github.com/o2sh/onefetch/pull/611#discussion_r829604856), some complexity could probably be reduced. Perhaps `ui::ascii_art::add_colored_segment` could use a wrapper type that implements `Colorize` and ignores `White`.

owo-colors also looks like it has an interesting way to colorize that could probably greatly reduce complexity, as opposed to needing to rely on `colored`'s (limited) `Color` enum. I think [`Default`](https://docs.rs/owo-colors/latest/owo_colors/colors/struct.Default.html) would do what we need. We can make a const called `White` that aliases `Default` to reduce diff size and prevent confusion with new contributors.

o2sh:
> owo-colors also looks like it has an interesting way to colorize that could probably greatly reduce complexity, as opposed to needing to rely on colored's (limited) Color enum. I think [Default](https://docs.rs/owo-colors/latest/owo_colors/colors/struct.Default.html) would do what we need. We can make a const called White that aliases Default to reduce diff size and prevent confusion with new contributors.

This is exactly what we need. Let's do it :rocket:

spenserblack:
> Let's do it 🚀

Sounds good! I'll leave this PR open for now until the switch to owo-colors is merged. It's unlikely we'll reference commits from this specific PR, but better safe than sorry 🙂

PR #613, opened 2022-03-14
Make time test relative to current time

### LINKED ISSUE #612: Test `info::repo::tests::display_time_as_human_time_current_time_arbitrary` can fail
Running `cargo test`, it seems like this test can fail, at least on my machine. It returns `"2 years ago"` instead of the expected `"a year ago"`. I ran a bisect to make sure that there were no changes causing this test to fail, and bisect selected 3d7b14b856300b145f2b2dfec2ca62cb5c0785e8

spenserblack:
This test was failing after a year, because it was testing a conversion
of 2020/09/13 to "a year ago." This changes the test to be relative to
the current time, and subtract roughly a year.

Fixes #612

### DIFF
--- a/src/info/repo.rs
+++ b/src/info/repo.rs
@@ -364,7 +364,7 @@ mod tests {
 
     use super::*;
     use git2::Time;
-    use std::time::SystemTime;
+    use std::time::{Duration, SystemTime};
 
     #[test]
     fn display_time_as_human_time_current_time_now() {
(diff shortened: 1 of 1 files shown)


o2sh:
Maybe that was a [Pi Day](https://www.piday.org) easter egg by @HallerPatrick 😄 

Good catch! @spenserblack 🚀

HallerPatrick:
> Maybe that was a [Pi Day](https://www.piday.org) easter egg by @HallerPatrick 😄 
> 
> 
> 
> Good catch! @spenserblack 🚀 

Happy Pi Day! Thanks for finding my (completely intentional) Easter egg @spenserblack

Issue #605, opened 2022-03-07
Change coloring dependency


spenserblack:
Not something that needs to be done right now IMO, but something we might want to do eventually. `colored` hasn't been updated in a while, and may not be maintained anymore.

I'm creating this issue to document potential alternatives should we ever decide to switch to a different dependency (this isn't a definitive list, I'll add more as a find them 🙂)

- [`owo-colors`](https://github.com/jam1garner/owo-colors)

o2sh:
`owo-colors` seems like a great candidate for this with very low - to no - efforts on our part.

PR #620, opened 2022-03-19
Replace colored with owo-colors

o2sh:
#605


spenserblack (approved):
### 🚀 

This looks so much cleaner!

o2sh:
> This looks so much cleaner!

I agree, thanks for suggesting this library, it does exactly what we need and much more 💯 .

PR #625, opened 2022-03-22
Use `Default` color instead of `White`

spenserblack:
Allows the "white" color to have the greatest chance of being a dark,
visible color for light terminal themes. A macro is used to remap
`White` to `Default` to minimize size of the diff and prevent
contributors from needing to be familiar with the `Default` color on top
of the 8 standard colors.

For #33
Supersedes #611

### TODO

- [x] Review other uses of `White` where `Default` should be used

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -335,7 +335,7 @@ impl Info {
                 languages.push((
                     "Other".to_string(),
                     other_perc,
-                    DynColors::Ansi(AnsiColors::White),
+                    DynColors::Ansi(AnsiColors::Default),
                 ));
                 languages
             } else {
--- a/src/info/langs/language.rs
+++ b/src/info/langs/language.rs
@@ -8,9 +8,20 @@ pub struct Colors {
     true_colors: Option<Vec<DynColors>>,
 }
 
+/// Maps colors to preferred versions. Used to allow contributors to include
+/// colors with minimal confusion.
+macro_rules! clean_color {
+    ( White ) => {
+        clean_color!(Default)
+    };
+    ( $color:ident ) => {
+        DynColors::Ansi(AnsiColors::$color)
+    };
+}
+
 macro_rules! define_colors {
-    ( [ $($color:ident),+ ] ) => { Colors { basic_colors: vec![$(DynColors::Ansi(AnsiColors::$color)),+], true_colors: None } };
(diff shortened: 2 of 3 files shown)


spenserblack:
Sadly, I believe this does break the artistic intent of some logos on light terminals. For example, the Golang Gopher's eyes will have black sclera and white pupils on a light terminal. However, before this change, the white-on-white coloration of the gopher's eyes would give it the appearance of empty eye sockets.

I think that, to a certain extent, breakage of artistic intent is unavoidable on a light terminal, as many logos assume that empty space will be black. In other words, many of these logos were designed on and for a black "canvas".

o2sh:
>I think that, to a certain extent, breakage of artistic intent is unavoidable on a light terminal, as many logos assume that empty space will be black. In other words, many of these logos were designed on and for a black "canvas".

Yeah, dark terminals are the de facto standard and all logos should be designed with black "canvas" in mind first. Mapping `White` to `default` is the best compromise IMO.

spenserblack:
I didn't modify `num_to_color`. Even though this a function with [a very similar implementation in `neofetch`](https://github.com/dylanaraps/neofetch/blob/ccd5d9f52609bbdcd5d8fa78c4fdb0f12954125f/neofetch#L4725-L4728), `num_to_color` is pretty much only for user-specified overrides, and I think we can trust users to choose the correct foreground color.

src/info/mod.rs
```diff
@@ -335,7 +335,7 @@ impl Info {
                 languages.push((
                     "Other".to_string(),
                     other_perc,
-                    DynColors::Ansi(AnsiColors::White),
+                    DynColors::Ansi(AnsiColors::Default),
```
o2sh, line 338: `on_color()` with `Default` uses the default Terminal Bg color which makes it invinsible.

spenserblack: Oh, sorry about that. That should definitely be kept as `White` then.

spenserblack: Alternatively block characters can be used to take advantage of the foreground color, but that might have less compatibility among terminals.

spenserblack: Or perhaps it's OK if the block is invisible? It should always be a negligibly small percentage, after all?

o2sh: Not necessarily, let's say we have a repo of 10 languages with an equal distribution then `other` would account for ~50%. Besides, this is what Github does.

spenserblack: Good point, I didn't consider that. Unless we're using block characters, I think we'll have to switch back to `::White`, then.

I think white BG would be visible enough even on a light terminal, since there's no "alpha channel" like there is with ASCII characters. So I think just switch back to `::White` for that, what do you think?

PR #650, opened 2022-04-27
Add bug report issue template

spenserblack:
It seems like some users are using the language request issue template to submit other types of issues. Hopefully and additional issue template can help reduce that.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,19 @@
+---
+name: Bug Report
+about: Report a bug or unexpected behavior with onefetch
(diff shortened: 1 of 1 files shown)


o2sh:
Thanks @spenserblack 👍 
Another issue template we could add would be for feature requests ([example](https://github.com/extrawurst/gitui/blob/master/.github/ISSUE_TEMPLATE/feature_request.md)), like that we would cover most cases.

o2sh:
Thanks @spenserblack 👍 
Another issue template we could add would be for feature requests ([example](https://github.com/extrawurst/gitui/blob/master/.github/ISSUE_TEMPLATE/feature_request.md)), like that we would cover most cases.

PR #657, opened 2022-05-05
Add `--completion` option

spenserblack:
This allows the user to generate completions by passing `--completion <SHELL>`. This PR also

- Moves the CLI App creation to a separate `build_cli` function. This was necessary as the
  `clap_generate` docs seem to imply that a separate instance of the CLI app struct is necessary
  for generating completions. For the purposes of the App struct's lifetime, a few borrowed
  iterators had to be changed to be taken by the app struct.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -235,7 +235,6 @@ pub fn print_completions<G: Generator>(gen: G) {
 
 pub(crate) fn build_cli() -> clap::Command<'static> {
     let possible_backends = ["kitty", "iterm", "sixel"];
-    let possible_shells = Shell::possible_values();
 
     let color_values = &[
         "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15",
@@ -492,7 +491,7 @@ pub(crate) fn build_cli() -> clap::Command<'static> {
         .arg(
             Arg::new("completion")
                 .long("completion")
-                .possible_values(possible_shells)
+                .possible_values(Shell::possible_values())
                 .value_name("SHELL")
                 .help("Print shell completions to STDOUT and exit")
         )
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -5,7 +5,7 @@ use crate::ui::image_backends;
 use crate::ui::image_backends::ImageBackend;
 use crate::ui::printer::SerializationFormat;
 use anyhow::{Context, Result};
-use clap::{crate_description, crate_name, crate_version, AppSettings, Arg};
+use clap::{crate_description, crate_name, crate_version, AppSettings, Arg, ValueHint};
 use image::DynamicImage;
 use regex::Regex;
 use std::process::Command;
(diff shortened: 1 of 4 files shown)


spenserblack:
Note that the "real" diff is actually much smaller than the git diff. A simple change in indentation results in a massive number of lines changed :rofl:

src/cli.rs
```diff
@@ -39,16 +41,207 @@ pub struct Config {
...
+    }
+}
+
+pub fn print_completions<G: Generator>(gen: G) {
+    let mut cmd = build_cli();
+    let name = cmd.get_name().to_string();
+    generate(gen, &mut cmd, name, &mut io::stdout());
+}
+
+pub(crate) fn build_cli() -> clap::Command<'static> {
```
o2sh, line 236: why use pub(crate) and not just pub? The project consists of only one crate.

src/cli.rs
```diff
@@ -289,179 +485,15 @@ impl Config {
...
-    let version = Command::new("git").arg("--version").output();
-
-    match version {
-        Ok(v) => String::from_utf8_lossy(&v.stdout).replace('\n', ""),
-        Err(_) => String::new(),
-    }
+        .arg(
+            Arg::new("completion")
+                .long("completion")
+                .possible_values(possible_shells)
```
o2sh, line 495:
```suggestion
                .possible_values(Shell::possible_values())
```

More readable IMO

src/cli.rs
```diff
@@ -289,179 +485,15 @@ impl Config {
...
+                .value_name("SHELL")
+                .help("Print shell completions to STDOUT and exit")
```
o2sh, line 497:
```suggestion
                .help("Print shell completion to STDOUT and exit.")
```

Typo

o2sh (approved):
That's a neat feature. I wasn't aware of shell completion.

Left a few minor comments.

src/cli.rs
```diff
@@ -39,16 +41,207 @@ pub struct Config {
...
+
+    match version {
+        Ok(v) => String::from_utf8_lossy(&v.stdout).replace('\n', ""),
+        Err(_) => String::new(),
+    }
+}
+
+pub fn print_completions<G: Generator>(gen: G) {
+    let mut cmd = build_cli();
+    let name = cmd.get_name().to_string();
```
o2sh, line 232: can't we just do `let name = crate_name!();` ?

src/cli.rs
spenserblack: No particular reason, it was just a force of habit. You're right, it doesn't matter for us since we don't have a lib, only a bin.

src/cli.rs
spenserblack: Yeah, this was from when I was experimenting with the lifetimes of the Args.

o2sh (approved):

src/cli.rs
spenserblack: I was mostly just following `clap_complete`'s docs, but I don't see any harm in this for our use.

o2sh: You're right, I thought we could save the call to build_cli() but `cmd` is passed to the generate function.

Sorry for the confusion

spenserblack: No worries, the second call to `build_cli` in their docs looked really weird to me, too. Also that the clap command was passed as a mutable reference. I didn't look too much into it, though.

PR #669, opened 2022-05-18
worktree support via gitoxide 0.17 (up from 0.16)

### LINKED ISSUE #651: Upcoming v2.13 doesn't work at all with worktrees
This is due to `gitoxide` currently not handling them at all.

In the example below, `onefetch` is the latest build from `main` whereas `onefetch-pre-gitoxide` is the one using `git2`.
```
➜  gitoxide.worktree git:(gitoxide.worktree) onefetch
Error: Could not find a valid HEAD reference

Caused by:
    The ref partially named 'HEAD' could not be found
➜  gitoxide.worktree git:(gitoxide.worktree) onefetch-pre-gitoxide
                 R RR RR                  Sebastian Thiel ~ git version 2.32.0 (Apple Git-132)
              R RRRRRRRR R          R     ----------------------------------------------------
 R RR       R RRRRRRRRRRRRR R      RR     Project: gitoxide (455 tags)
rR RRR    R RRRRRRRRRRRRRRRRR R   RRR R   HEAD: df8fe68 (gitoxide.worktree, worktree-stack)
RRR RR   RRRRRRRRRRRRRRRRRRRRRRR  RRRRR   Version: git-sequencer-v0.0.0
 RRRRR  RRRRRRRRRRRRRRRRRRRRRRRR  RRRR    Created: 3 years ago
  RRR RRRRRRRRRRRRRRRRRRRRRRRRRRRR RR     Languages:
    R  RRRRRRRRRR=  RR = RRRRRRRRRRR                 ● Rust (96.9 %) ● Shell (3.1 %)
     RRRRRRRRRRRR=  RR = RRRRRRRRRR       Dependencies: 12 (cargo)
      RRRRRRRRRRR   RR   RRRRRRRRRR       Authors: 95% Sebastian Thiel 6398
     RR==RRRRRRRRRRRRRRRRRRRRRR===RR               2% Edward Shen 147
     RR =  ==RRRRRRR  RRRRRR==  = RR               0% Conor Davis 33
      RR =     ===========     = RR       Last change: 7 minutes ago
       RR                        R        Contributors: 30
        R                       R         Repo: /Users/byron/dev/github.com/Byron/gitoxide
         R                                Commits: 6715
                                          Lines of code: 75163
                                          Size: 6.37 MiB (1212 files)
                                          License: Apache-2.0, MIT



➜  gitoxide.worktree git:(gitoxide.worktree)
```

I'd consider this a blocker for the v2.13 release and will implement basic support for this next.

Byron:
Fixes #651 .

### DIFF
--- a/src/info/repo.rs
+++ b/src/info/repo.rs
@@ -155,7 +155,7 @@ impl Commits {
 
 impl Repo {
     pub fn new(git2_repo: Repository) -> Result<Self> {
-        let repo = git::open(git2_repo.workdir().unwrap_or(git2_repo.path()))?;
+        let repo = git::open(git2_repo.path())?;
 
         Ok(Self { repo, git2_repo })
     }
--- a/src/info/repo.rs
+++ b/src/info/repo.rs
@@ -155,7 +155,7 @@ impl Commits {
 
 impl Repo {
     pub fn new(git2_repo: Repository) -> Result<Self> {
-        let repo = git::open(git2_repo.path())?;
+        let repo = git::open(git2_repo.workdir().unwrap_or(git2_repo.path()))?;
 
         Ok(Self { repo, git2_repo })
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,9 +202,9 @@ checksum = "e3b5ca7a04898ad4bcd41c90c5285445ff5b791899bb1b0abdd2a2aa791211d7"
 
 [[package]]
 name = "byte-unit"
-version = "4.0.13"
+version = "4.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "956ffc5b0ec7d7a6949e3f21fd63ba5af4cffdc2ba1e0b7bf62b481458c4ae7f"
+checksum = "95ebf10dda65f19ff0f42ea15572a359ed60d7fc74fdc984d90310937be0014b"
 dependencies = [
  "utf8-width",
 ]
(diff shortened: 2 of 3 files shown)


src/info/repo.rs
```diff
@@ -155,7 +155,7 @@ impl Commits {
 
 impl Repo {
     pub fn new(git2_repo: Repository) -> Result<Self> {
-        let repo = git::open(git2_repo.path())?;
+        let repo = git::open(git2_repo.workdir().unwrap_or(git2_repo.path()))?;
```
spenserblack, line 158: As a quick double-check, looks like `.workdir()` only returns `None` if the repository is bare. So effectively this attempts to open either the workdir path or the path to the bare repository, right? Just checking that that's wanted.

spenserblack (approved):
Overall LGTM! Just a note for a quick double-check.

src/info/repo.rs
Byron: That's a great catch! Actually this was a workaround to `gitoxide` being unable to open `.git/worktrees/<name>` paths at the time, which has since been fixed. So I rolled it back to `git2_repo.path()` exactly like before.

As a side-note, while implementing worktree support it became clear that I have to change my thinking about what `bare` really means. In short: don't assume that `bare` repositories have no worktree. In fact, they can have several _linked_ worktrees via `git worktree`. The only way to know if there is a worktree is to open a repository and checking for the presence of a `work_dir()`. Oh, to make things more fun: this directory might not even exist as it's just something pointed to by a file that we didn't have to touch yet (and which is exactly what can happen if a repo is opened through `.git/worktrees/<name>`.

Issue #638, opened 2022-04-05
Add WinGet as an installation option


russellbanks:
It would be nice if this could be installed via [WinGet](https://github.com/microsoft/winget-cli)

o2sh:
Sorry for the late answer. I'll take care of it and report back once it's done.

russellbanks:
Hi, I've looked into WinGet a lot since I made this issue and the reason I closed this issue was WinGet currently does not support zip files, only installers:

> Note: At this time installers must either be MSIX, MSI or .exe application installers. Standalone or portable executables, compressed .zip files, and fonts are not currently supported.

Therefore, it looks like this project would have to compile a binary that is installable in order to package it on WinGet.

o2sh:
That's right, but we should be able to create a windows installer using [Inno Setup](https://jrsoftware.org/isinfo.php) and make it part of the CD pipeline.

spenserblack:
Coincidentally, I stumbled on [microsoft/setup-msbuild](https://github.com/microsoft/setup-msbuild) after seeing it used [in this workflow](https://github.com/cli/cli/blob/a0a9099037ab91e10fc389b4cd07c5fe7ed1bd5d/.github/workflows/releases.yml#L133). Don't know much about making Windows installers, but it might be useful.

o2sh:
I found a rust project that does exactly what we want: it calls a [shell script](https://github.com/wez/wezterm/blob/6f0888cd7dda052137d3f304a5d7a95dea51bd8c/ci/deploy.sh#L100) from ci and it uses `iscc.exe` (Inno Setup) with an [iss file](https://github.com/wez/wezterm/blob/main/ci/windows-installer.iss) to generate the installer.

o2sh:
Onefetch 2.12.0 can now be installed using winget 🎉 

I have updated the [wiki ](https://github.com/o2sh/onefetch/wiki/Installation) with a new entry for winget and #668 should take care of generating a new installer on release.

> If it's helpful, I use the [Winget Releaser](https://bittu.eu.org/docs/wr-intro) GitHub action in my repositories that runs on release and will automatically create a pull request on WinGet packages. 

Indeed, this would be the last task to have a fully automated winget deployment process..

> I'm happy to make a PR for this if it would be beneficial.

That would be great @russellbanks 🚀

russellbanks:
> Onefetch 2.12.0 can now be installed using winget 🎉

Just installed and working great, thank you!

[...]

Issue #442, opened 2021-06-02
Translate README.md in Russian 🇷🇺


o2sh:
It would be great to have a multiple translations of the [`README.md`](https://github.com/o2sh/onefetch/blob/master/README.md) in order to make it more accessible for non English-speaker to use the tool. These translations will be accessible directly from the English `README.md` via hyperlinks and placed in the `docs` folder.

If you're a native Russian-speaker :ru: and are willing to help, your contribution would be more than appreciated :100:

Vinegret43:
I'm a native Russian speaker and also know English quite well, i think i'll try to do this. I'm absolutely new to this project, is there anything i should know before contributing? Also if you need translations for other files (Documentation, for example), please let me know, i'm always glad to help

Vinegret43:
I have a question. Look at this paragraph:
```
It automatically detects open source licenses from texts and
provides the user with valuable information like code distribution,
pending changes, ...
```
What does "code distribution" mean? Does it mean percentage of every language in a repo? This can be translated in many different ways without context

o2sh:
> What does "code distribution" mean? Does it mean percentage of every language in a repo?

Correct, I meant to refer to this info line: 

![cd](https://user-images.githubusercontent.com/13710835/132137521-2288efeb-3084-46cd-862e-24525edbda83.png)

Looking forward to see your  translation added to the project :muscle:

PR #673, opened 2022-05-26
Create WinGet workflow

russellbanks:
This pull request adds a workflow that automatically publishes Onefetch to WinGet every release, as discussed in #638. *Please do not merge this pull request without reading the below and making the required changes* (The token part is the important bit) :)

- `runs-on: windows-latest`

The WinGet Releaser action can only run on Windows.

</br>

- `identifier: o2sh.onefetch`

This is the identifier for Onefetch on Winget. It will use this to find Onefetch's current package on WinGet.

</br>

- `installers-regex: '\.exe$'`

This will make all `.exe` files in the assets of the latest Onefetch release to be included in the pull request.

</br>

- `delete-previous-version: 'true'`

This is optional, and you are welcome to change this to false (or omit it; the default is false). This will essentially overwrite the previous version on WinGet packages, so that only this new version is stored on WinGet. Again, entirely optional.

</br>

- `token: ${{ secrets.WINGET_TOKEN }}`

This is the bit that will require action from yourself. This is a GitHub token with which the action will authenticate with GitHub and create a pull request on the [winget-pkgs](https://github.com/microsoft/winget-pkgs) repository.

To do this, you will need to:

1. Create a Personal Access Token (PAT) with `public_repo` scope on your GitHub account.
image

2. Create a repository secret containing the token. **Super important: call this token: `WINGET_TOKEN`**.
See [using encrypted secrets in a workflow](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) for more details.

</br>

**Finally, make sure you have a fork of [winget-pkgs](https://github.com/microsoft/winget-pkgs) on your GitHub account. Please do not delete the fork after a pull request has been merged as the action will use that fork for making a branch and merging it with the upstream [winget-pkgs](https://github.com/microsoft/winget-pkgs) repository on every Onefetch release.**

The above is made on the assumption that the account that the token is from (and the fork is present) is @o2sh. If you would like this to be done by another account, the token would have to be from the different user and an extra line would have to be added: `fork-user: # Username of other user` (this defaults to the repository owner, which is why it doesn't have to be added)

If you would like to read about this action further, the documentation is [here](https://bittu.eu.org/docs/wr-intro) and the source code is [here](https://github.com/vedantmgoyal2009/vedantmgoyal2009/blob/main/winget-pkgs-automation/releaser-action/action.yml).

If you are interested in what a pull request made by this action looks like, here is one from my account: [#61982 on winget-pkgs](https://github.com/microsoft/winget-pkgs/pull/61469).

~~If you are concerned about the generated executable (#668) not getting included in the action when it checks the assets, it actually has a retry logic that will retry for up to an hour, checking every ten minutes for the required asset in the created release. If it's still not in the assets at that point, it will fail and you can re-run it at a later date when it is.~~ Merged with CD workflow so this will always run after the assets have been built.


spenserblack (changes requested):
:thinking: It would probably be ideal if this could be combined with the `CD` workflow. Then you can [set that this job needs `deploy`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds), which seems necessary as it looks like this action relies on assets already being published. Probably better than just retrying every 10 minutes up to the 5th try. It's also ideal for this job to simply not run if the `deploy` job fails.

******

You can ignore my note on `uses: vedantmgoyal20009/...`. I'm not requesting a change there, just making a note.

spenserblack:
>The WinGet Releaser action can only run on Windows.

It could be made into a [docker action](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action) to allow it to run on ubuntu/macos, but then it wouldn't be able to run on windows :laughing: :upside_down_face: 
(though the docker image itself could run Windows AFAIK)

russellbanks:
> 🤔 It would probably be ideal if this could be combined with the `CD` workflow. Then you can [set that this job needs `deploy`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds), which seems necessary as it looks like this action relies on assets already being published. Probably better than just retrying every 10 minutes up to the 5th try. It's also ideal for this job to simply not run if the `deploy` job fails.

I actually made an issue about a similar thing I had: [#442](https://github.com/vedantmgoyal2009/vedantmgoyal2009/issues/442), asking for it to be allowed to be ran on other workflow events. However, they [said they ran into an issue with this](https://github.com/vedantmgoyal2009/vedantmgoyal2009/issues/442#issuecomment-1122096742):

> @russellbanks I was testing the action with the `workflow_run` event, but the problem is the action won't be able to get the release information automatically as it does now and my main concern while creating the action was only that a user would not have to supply everything as an input for the action.

spenserblack:
Huh, just took a quick look, but it seems like this could honestly be automated with [`winget-create`](https://github.com/microsoft/winget-create), which also supports submitting PRs.

spenserblack:
> I actually made an issue about a similar issue I had: [#442](https://github.com/vedantmgoyal2009/vedantmgoyal2009/issues/442), asking for it to be allowed to be ran on other workflow events.

With the `needs:` option a job can be delayed until another job is finished, and still be triggered with `on: release` by being part of the same overall workflow.

russellbanks:
> With the needs: option a job can be delayed until another job is finished, and still be triggered with on: release by being part of the same overall workflow.

I didn't consider this. Thank you :)

> Huh, just took a quick look, but it seems like this could honestly be automated with [winget-create](https://github.com/microsoft/winget-create), which also supports submitting PRs.

I imagine this reason that this hasn't be done already is because wingetcreate isn't as verbose in its options as the script is maybe? For example, being able to delete a previous version.

</br>

Meanwhile, I could tag always vedantmgoyal2009 for visibility on this? I imagine they'd be happy to help

spenserblack:
AFAIK there wouldn't be any problem changing the action trigger type to `released`, but we should get confirmation from @o2sh since they've been creating the releases so far, and see if they care about the timing. Although wouldn't all needed event attributes (assets, tag, publish date, etc.) be available with the `published` release type, too?

o2sh:
TBH, I've just learned - thanks to @vedantmgoyal2009 comment - the distinction between the `published`  and the `released` trigger type 😅.  I agree with @vedantmgoyal2009, the CD pipeline should only be triggered on release --> ac338c8

o2sh:
@russellbanks I created the token:
Screenshot 2022-05-29 at 22 18 48

russellbanks:
> @russellbanks I created the token

Brill, just make sure you've added that token to this repository and then this can be merged :)

o2sh:
done:

Screenshot 2022-05-29 at 22 23 49

I'll let @spenserblack finish his review and approve the PR, did you settle on [this](https://github.com/o2sh/onefetch/pull/673#pullrequestreview-986643860)?

spenserblack:
Thanks for the changes, @o2sh! With this, we should be able to add this as a job to the CD workflow, so that it only triggers after a release is created. Something vaguely like this
```yaml
jobs:
  deploy:
    # ...
  winget-pkg-release:
    needs: [deploy]
    runs-on: windows-latest
    # etc.
```

******

BTW, a `max-retries` input `winget-pkgs-automation/releaser-action` might be cool. In our specific case, as long as we have guaranteed that `deploy` has completed before this action, then it would probably fail as soon as possibly instead of retrying. This is because, if there are no assets available after `deploy` is finished, something has gone *really* wrong and another try wouldn't really help.

spenserblack:
Oh, also I just thought about this:
@vedantmgoyal2009 currently your action requires a special token instead of the built-in `GITHUB_TOKEN` because the action uses another repository to generate a PR. But you could actually clone `winget-pkgs` into the same repo and generate a PR from there, thus allowing you to use `GITHUB_TOKEN`. Here's an example of what that would look like:

```bash
git remote add winget https://github.com/microsoft/winget-pkgs.git
git fetch winget
git branch winget-pkg winget/master
git checkout winget-pkg
# make changes
git push -u origin winget-pkg
# make PR from <repo>:winget-pkg to microsoft/winget-pkgs:master
```

It's not blocking this PR, but it's a way to make your action more friendly to users 🙂

russellbanks:
I've made those changes

spenserblack (approved):
LGTM!

Issue #552, opened 2021-12-08
Support colors when using custom ascii image (`--ascii-input` option)


Pomroka:
Using `--ascii-colors` don't affect my image, only text.

![image](https://user-images.githubusercontent.com/3210260/145279222-3d0b2c06-b129-4490-9bd4-35fe4e887605.png)

spenserblack:
Looks like an empty `Vec` is currently getting passed as the colors when custom ASCII art is used.
https://github.com/o2sh/onefetch/blob/3a1368a971bcd3ddca55563ac5671cf958c52f36/src/ui/printer.rs#L41-L65

spenserblack:
In fact, the current behavior is that colors are effectively disabled when using custom ascii art.

I'll change the title to reflect that we need to make a change to support this.

o2sh:
NGL, it never crossed my mind to combine the two flags `--ascii-colors` and `--ascii-input` together 😅 

It shouldn't be to hard to make it work 👍

Pomroka:
As i didn't find a way to load custom ascii image directly i used `--ascii-input` and when i saw no color i try use only logical option i found. 
Load custom ascii logo from file would be my another `Feature Request`

spenserblack:
>Load custom ascii logo from file

Until that feature is implemented I think that `onefetch --ascii-input "$(cat my-ascii-art)"` would work.

o2sh:
@Pomroka with #680, it should work as expected.
If you agree, don't forget to close the issue.

Pomroka:
👍

PR #680, opened 2022-06-11
fix for --ascii-colors and --ascii-input

bitbrain-za:
use `self.info.ascii_colors` instead of empty array.

#552

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -18,18 +18,14 @@ pub fn get_ascii_colors(
         dominant_language
     };
 
-    let colors = language.get_colors(true_color);
+    let mut colors: Vec<DynColors> = ascii_colors.iter().map(|s| num_to_color(s)).collect();
 
-    let colors: Vec<DynColors> = colors
-        .iter()
-        .enumerate()
-        .map(|(index, default_color)| {
-            if let Some(color_num) = ascii_colors.get(index) {
-                return num_to_color(color_num);
-            }
-            *default_color
-        })
-        .collect();
+    let mut language_colors: Vec<DynColors> = language.get_colors(true_color);
+
+    if language_colors.len() > colors.len() {
+        let mut missing = language_colors.drain(colors.len()..).collect();
+        colors.append(&mut missing);
+    }
     colors
 }
 
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -18,15 +18,19 @@ pub fn get_ascii_colors(
         dominant_language
     };
 
-    let mut colors: Vec<DynColors> = ascii_colors.iter().map(|s| num_to_color(s)).collect();
-
     let mut language_colors: Vec<DynColors> = language.get_colors(true_color);
 
-    if language_colors.len() > colors.len() {
-        let mut missing = language_colors.drain(colors.len()..).collect();
-        colors.append(&mut missing);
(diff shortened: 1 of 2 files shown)


src/ui/printer.rs
```diff
@@ -38,7 +38,6 @@ impl<W: Write> Printer<W> {
                 let center_pad = " ".repeat(CENTER_PAD_LENGTH);
                 let info_str = format!("{}", &self.info);
                 let mut info_lines = info_str.lines();
-                let colors: Vec<DynColors> = Vec::new();
```
o2sh, line 41: You can get rid of `use owo_colors::DynColors;` which is now an unused import.

src/ui/printer.rs
```diff
@@ -62,7 +61,11 @@ impl<W: Write> Printer<W> {
                     );
                 } else {
                     let mut logo_lines = if let Some(custom_ascii) = &self.info.config.ascii_input {
-                        AsciiArt::new(custom_ascii, &colors, !self.info.config.no_bold)
+                        AsciiArt::new(
```
o2sh, line 64: This may be a little harder than that.
Indeed, `self.info.ascii_colors`'s capacity is based on the dominant language definition (look [here](https://github.com/o2sh/onefetch/blob/86f730791348357e9306d8b028a64341f1ed618e/src/ui/mod.rs#L9)). As a result, depending on the dominant language, output may differ:

Example 1 (shell has only 1 color):

Screenshot 2022-06-12 at 01 23 25

Example 2 (TypeScript has 2 colors):

Screenshot 2022-06-12 at 01 24 55

Example 3 (c# has more than 3 colors):

Screenshot 2022-06-12 at 01 25 52

o2sh (changes requested):

src/ui/printer.rs
bitbrain-za: Thank you for your guidance. I'll put a WIP prefix until I've made the requested changes.

bitbrain-za: image

This handles your example cases.
Still rather new to rust so any criticism on the approach taken would be much appreciated.

src/ui/mod.rs
```diff
@@ -18,18 +18,14 @@ pub fn get_ascii_colors(
         dominant_language
     };
 
-    let colors = language.get_colors(true_color);
+    let mut colors: Vec<DynColors> = ascii_colors.iter().map(|s| num_to_color(s)).collect();
```
o2sh, line 21:
```rust
let mut language_colors: Vec<DynColors> = language.get_colors(true_color);

if ascii_colors.is_empty() {
    language_colors
} else {
    let mut colors: Vec<DynColors> = ascii_colors.iter().map(|s| num_to_color(s)).collect();

    if language_colors.len() > colors.len() {
        let mut missing = language_colors.drain(colors.len()..).collect();
        colors.append(&mut missing);
    }
    colors
}
```
minor: for better performance --> exit early when `ascii_collors` is empty.

o2sh (approved):

src/ui/mod.rs
bitbrain-za: Thanks for the suggestion, I've added it.

o2sh:
Thanks a lot for you contribution @bitbrain-za 

It works great 🎉 


Screenshot 2022-06-12 at 18 58 45

PR #686, opened 2022-06-18
Update cd.yml

vedantmgoyal2009:


spenserblack (approved):
LGTM!

Is your release plan to force-push the `latest` tag on additional releases?

One thing we might want to watch out for is breaking changes: if we always use the `latest` tag, then we'll silently "update" to a new release that could break the workflow. @o2sh what do you think?

vedantmgoyal2009:
> Is your release plan to force-push the latest tag on additional releases?

Yes, since the core logic is highly unlikely to change, I've decided to do so.

PR #693, opened 2022-06-28
adding french documentation support

Kaderovski:
Hello, Bonjour

As a French native I propose you a French version of your documentation. :wine_glass: :baguette_bread: 

Thx for your great work

PR #694, opened 2022-06-28
Ignore docs for CI runs

spenserblack:
This will only run the CI if a contributor has edited
- the `src/` directory
- the `resources/` directory
- the manifest files

This prevents unnecessary workflow runs. For example, on documentation
changes.

For example, #693 didn't need to be in a "workflow run pending" status since none of the changes are tested by the workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,16 @@ name: CI
 on:
   push:
     branches: ["*"]
+    paths:
+      - 'src/**'
+      - 'resources/**'
+      - 'Cargo.*'
   pull_request:
     branches: [main]
+    paths:
+      - 'src/**'
+      - 'resources/**'
+      - 'Cargo.*'
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,16 +2,14 @@ name: CI
 on:
   push:
     branches: ["*"]
-    paths:
-      - 'src/**'
-      - 'resources/**'
-      - 'Cargo.*'
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
   pull_request:
     branches: [main]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,8 +2,16 @@ name: CI
 on:
   push:
     branches: ["*"]
+    paths:
+      - 'src/**'
+      - 'resources/**'
+      - 'Cargo.*'
```
o2sh, line 8:
```suggestion
    paths-ignore:
      - "docs/**"
      - "**.md"
```

It may be safer that way.

.github/workflows/ci.yml
```diff
@@ -2,8 +2,16 @@ name: CI
...
   pull_request:
     branches: [main]
+    paths:
+      - 'src/**'
+      - 'resources/**'
+      - 'Cargo.*'
```
o2sh, line 14:
```suggestion
    paths-ignore:
      - "docs/**"
      - "**.md"
```

Same remark

o2sh (approved):
Great idea @spenserblack  🚀 

I left a minor comment.

.github/workflows/ci.yml
spenserblack: Yeah, I was debating between the two, but you're probably right. But there are a lot of other paths to ignore: assets, config files, etc. But this should prevent the majority of unnecessary runs, so applying.

Viyom29:
> This will only run the CI if a contributor has edited
> 
> - the `src/` directory
> 
> - the `resources/` directory
> 
> - the manifest files
> 
> 
> 
> This prevents unnecessary workflow runs. For example, on documentation
> 
> changes.
> 
> 
> 
> For example, #693 didn't need to be in a "workflow run pending" status since none of the changes are tested by the workflow.

PR #695, opened 2022-06-28
contributors templates documentation

Kaderovski:
@spenserblack, @o2sh can this help ?


CONTRIBUTING.md
```diff
@@ -82,6 +82,8 @@ In order to make Onefetch more accessible for non English-speakers, we are seeki
 
 These translations will be accessible directly from the English README.md via hyperlinks.
 
+You can use `partials/templates/Readme.contribute.md` file as starter template and replace elements between `{{ }}` with your content.
```
spenserblack, line 85:
```suggestion
You can use `partials/templates/Readme.[language-code].md` file as starter template for your translation and replace elements between `{{ }}` with your content.
```
And also rename the partial, of course :slightly_smiling_face:

spenserblack (changes requested):
### From #693 
@spenserblack:
>@o2sh Perhaps we should look into using markdown partials/templates so that documentation contributors, especially translators, don't have to repeat anything, like badge links.

@Kaderovski:
>Like : https://github.com/o2sh/onefetch/pull/695 ?

******

@Kaderovski kind of, yeah. Ideally the translation documents (`readme.fr.md`, etc.) would be built by a workflow. The lowlighter/metrics repository has examples of using [partials](https://github.com/lowlighter/metrics/tree/master/.github/readme) and [building them with a script](https://github.com/lowlighter/metrics/blob/master/.github/scripts/build.mjs). We don't need something nearly as complex as that solution, but if we could get to a point somehow where a translation is parsed into a README template (or badges are parsed into a translation?), possibly in an automated way, that would be pretty cool.

But your PR now is OK for me, with one small change.

Kaderovski:
I've also seen this method on other repos (if it might help?)

- [you]Create a onefetch/template repo (where you put basically the skeleton markdown template)
- [contributor] will create a repo on their own & leave it empty
- [contributor] add onefetch template as a remote: git remote add template git@github.com:onefetch/template.git
- [contributor] pull from it: git pull template main --depth=1
- [contributor] delete the remote: git remote remove template
- [contributor] then they can work directly with

spenserblack (approved):

spenserblack:
>I've also seen this method on other repos (if it might help?)
>...

That's a pretty interesting solution! I'll have to think about the best way to accomplish this, but I'd like to make it as easy for contributors as possible. Thinking purely from a contributor workflow, something like this:
1. fork onefetch
2. create a file like `fr.translation.<extension>` (maybe YAML?)
3. make PR

And on our end something like this:
on push to `main` with path `docs/translations`: parse values from translation file to template

Besides ease of contributing, my secondary goal (OK, maybe actually primary :sweat_smile:) is to reduce duplicated work as much as possible (e.g. changing image links). This does add some complication to the repo though, possibly enough to warrant a separate docs repo, so it may not be worth it.

o2sh:

Issue #696, opened 2022-07-03
Automated logo rendering/visualization


spenserblack:
### Summary 💡

Started the conversation [here](https://github.com/o2sh/onefetch/issues/328#issuecomment-1172794745), but my comment was unrelated enough that it should have been a new issue.

Basic idea is that we could have a solution to automate previews of the logos. This can probably be done with a serverless function.

Some things to consider:
- What tool to use for the serverless function. I've used Vercel, but there are plenty of options out there.
- We'll need the languages' color definitions to be available to the serverless function. We can do this in a few ways:
  - Convert the language definitions to a data format (e.g. JSON) so that a serverless function can read it. If we do this we can
    - make a proc macro to parse the JSON in the Rust code
    - use a build binary to generate the Rust code
    - either way we probably wouldn't want to do something that results in runtime JSON parsing
  - Add another binary to the onefetch crate, that will be able to import the language definitions and act as the serverless function
  - Export the language definitions with a minimal library and use a different crate for the serverless function. If we do this, a couple more things to consider (this can be influenced by the serverless utility and how it reacts to repository changes):
    - Should the serverless function be part of a workspace, in the same repo?
    - ~~Should it exist in a different repo?~~ (we probably want it in the same repo to listen for changes and be able to import the latest change from the main onefetch project)
- Does the serverless function render the HTML? It may be more efficient to generate HTML once and serve a static site, but we can dynamically generate HTML per request.
- Do we simply serve HTML, or should we have an API for the language stats? If we have an API, we can use a reactive library to read JSON and generate a preview. The API could be a simple `api/language[/:name]`.

Also if we just generate the HTML once, we can actually just use GH Pages instead of a serverless tool. But one thing I've liked about Vercel is that it creates preview builds for PRs (see motivation).

### Motivation 🔦

When reviewing PRs, currently we have to pull down the PR's changes and rebuild the binary to review any added or changed logos. This can be greatly simplified by making previews that are automatically generated, which can allow maintainers to review the logo even if they're not at their main device.

o2sh:
I am very excited for this new feature 😃
I've been eyeing the `onefetch.run` domain name for a while now 😅 

> What tool to use for the serverless function. I've used Vercel, but there are plenty of options out there.

Vercel seems like a good choice (nice UI, good integration w/ GH). 

> We'll need the languages' color definitions to be available to the serverless function.

[...]

PR #699, opened 2022-07-05
Extract language definitions into data file

spenserblack:
This uses a templating tool, [Tera], to read the language data and generate the appropriate Rust code,
instead of using a macro. Some of the benefits of this:
- A familiar format that makes it easy for users to contribute (github-linguist also uses YAML).
- The templating language allows more enforcement at compile time instead of runtime. For example,
  matching basic and true color lengths are now enforced with a compiler macro instead of a test runtime.
- Other tools will be able to read this data without onefetch needing to export it with a public API.

There are some unnecessary commits in this PR: making a temporary onefetch lib implementation,
making a temp binary to export onefetch data to JSON, etc. That doesn't need to stay in this PR
and can be rebased away, but I thought it was useful to preserve the history in some form.

For #696

## To do

- [x] Finish redefining the languages' colors
- [x] Update `CONTRIBUTING.md`

[Tera]: https://github.com/Keats/tera


spenserblack:
There will likely be more changes needed, but I think this is at least ready for an initial review.

spenserblack:
These types of things always make we wonder if the data files should be discoverable by linguist, since in this case it's used in code generation.

spenserblack:
I think we're going to need to add some integration tests at some point. This PR shouldn't change any behavior at all, but right we basically ensure that by checking if it the output looks good. I automated the bulk of this with [a quick, lazy, buggy script](https://github.com/o2sh/onefetch/pull/699/commits/5d7bfc027662553bf4be39659a55960be97ed4d9), so I don't have 100% confidence that there weren't some unintended changes to logos. It definitely mangled the Haskell logo, which I had to fix.

languages.yaml
```diff
@@ -0,0 +1,2572 @@
...
+  colors:
+    ansi:
+      - default
+      - cyan
+      - blue
+    rgb:
+      - [255, 255, 255]
+      - [0, 24, 201]
+      - [12, 10, 124]
+    chip: [2, 248, 140]
```
o2sh, line 34:
```suggestion
    chip: Rgb[2, 248, 140]
```
We should maybe keep the `Rgb` prefix everywhere to make it more self explanatory and obvious we're talking about colors.

languages.yaml
```diff
@@ -0,0 +1,2572 @@
...
+    {0}             *************    ******
+    {0}  /| |                           ****
+    {0} /-|(|(|                     ********
+    {0}                        ************
+    {0}Time-tested, safe     *************
+    {0}and secure            ***********
+    {0}                         *****
+  colors:
+    ansi:
+      - default
```
o2sh, line 27: It would be best if we good find a way to use `white` and make it map to `default` in the template.

templates/language.rs
```diff
@@ -0,0 +1,129 @@
...
+    {% endif %}
+
+    {% for line in lines %}
+        {% set cleaned_line = line | strip_color_indices %}
+        {% set width = cleaned_line | length %}
+        {% if width > max_width %}
+            compile_error!("{{ language }}: ascii art line {{ loop.index }} must be {{ max_width }} or less characters wide");
+        {% endif %}
+    {% endfor %}
+{% endfor %}
```
o2sh, line 129: I'm not sure why but if you look at the generated `language.rs` by tera in the /target/debut/onefetch-*/out folder, this part of the template results in thousands of empty lines.

Besisdes, IMO these should be unit tests not compilation errors. We should maybe find a way to put them back in the original `language.rs` file.

languages.yaml
spenserblack: Is that valid YAML? AFAIK that would deserialize into a string, which would require some additional parsing in addition to deserialization in `build.rs`.
It could perhaps be an object, like
```yaml
chip:
  r: 2
  g: 248
  b: 140
```
but I was trying to reduce the length of the config while maintaining human readability. Or maybe change the key to `chip-rgb`.

o2sh:
Very impressive @spenserblack :open_mouth: 
You must have broken the record of the biggest PR (number of files changed) on this repo 😄 🎉 

I left a few comments, but I'll finish the review over the week end 👍

templates/language.rs
spenserblack: Yeah, the empty lines are from the `if` statements never being entered. E.g.
```
{# 3 empty lines #}
{% for n in [1,2,3] %}
  {% if n > 4 %}println!("Hello!"){% endif %}
{% endfor %}
```
IMO we don't need to worry too much about how pretty the generated code is: generated code isn't really intended to be human-readable.

For compile errors vs tests, the benefit of compile errors is that you can't even run onefetch if the values are invalid. Additionally, IMO it's preferable to assert the validity of values at compile-time instead of a runtime (in this case the test runtime). IMO tests should really be for behavior, not validating internal values. As a reductionist example, it's kind of like doing this:
```rust
/// A number between 0 and 255
const SMALL_NUMBER: usize = make_number!();

#[test]
fn number_is_small() {
    assert!(SMALL_NUMBER < 256);
}
```

instead of this:
```rust
/// A number between 0 and 255
const SMALL_NUMBER: u8 = make_number!();
```
Making them compile errors also should result in a faster CI, since the presence of a `compile_error!` macro in the generated code will make compilation fail pretty quickly.

languages.yaml
o2sh: Great idea! we're getting closer to the GitHub linguist format: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

`hex`  sounds  good to me.

templates/language.rs
o2sh: Agreed 👍 

> Yeah, the empty lines are from the if statements never being entered. E.g.

Maybe add some `-%`  to reduce the clutter.

src/info/langs/language.tera.rs
```diff
@@ -0,0 +1,129 @@
...
+    Prose,
+    Data,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash, EnumIter, clap::ValueEnum)]
+#[allow(clippy::upper_case_acronyms)]
+#[clap(rename_all = "lowercase")]
+pub enum Language {
+    {% for language, attrs in languages -%}
+        {% if attrs.serialization %}#[clap(name="{{ attrs.serialization }}")]{% endif %}
```
o2sh, line 23:
```suggestion
        {% if attrs.serialization %}#[clap(name="{{ attrs.serialization }}")]{% endif -%}
```

languages.yaml
spenserblack: @o2sh done 👍 https://github.com/o2sh/onefetch/pull/699/commits/9f02ad73bb59d84c9254b253e4406d51e95cb8b6

o2sh:
You can maybe simplify by doing

{{ color | capitalize | replace(from="Default", to="White" }}

src/info/langs/language.tera.rs
```diff
@@ -0,0 +1,137 @@
...
+                Language::{{ language }} => "{{ attrs.ascii | addslashes }}",
+            {% endfor %}
+        }
+    }
+
+    pub fn get_colors(&self, true_color: bool) -> Vec<DynColors> {
+        let colors = match self {
+            {% for language, attrs in languages -%}
+                Language::{{ language }} => Colors {
+                    basic_colors: vec![{%- for color in attrs.colors.ansi -%}Ansi(AnsiColors::{{ color | cleanansi }}),{% endfor %}],
```
o2sh, line 72: Could be simplified
```suggestion
                    basic_colors: vec![{%- for color in attrs.colors.ansi -%}Ansi(AnsiColors::{{ color | capitalize | replace(from="Default", to="White")}}),{% endfor %}],
```

o2sh:
I pushed a few changes (I hope you don't mind).

@spenserblack Do you think we can merge the PR as is? Or do we wait for the unit/integration tests?

CONTRIBUTING.md
```diff
@@ -3,7 +3,6 @@
 - [How to Contribute](#how-to-contribute)
   - [Contributing via Pull Requests](#contributing-via-pull-requests)
     - [Adding support for a new language](#adding-support-for-a-new-language)
-      - [Ascii logo](#ascii-logo)
```
spenserblack, line 6: TBH, now that GitHub has its own table of contents for Markdown files, we can probably remove this table of contents altogether 🙂

o2sh: You're right, I didn't even know about this (https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/) 😅

spenserblack:
>I pushed a few changes (I hope you don't mind).

I don't mind at all, please and thanks!

>Do you think we can merge the PR as is?

I'm definitely not opposed to adding some tests, but I looped through all of the languages and they all looked good. Looks like you already fixed the logos in commits like https://github.com/o2sh/onefetch/pull/699/commits/3e759a8606fe48c57aeedbac2af2c6c38e4544e5, thanks!

******

Given that the only thing that's "public" is the binary itself and its output to the terminal, I was thinking we could start with testing several variations of [`Printer`](https://github.com/o2sh/onefetch/blob/4130e7065c9706a4afa5b751e9c3df262c919973/src/ui/printer.rs#L21). For this PR in particular, the biggest concern was breaking the logo output.

TBH I was thinking of putting off these types of tests until Hacktoberfest, because it would be pretty repetitive 😅 . IMO we'd want 2 tests for each ASCII logo, both basic and true color, to guarantee stability. At this point, I think the output is stable enough that we'd want tests to guarantee stability during refactors like this.

I'm focusing mainly on logos just because of this PR, but we'd probably want to test the stats output, too. And with that comes variations of the git repository being analyzed, so that would come with testing different repositories (either real repositories or by mocking gitoxide). We might also want to test the CLI itself. While working on https://github.com/o2sh/onefetch/pull/685 I was surprised by a few panics. Tests could ensure that our CLI that we put a lot of work in doesn't break with clap updates, or for any other reason.

So that was a lot of words to say that, because this is a binary with no library, if our tests guarantee stability of input via the CLI and output to the terminal, I think we're good.

o2sh:
Glad to see you've already put some thoughts into it. IMO, we won't be able to skip on tests for much longer as more people seem to be playing with this tool.

Maybe we can merge this PR as is to be able to work on #696 and create a separate issue on that matter an mark it as a milestone for the next release.

Issue #700, opened 2022-07-09
Improve test coverage in `onefetch`


o2sh:
We're lacking integration and unit tests to make sure we don't run into regressions after new development/refactoring.

Modules that should be covered by unit tests:

- [ ] cli.rs
- [ ] license.rs
- [ ] info_field.rs
- [ ] head_refs.rs
- [ ] author.rs
- [ ] package_manager.rs
- [ ] text_colors.rs

Modules that should be covered by integration tests:

- [ ] printer.rs

spenserblack:
BTW I've used codecov.io on a few projects. I like its visualizations, and its GitHub integration can add checks to commits and PRs for coverage deltas.

spenserblack:
Great! Would you want to add a coverage badge? The markdown can be found in the [settings page](https://app.codecov.io/gh/o2sh/onefetch/settings) (I don't have access to the settings).

o2sh:
[![codecov](https://codecov.io/gh/o2sh/onefetch/branch/main/graph/badge.svg?token=P3oHskWePa)](https://codecov.io/gh/o2sh/onefetch)

o2sh:
I mean it's very low...maybe we should wait before putting it on display 😅 

[![codecov](https://codecov.io/gh/o2sh/onefetch/branch/main/graph/badge.svg?token=P3oHskWePa)](https://codecov.io/gh/o2sh/onefetch)

alessandroasm:
Hello @o2sh and @spenserblack! I was wondering on how to test the git integration in the mentioned files. I see two possible solutions: using mockall crate or testing this operations on simple temporary git repos. The latter approach is used by gitui project; here is the function they use to initilize the repo: https://github.com/extrawurst/gitui/blob/986d34a5acd520fbec91386675bec8013affc6bd/asyncgit/src/sync/mod.rs#L212-L255

What do you think about this?
Thank you

spenserblack:
To be honest I like both approaches.

For unit tests, I think mocking is a great idea, but for integration tests I like creating temporary git repos.

I like mocking especially when interacting with a library that can have many reasons to return an error. It allows the test to be "if we receive x response, we should do y" instead of "in this specific state, we should receive x response, and then do y."

In other words, I like making small unit tests for "do we handle this git library's responses correctly" and larger integration tests for "do we handle this git repository correctly".

atluft:
Hello @alessandroasm,
[...]

PR #705, opened 2022-07-22
gitoxide update

Byron:
### Tasks

* [x] update
* [x] use `repo.config()` instead of `git2` config in all places

### DIFF
--- a/src/info/repo.rs
+++ b/src/info/repo.rs
@@ -25,7 +25,6 @@ pub struct Commits {
 }
 
 pub struct Repo {
-    git2_repo: Repository,
     repo: git::Repository,
 }
 
@@ -155,10 +154,8 @@ impl Commits {
 }
 
 impl Repo {
-    pub fn new(git2_repo: Repository) -> Result<Self> {
-        let repo = git::open(git2_repo.path())?;
-
-        Ok(Self { repo, git2_repo })
+    pub fn new(repo: git::Repository) -> Result<Self> {
+        Ok(Self { repo })
     }
 
     pub fn gitoxide(&self) -> git::Repository {
--- a/src/info/repo.rs
+++ b/src/info/repo.rs
@@ -191,14 +191,11 @@ impl Repo {
         Ok(number_of_branches)
     }
 
-    pub fn get_git_username(&self) -> Result<String> {
-        let config = self.git2_repo.config()?;
-        let username = match config.get_entry("user.name") {
-            Ok(v) => v.value().unwrap_or("").into(),
-            Err(_) => "".into(),
-        };
-
-        Ok(username)
+    pub fn get_git_username(&self) -> String {
+        self.repo
+            .committer()
+            .map(|c| c.name.to_string())
+            .unwrap_or_default()
     }
 
     pub fn get_version(&self) -> Result<String> {
--- a/src/info/repo.rs
+++ b/src/info/repo.rs
@@ -241,23 +241,16 @@ impl Repo {
             None => return Ok(Default::default()),
         };
 
-        let mut repository_name = String::new();
-        let name_parts: Vec<&str> = remote_url.split('/').collect();
-        if !name_parts.is_empty() {
-            let mut i = 1;
-            while repository_name.is_empty() && i <= name_parts.len() {
-                repository_name = name_parts[name_parts.len() - i].to_string();
-                i += 1;
-            }
-        }
+        let url = git::url::parse(remote_url.as_bytes())?;
+        let path = git::path::from_bstr(url.path.as_bstr());
+        let repo_name = path
+            .with_extension("")
+            .file_name()
+            .expect("non-empty path")
+            .to_string_lossy()
+            .into_owned();
 
-        if repository_name.contains(".git") {
-            let repo_name = repository_name.clone();
-            let parts: Vec<&str> = repo_name.split(".git").collect();
-            repository_name = parts[0].to_string();
-        }
-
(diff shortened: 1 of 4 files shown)


Byron:
I think it's ready for review. The changes were tested on MacOS and Windows 11 just to be sure the repo name can indeed be figured out with the refactored code.

Now the last bastion of `gitoxide` to conquer is worktree changes, and I am looking forward to that. As mentioned before, it might not happen this year, but it's not for a lack of motivation. Let's see :).

src/info/repo.rs
```diff
@@ -191,14 +188,11 @@ impl Repo {
...
-            Ok(v) => v.value().unwrap_or("").into(),
-            Err(_) => "".into(),
-        };
-
-        Ok(username)
+    pub fn get_git_username(&self) -> String {
+        self.repo
+            .committer()
+            .map(|c| c.name.to_string())
+            .unwrap_or_default()
```
spenserblack, line 195: What do you think about defaulting to something like `"[UNKNOWN]"` if we cannot unwrap the committer? Technically a `user.name` *could* be `[UNKNOWN]`, but I think it's unlikely enough 😆 

Or if we stick with the default, why not use [`committer_or_default`](https://docs.rs/git-repository/0.20.0/git_repository/struct.Repository.html#method.committer_or_default)?

src/info/repo.rs
```diff
@@ -224,53 +218,39 @@ impl Repo {
     }
 
     pub fn get_name_and_url(&self) -> Result<(String, String)> {
```
spenserblack, line 220: Should the `return Ok(Default::default())` be changed to return `Err`s?

spenserblack (approved):
This is awesome work, and a huge step towards making onefetch more oxidized (I finally realized that the name is a reference to a "Rusty" version of git, right?) :crab: :tada:

I don't want this to block merging, but this PR might be a good opportunity to start adding some tests. What do you think the best testing strategy would be? Generating temporary repos? Mocking gitoxide functions/methods? Or something else?

src/info/repo.rs
Byron: A great catch, for this round I left it functionally similar, so if nothing is configured there are empty strings. This behaviour is also present in `repo_name_and_url()` even though `null` would probably bode better in with json output, I don't know if the UI has any special handling for these empty strings though. Generally I'd certainly prefer options.

This also answers why `committer_or_default()` wasn't chosen - you actually get an entirely made up user name `"gitoxide <gitoxide@localhost>"` which probably is misleading - it's akin to what git does if nothing is configured, using the current user name along with some way of figuring out a local email address.

To sum it up, I'd prefer returning `Option` instead of empty strings and have the UI decide how to display these.

src/info/repo.rs
Byron: I think this would break onefetch as it aborts the information retrieval process for repository information entirely. If there was an improvement to make, I'd certainly go for `Result<Option<(…)>>` here.

src/info/repo.rs
spenserblack: >for this round I left it functionally similar

Got it, that's probably for the best :+1:
Sorry, I should have checked the original behavior more closely

src/info/repo.rs
spenserblack: Got it, I'm OK with the existing code, then :smiley: :+1:

o2sh:
>Now the last bastion of gitoxide to conquer is worktree changes, and I am looking forward to that. As mentioned before, it might not happen this year, but it's not for a lack of motivation. Let's see :).

Looking forward to it!

>I don't want this to block merging, but this PR might be a good opportunity to start adding some tests. What do you think the best testing strategy would be? Generating temporary repos? Mocking gitoxide functions/methods? Or something else?

I'll let you address this part of @spenserblack's comment. I don't mind merging the PR as is but this is typically the kind of work that needs non regression tests. You can just give us some pointers and we will work on it as part of #700

Byron:
Oh, sorry, I completely and accidentally skipped over @spenserblack 's suggestion about adding some tests. Also I am happy to pick up on @o2sh to make suggestions instead of adding tests.

`gitoxide` mostly uses real git repositories for testing, and has amassed a set of utilities that are available via [git-testtools](https://docs.rs/git-testtools/latest/git_testtools/). The most important one here I imagine would be [`scripted_fixture_repo_read_only(…)`](https://docs.rs/git-testtools/latest/git_testtools/fn.scripted_fixture_repo_read_only.html) which executes some script while keeping the result locally available and dealing with re-executing it if its contents changes. That way, one could write shell scripts that clone repositories or set them up which can then be fed into `Repo::new(fixture)` to assert on the result accordingly. The system also comes with an archive feature that in conjunction with `git-lfs` helps to speed up CI as it won't have to clone repositories all the time anymore (instead, it downloads from `git-lfs`). That however, is optional and it's possible to disable archive creation entirely using `.gitignore` files. Now that I write it here I notice that the archival functionality is definitely still under-documented. I will correct that shortly (maybe by the time you read it it's already done).

Byron:
> This is awesome work, and a huge step towards making onefetch more oxidized (I finally realized that the name is a reference to a "Rusty" version of git, right?) 🦀 🎉

Yes, that's right :D. It's also the second attempt at naming the project, which was called `grit` before, it's `git`, but with some **r**ust inserted. The motivation to change the name was the programs name, `grit`, which took too long to type and `gix` is a little better even though I think most won't ever use it. Ideally, one day people prefer `ein` over `git` for typical workflows, but I digress :).

PR #723, opened 2022-08-01
Add stylechecks to Vercel CI

spenserblack:


docs/vercel/package.json
```diff
@@ -7,12 +7,17 @@
     "dev": "vite",
     "build": "vite build",
     "preview": "vite preview",
-    "check": "svelte-check --tsconfig ./tsconfig.json"
+    "check": "check:svelte && check:prettier && check:eslint",
+    "check:svelte": "svelte-check --tsconfig ./tsconfig.json",
+    "check:prettier": "prettier --check 'src/**/*.{ts,js,svelte}'",
```
o2sh, line 12:
```suggestion
    "check:prettier": "prettier --check **/*.{ts,js,svelte,css,html,json}",
```
assuming it ignores the `node_modules`.

.github/workflows/vercel-ci.yml
```diff
@@ -32,4 +32,10 @@ jobs:
...
       # redundant to test it here, too
       - name: Svelte Check
-        run: npm run check -- --fail-on-warnings --fail-on-hints
+        run: npm run check:svelte -- --fail-on-warnings --fail-on-hints
+
+      - name: ESLint
+        run: npm run check:lint
+
+      - name: Prettier
+        run: npm run check:prettier
```
o2sh, line 41: Will the CI fail on warnings?

.github/workflows/vercel-ci.yml
```diff
@@ -32,4 +32,10 @@ jobs:
       # NOTE Since build is automated by vercel, it's probably
       # redundant to test it here, too
       - name: Svelte Check
-        run: npm run check -- --fail-on-warnings --fail-on-hints
+        run: npm run check:svelte -- --fail-on-warnings --fail-on-hints
+
+      - name: ESLint
+        run: npm run check:lint
```
o2sh, line 38: Will the CI fail on warnings?

o2sh (approved):

.github/workflows/vercel-ci.yml
spenserblack: `check:prettier` has exit code 1 on warnings. Warnings have exit code 0 with ESLint, but we can use `--max-warnings=0` if we want to be strict about warnings.

docs/vercel/package.json
spenserblack: Yeah, [`node_modules` is ignored by default](https://prettier.io/docs/en/cli.html#--with-node-modules).

Good call, I've always ESLint way more than Prettier so I forgot that it also works for the non-programming languages, too :laughing: :facepalm:

.github/workflows/vercel-ci.yml
o2sh: Got it, ESLint does return an exit code of 1 on errors, but not on warnings:

Screenshot 2022-08-03 at 01 18 12

PR #752, opened 2022-08-23
update gitoxide to v0.21.1

Byron:
The most notable improvement is its ability to open submodule
repositories, thus increasing the amount of repositories onefetch
can handle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,16 +140,7 @@ dependencies = [
  "block-padding",
  "byte-tools",
  "byteorder",
- "generic-array 0.12.4",
-]
-
-[[package]]
-name = "block-buffer"
-version = "0.10.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0bf7fe51849ea569fd452f37822f606a5cabb684dc918707a0193fd4664ff324"
-dependencies = [
- "generic-array 0.14.6",
+ "generic-array",
 ]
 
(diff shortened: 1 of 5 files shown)


Byron:
It appears that CI is testing more environments now and the failure is legitimate. It tries to compile SHA-1 assembly on MSVC which is known to fail. However, [gitoxide tries to not turn it on](https://github.com/Byron/gitoxide/blob/517677147f1c17304c62cf97a1dd09f232ebf5db/git-features/Cargo.toml#L138-L139).

That seems to not work though and it's not the first time. One could experiment with other ways of turning it off on MSVC potentially from the manifest of `onefetch`, but ideally that's entirely handled by `gitoxide`. Lastly, if nothing else works, ASM for SHA-1 can be disabled entirely as I don't think hashing performance is important for `onefetch`'s workload.

Thanks for your suggestions.

spenserblack (approved):

o2sh:
Thanks a lot @Byron 
Sorry for that painful update 😅

Byron:
> Thanks a lot @Byron
> Sorry for that painful update 😅

I am glad these issues trigger here so we can sort them out (like I thought I did a while ago but something clearly isn't working). @davidkna posted a PR today which might also affect how sha1-asm builds are handled and also [created a fix upstream](https://github.com/RustCrypto/hashes/pull/388). To be continued….

PR #756, opened 2022-08-30
Fix typo in CHANGELOG.md

Piturnah:


spenserblack (commented):
Thank you! It looks like a lot of other instances of `where` in the changelog should actually be `were`. It also looks like some should be `was` instead.

Would you like to add that to this PR?

Piturnah:
I've fixed the other cases I could find. The blame says it was generated with `cargo changelog --write`, so it appears this must be an issue with that command.

spenserblack (approved):

spenserblack:
Thank you! I've opened up https://github.com/Byron/gitoxide/issues/513, so hopefully we won't have this issue in the future.

PR #790, opened 2022-10-03
WIP: Config file for theming/customization

### LINKED ISSUE #745: Config file for theming/customization
### Summary 💡

Have a configuration in which you could specify visual aspects of onefetch, for example:

General options:

- The glyph to use for the separator (use "->" instead of ":")
- The padding between the output and its surroundings
- The spacing between the separator and the info
- The text colors (would be equivalent to  `--text-colors`)
- Whether to hide the ASCII logo
- image path to display by default instead of the ASCII logo
- ...

Palette:

- Which glyph to use for the palette (use "●" instead of " ")
- Whether to hide the palette.
- ...

Info lines:

- Define the text to display as a title for each info lines (use "🖴" instead of "Size")
- ...

The configuration file could be placed in `${HOME}/.config/onefetch/config.conf`

### Motivation 🔦

Allow users to customize the look and feel of onefetch by creating themes and config files that can shared and enjoyed by others.

For reference --> https://github.com/Chick2D/neofetch-themes

amelenty:
Resolves #745.

### General
- [x] Add serialization/deserialization of Config into a config file
- [ ] Add tests for new code logic
- [ ] Add glyphs for palette and separators

### Bugs / Improvements
- Right now, the config file is written to config dir directly instead of creating a `onefetch` subdirectory. I'm still figuring out how to best ensure that a parent dictionary is created if necessary.
- I want to ensure that if the user-supplied config is malformed, `onefetch` prints a warning about that and uses the default config values. Or would it be better to fail completely instead of showing the info with default options?

### Other questions
- `cli.rs` got a lot longer. I'm thinking of moving config-specific code to a separate file, but that's most of `cli.rs`. What would be a good way to separate them?
- Which of the config value sources has more priority, the config file or the command line arguments? 
  - The way I see it, the user should be able to override any option via command line, regardless of what's in the config file. 
  -  But there are options where it's not practical. For example, it would make sense if `exclude` paths from both sources counted. Otherwise, the user would need to write out all the excluded paths in the config if they want to include an extra path.
  - As of right now, for consistency, all `bool` values are merged basically as an "or" (if it's `true` in the config file, then the command-line argument doesn't matter), and all other values are completely overwritten. This is obviously not ideal, and I plan to change it, but I'd like to hear your perspective on what I should change it to.
- I feel like there's an easy attribute solution for serializing `MyRegex` instead of the trait implementation I wrote, but I couldn't think of one.

I'm very new to Rust, so if something I wrote is not idiomatic, I'll be happy to change that if you point it out.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -8,13 +8,16 @@ use clap::AppSettings;
 use clap::{Command, Parser, ValueHint};
 use clap_complete::{generate, Generator, Shell};
 use regex::Regex;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::de::{self, Visitor};
 use std::env;
 use std::io;
+use std::fmt;
 use std::path::PathBuf;
 use std::str::FromStr;
 use strum::IntoEnumIterator;
 
-#[derive(Clone, Debug, Parser, PartialEq, Eq)]
+#[derive(Clone, Debug, Parser, PartialEq, Eq, Deserialize, Serialize)]
 #[clap(version, about, long_about = None, rename_all = "kebab-case")]
 #[clap(global_setting(AppSettings::DeriveDisplayOrder))]
 pub struct Config {
@@ -146,6 +149,7 @@ pub struct Config {
     pub r#type: Vec<LanguageType>,
     /// If provided, outputs the completion file for given SHELL
     #[clap(long = "generate", value_name = "SHELL", arg_enum)]
+    #[serde(skip)]
     pub completion: Option<Shell>,
 }
 
@@ -184,7 +188,7 @@ pub fn print_completions<G: Generator>(gen: G, cmd: &mut Command) {
     generate(gen, cmd, cmd.get_name().to_string(), &mut io::stdout());
 }
 
-#[derive(clap::ValueEnum, Clone, PartialEq, Eq, Debug)]
+#[derive(clap::ValueEnum, Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
 pub enum When {
     Auto,
     Never,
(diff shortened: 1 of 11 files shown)


src/cli.rs
```diff
@@ -143,12 +166,59 @@ pub struct Config {
         short = 'T',
         arg_enum,
     )]
+    #[merge(strategy = overwrite_vector)]
     pub r#type: Vec<LanguageType>,
+    /// Specify a custom path to a config file.
+    /// Default config is located at ${HOME}/.config/onefetch/config.conf.
```
spenserblack, line 172: Just an FYI that this is only the default path on Linux if you're using `dirs`, AFAIK.

src/cli.rs
```diff
@@ -253,36 +323,6 @@ mod test {
...
-            true_color: When::Auto,
-            show_logo: When::Always,
-            text_colors: Default::default(),
-            iso_time: Default::default(),
-            email: Default::default(),
-            include_hidden: Default::default(),
-            r#type: vec![LanguageType::Programming, LanguageType::Markup],
-            completion: Default::default(),
-        }
-    }
```
spenserblack, line 285: A wise choice to move this into a `Default` implementation :+1:

src/configuration.rs
```diff
@@ -0,0 +1,47 @@
...
+                write_default_config(&default_path);
+                Err(anyhow!("Default config file did not exist: {:?}", default_path))
+            }
+        } 
+    }
+}
+
+fn write_default_config<P: AsRef<Path>>(default_path: &P) {
+    let toml = toml::to_string(&Config::default()).expect("Config should be serializable");
+    fs::write(default_path, toml).expect("Should be able to write to config dir");
```
spenserblack, line 36: Honestly, maybe these should just be `unwrap`s to show the underlying error. For example, `fs::write` might fail if there's no more space on the disk. If that happens, the user should probably see that, and only that.

spenserblack (commented):
>I want to ensure that if the user-supplied config is malformed, `onefetch` prints a warning about that and uses the default config values.

Just an FYI I did exactly that in another project 🙂 
https://github.com/spenserblack/repofetch/blob/d87c577bc1987f0a7bb7434dfeb5a5ac444d9584/src/main.rs#L97-L108
https://github.com/spenserblack/repofetch/blob/d87c577bc1987f0a7bb7434dfeb5a5ac444d9584/src/configuration/mod.rs#L85-L119

Though, in hindsight, the way I did it can hide the reason *why* the config is malformed, confusing users. I think it's important to show the user exactly why the config couldn't be read. IMO either a warning or an error would both be OK as long as the necessary information for the user to fix the malformed config is accessible.

>I'm thinking of moving config-specific code to a separate file, but that's most of `cli.rs`. What would be a good way to separate them?

I think the config logic can be moved to a `configuration.rs` file, with `cli.rs` focused on both creating the CLI arguments, and *using* the CLI arguments to build that config, if that makes sense. I.e. `configuration.rs` returns the config struct, `cli.rs` mutates that struct.

>Which of the config value sources has more priority, the config file or the command line arguments?

I think CLI arguments should always override config file values. Yes, it can be annoying if options like `exclude` aren't merged, but I think it can be confusing if CLI arguments don't all have the same behavior. IMO using a CLI arg has always meant "use this instead". There can be a new CLI arg, `extra-exclude`, to "merge" this value if it's needed. As an example, `pip` has both `--index-url` and `--extra-index-url` args -- one overwrites, one appends.

PR #800, opened 2022-10-06
Automate publish to crates.io

spenserblack:
This is pretty much just copied and pasted from [here](https://github.com/spenserblack/repofetch/blob/342ab94c3db7275048b8ed34581347e9e6621533/.github/workflows/release.yml).

I thought we were doing this already, but a quick search on cs.github.com didn't turn anything up.

As it doesn't use the existing file from `make build` (`cargo publish` always rebuilds AFAIK), I
made this a separate job.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -62,3 +62,24 @@ jobs:
           installers-regex: '\.exe$'
           delete-previous-version: 'true'
           token: ${{ secrets.WINGET_TOKEN }}
+
+  cargo-publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          components: clippy
+
+      - name: Publish to crates.io
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -62,3 +62,24 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          components: clippy
+
+      - name: Publish to crates.io
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        run: cargo publish
```
o2sh, line 85:
```suggestion
      - name: Publish to crates.io
        if: matrix.os == 'ubuntu-latest'
        env:
          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
        run: cargo publish
```

Why not just make it a step like the others?

o2sh (approved):
Thanks a lot @spenserblack
This will definitely save me some time.

I created the action secret containing the crates.io API token. So we should be all set for the next release.

image

.github/workflows/cd.yml
spenserblack: If it's a separate job it can run in parallel without waiting on other steps or making other steps wait on it. But I'm not too attached to making it a separate job, so if you prefer it as a step then :+1: from me 🙂 
I'll be happy either way as long as publishing to crates.io is somehow part of the workflow :smiley:

PR #812, opened 2022-10-09
testing get_git_username using git-testtools for #700

atluft:
Trying to figure out how to add a git repo test.

Testing:
```
cargo test -- --nocapture
```


atluft:
Hi @o2sh and @spenserblack,
Please advise about the code coverage failures below, is the new test failing or is it that code coverage went negative by adding more code into measurement, or something else :-) ?

o2sh:
I suspect an [update of tarpaulin](https://github.com/xd009642/tarpaulin/releases/tag/0.22.0) may be the cause for the drop:

image

Indeed we always pull the latest version when generating the codecov report.

```yml
- name: Install cargo-tarpaulin
  uses: actions-rs/install@v0.1
  with:
    crate: cargo-tarpaulin
    version: latest 
````

File: src/info/title.rs [hunk 1]
```diff
@@ -1,83 +1,155 @@
-use super::get_style;
```
spenserblack, line 1: :thinking: Looks like this file was (correctly) converted to LF line endings.
For the sake of a reasonable diff, I think we should convert all files to LF so that it's easier to compare this and https://github.com/o2sh/onefetch/pull/813

atluft: I saw the CR LF in title.rs, was confused and ran dos2unix on it, but then had a bunch of diffs for this change so continued with the "match existing" attittude.

Respond one more time to confirm ok to convert the line endings to LF and I will do so.

spenserblack: >Oops, did I do that?

You *correctly* did that (See `.editorconfig` file) :laughing: 

I *incorrectly* set up an EditorConfig and never enforced its settings in a workflow.

spenserblack:
Thanks @atluft! If you feel comfortable rebasing onto `main` then can you please do so?
There's a good chance that this might be squash-merged, so a normal merge commit is fine, too.

`git checkout --ours` and `git checkout --theirs` are useful when resolving these full-file conflicts.

atluft:
@spenserblack 
I think this PR is in good shape now.
git rebase main, correct conflicts, continue, push force at end.

squash is a wonderful feature!

atluft:
Discovered the "Ready for Review" button!

src/info/title.rs
```diff
@@ -81,3 +81,72 @@ impl std::fmt::Display for Title {
...
     }
 }
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use git_repository::{open, Repository, ThreadSafeRepository};
+    use git_testtools;
+    use owo_colors::AnsiColors;
+
+    pub type Result<T = ()> = std::result::Result<T, Box<dyn std::error::Error>>;
```
spenserblack, line 91:
```suggestion
    type Result<T = ()> = std::result::Result<T, Box<dyn std::error::Error>>;
```
I don't think this needs to be `pub`. It's not going to be used outside of the `tests` mod.

src/info/title.rs
```diff
@@ -81,3 +81,72 @@ impl std::fmt::Display for Title {
...
+        title.git_version = "git version 2.37.2".to_string();
+        let expected_title = [
+            "\u{1b}[31;1monefetch-committer-name",
+            "\u{1b}[0m \u{1b}[37;1m~\u{1b}[0m ",
+            "\u{1b}[31;1mgit version 2.37.2",
+            "\u{1b}[0m\n",
+            "\u{1b}[34m--------------------------------------------\u{1b}[39m\n",
+        ]
+        .join("");
+        assert_eq!(format!("{}", title), expected_title);
```
spenserblack, line 133: I'll have to dig into the code again to be sure (just looking at the PR diff itself right now), but I think this test will pass on some environments and fail on others. For example, someone could clone this repo to an environment where `NOCOLOR=true` is set, and it will fail. Might want to normalize some environment variables in these tests so that they pass. I wonder if there's a Rust testing library that sets environment variables and unsets them on `drop` (otherwise an explicit cleanup function).

atluft: Good comment, didn't think too much about other environments. All comments on testing a format function like this welcome.

spenserblack: OK, you can ignore my comment. I had assumed that `owo-colors` did this automatically, but it turns out it's something we would need to explicitly opt in to.
https://github.com/jam1garner/owo-colors/blob/ecafaabccc8cae02c0d54d73b4a67e725e604ddb/Cargo.toml#L20-L22
https://github.com/jam1garner/owo-colors/blob/ecafaabccc8cae02c0d54d73b4a67e725e604ddb/examples/supports_color.rs

src/info/title.rs
```diff
@@ -81,3 +81,72 @@ impl std::fmt::Display for Title {
...
+
+    fn basic_repo() -> Result<Repository> {
+        let name = "basic_repo.sh".to_string();
+        let repo_path = git_testtools::scripted_fixture_repo_read_only(name)?;
+        let safe_repo = ThreadSafeRepository::open_opts(repo_path, open::Options::isolated())?;
+        Ok(safe_repo.to_thread_local())
+    }
+
+    #[test]
+    fn test_get_git_username() -> Result<()> {
```
spenserblack, line 101:
```suggestion
    fn test_get_git_username() -> Result {
```
Since you set `T` to default to `()`, you should be able to remove these.

Otherwise, if you feel like `T` should *always* be specified, I'd recommend removing the type's default above.

src/info/title.rs
```diff
@@ -81,3 +81,72 @@ impl std::fmt::Display for Title {
...
+        let username = get_git_username(&repo);
+        assert_eq!(
+            username, "onefetch-committer-name",
+            "see git repo local config committer.name"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_title_format() -> Result<()> {
```
spenserblack, line 113:
```suggestion
    fn test_title_format() -> Result {
```

spenserblack (approved):
LGTM, thank you for helping us with testing! Leaving one minor nitpick.

PR #813, opened 2022-10-10
Improve code coverage of src/info/repo/commits.rs

alessandroasm:
Part of task #700

PR #815, opened 2022-10-10
Convert line endings to LF

spenserblack:
This should result in more reasonable diffs of future contributions
where a contributor's text editor would automatically convert line
endings.

See #812, #813

******

We may want to see if a linter/formatter can check this. TBH I'm surprised this
isn't the default behavior of `cargo fmt -- --check`.


o2sh:
> TBH I'm surprised this isn't the default behavior of cargo fmt -- --check.

I think we are missing the following `.rustfmt.toml`

```toml
# This file defines the Rust style for automatic reformatting.
# See also https://rust-lang.github.io/rustfmt

# Rust language edition to be used by the parser.
edition = "2021"

# Line endings will be converted to \n.
newline_style = "Unix"
```

PR #822, opened 2022-10-12
Add test info mod rs

atluft:
Another PR to explore testing using the `git-testtool` module.
This test is intentionally trying to open a "bare git" repo that onefetch should flag as an error and provide a specific error message.

All comments are welcome!


atluft:
TLDR: to test the expected repo, use a `mut` on config variable and change `config.input` after initialization.

The `discover` function was walking back up the tree and finding the `onefetch` repo itself instead of the intended bare repo created for the test. In most cases that is what any user would want, but in this case the test is to try to open a bare repository and observe that the error is triggered.

src/info/mod.rs
```diff
@@ -364,4 +366,27 @@ mod tests {
...
+    fn test_repo(name: &str) -> Result<Repository> {
+        let repo_path = git_testtools::scripted_fixture_repo_read_only(name)?;
+        let safe_repo = ThreadSafeRepository::open_opts(repo_path, open::Options::isolated())?;
+        Ok(safe_repo.to_thread_local())
+    }
+
+    #[test]
+    fn test_bare_repo() -> Result<()> {
+        let repo = test_repo(&"bare_repo.sh".to_string())?;
+        let mut config = Config::parse_from(&["."]);
```
Byron, line 381: `"."` is definitely picking up the onefetch repository because the CWD is the onefetch repository, not the test directory. Passing `repo_path` from line 373 might be better, but I acknowledge that I don't understand what `parse_from(…)` really does.

Here the actual `repo` seems entirely unused.

Byron (commented):
The usage of `git_testtools` generally looks fine, and I chose not to comment on the test itself is conducted as it might be what's desired in this codebase.

src/info/mod.rs
spenserblack: @atluft You might want to look at some of Rust's macros, and Cargo's default environment variables.
For example, `env!("CARGO_MANIFEST_DIR")` should get you the path to the root of this crate, which you can then concatenate.

atluft: Yeah, I need help on constructing a Config with the repo path created for the test.

I don't understand how to set `config.input` to anything but `.`

**TRY 1**
``` rust
let mut config : Config;
config.input = repo.path().to_path_buf();
```
That was rejected like this:
```rust
385 |         config.input = repo.path().to_path_buf();
    |         ^^^^^^^^^^^^ `config` partially assigned here but it isn't fully initialized
```

**TRY 2 and 3**
``` rust
let config = Config::parse_from(&[repo.path().to_path_buf()]);
let config = Config::parse_from(&[repo.path().to_display().to_string()]);
```

`parse_from` seems to come from here, but I don't know how to read this code, notice the default value is `.`

```rust
#[derive(Clone, Debug, Parser, PartialEq, Eq)]
#[command(version, about, long_about = None, rename_all = "kebab-case")]
pub struct Config {
    /// Run as if onefetch was started in <input> instead of the current working directory
    #[arg(default_value = ".", hide_default_value = true, value_hint = ValueHint::DirPath)]
    pub input: PathBuf,
```

spenserblack: @atluft I'll try and dig into this more to help later, but I hope this can get you started:

`parse_from` comes from the `Parser`, and is documented [here](https://docs.rs/clap/4.0.15/clap/trait.Parser.html#method.parse_from). There's definitely a lot of macro usage, which can make things less explicit, but what we're saying is
- `input` is a command-line argument
- if a value wasn't passed to the command-line, `input` is `"."`
- `input`'s default value is not shown in the help message (`onefetch -h`)
- `input` should *look like* a directory (useful for completion scripts)

Also, if you're going to build the `Config` struct manually, you will need to define *all* fields, not just `input`.

******

Keep in mind that testing is only easy when the project is testable. The design had progressed *a lot* before testing was considered, so some refactoring may need to be done to make the code more testable.
For instance, for your 1st try, it would be a lot easier if `Config` implemented `Default`. Which is kind of implemented here:
https://github.com/o2sh/onefetch/blob/00d3905a3b63c362b58917e50b81582492ed2834/src/cli.rs#L259-L288

atluft: Thanks @spenserblack,
Very helpful, especially the link to [parse_from](https://docs.rs/clap/4.0.15/clap/trait.Parser.html#method.parse_from)
Will try to create a refactoring for testing idea, for review here.

I agree this current approach isn't great, thank you for the ideas.

PR #827, opened 2022-10-17
Add test info mod rs

atluft:
Rebased PR #822 
Didn't realize that would close the other PR, sorry.


src/info/mod.rs
```diff
@@ -364,4 +373,135 @@ mod tests {
...
+                "          \u{1b}[31m\u{25CF}\u{1b}[39m \u{1b}[37mGo (100.0 %)\u{1b}[39m \u{1b}[39m\n",
+                "\u{1b}[37mDependencies\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37minfo test dependencies dependencies\u{1b}[39m\n",
+                "\u{1b}[37mAuthor\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37m\u{1b}[37m100% author1 name 100\u{1b}[39m\u{1b}[39m\n",
+                "\u{1b}[37mLast change\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37minfo test last_change last_change\u{1b}[39m\n",
+                "\u{1b}[37mRepo\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37minfo test repo repo_url\u{1b}[39m\n",
+                "\u{1b}[37mCommits\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37minfo test commits number_of_commits\u{1b}[39m\n",
+                "\u{1b}[37mLines of code\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37m100\u{1b}[39m\n",
+                "\u{1b}[37mSize\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37minfo test size repo_size (100 files)\u{1b}[39m\n",
+                "\u{1b}[37mLicense\u{1b}[0m\u{1b}[37m:\u{1b}[0m \u{1b}[37minfo test license license\u{1b}[39m\n" ].join("");
```
o2sh, line 482: What matters here (IMO) are the information displayed, not the formatting.

In that regard, I think asserting with a `contains` instead of looking for an exact match would suffice and improve the readability quite a lot. cf. cc7a59c for an example.

src/info/mod.rs
```diff
@@ -364,4 +373,135 @@ mod tests {
...
+                underline: DynColors::Ansi(AnsiColors::White),
+                subtitle: DynColors::Ansi(AnsiColors::White),
+                colon: DynColors::Ansi(AnsiColors::White),
+                info: DynColors::Ansi(AnsiColors::White),
+            },
+            no_color_palette: true,
+            no_bold: true,
+            dominant_language: Language::Go,
+            ascii_colors: vec![DynColors::Ansi(AnsiColors::White)],
+        };
```
o2sh, line 463: Why didn't you make use of `git-testtools` here 🤔 ? This seems like a good use case

spenserblack: The one argument I can see is that it's more of a unit test to simply have the test be "with the given `Info`, does it display correctly?" Parsing an actual git repository is closer to an integration test ("Do we read a git repository *and* display its parsed information correctly?").

o2sh: Right, I definitely had in mind an integration test for the `Info` struct - which is more or less just a container of `InfoFields` - which are individually unit tested. All that setup just to unit test the `fmt` method feels like a missed opportunity IMO.

src/info/mod.rs
atluft: @o2sh , thank you, will reduce to important aspects using contains and example

src/info/mod.rs
atluft: Thank you for the comments, wasn't sure how tight the formatting needed to be. This is an extreme of the code as implemented is absolutely correct and any changes are flagged as failures.

Will revisit this testing and explore using `git-testtools`.
[62% covered ](https://app.codecov.io/gh/o2sh/onefetch/blob/add-test-info-mod-rs/src/info/mod.rs)

spenserblack: If we ever want to *unit* test `Info` a bit more, I think it would be a good idea to implement `Default` on it. Then we can just unit test like so:
```rust
#[test]
fn test_foo() {
    let info = Info { foo: value, ..Default::default() };
    // ...
}
```

atluft:
Trying a new approach in 3e2757d
Test will creates repo and expected result using shell script via `git-testttools`
Test will read expected result from file created by shell script
Test will perform action
Test will compare actual to expected

spenserblack:
These assertion diffs are pretty hard to read :laughing:
I've used [pretty_assertions](https://docs.rs/pretty_assertions/1.3.0/pretty_assertions/) before to get better diffs.

atluft:
Thank you @spenserblack , here is a pretty assertion:

image

atluft:
@spenserblack  and @o2sh 
Really like the output of pretty_assert that is very easy to see the issues.

What are thoughts on using regular expressions in the expected output?
I have the wrong regex for REGEX_ANYTHING, ".*" is only good for end of lines.
Wanted a quick review and collect ideas before putting too much effort into this.

Would this be better if a phrases were used: MATCH_UNTIL_END_OF_LINE or MATCH_WORD ?

tests/fixtures/language_repo.sh
```diff
@@ -39,13 +39,14 @@ git commit -q -am c2
...
-# onefetch --show-logo never --ascii-language <language>
+# onefetch --show-logo never --ascii-language <language> | sed 's/%/%%/g' 
+# REGEX_ANYTHING manually inserted to handle git version numbers on different servers
 #
 
 case "${1}" in
   verilog)
 cat >expected << '__expected_output__'
-[39;1monefetch-committer-name[0m [39;1m~[0m [39;1mgit version 2.37.2[0m
+[39;1monefetch-committer-name[0m [39;1m~[0m [39;1mgit version REGEX_ANYTHING[0m
```
spenserblack, line 49: So as I understand it the intent is to write the expected value to a file and read that file in the test. TBH I think it would be better if the expected output just existed in the repo, in a file called for example `test_language_repo.stdout`. This would allow the usage of the `include_str!` macro so the test doesn't have to read a file at runtime. Also, replacing `REGEX_ANYTHING` with regex seems a bit hacky IMO.

I wish we could simply assert the exact output, but without a reasonable way to make the git version reproducible across systems (preferably in CI *and* local runs), I guess this can't be done. Perhaps this test should just assert some regex matches to ensure that the expected data exists, and in the expected format.

spenserblack (commented):
Those diffs look much better now, thank you!

src/info/mod.rs
```diff
@@ -408,13 +406,31 @@ mod tests {
...
         let info = Info::new(&config)?;
         let info_str = format!("{}", info);
-        let expected_info = fs::read_to_string(repo.path().join("..").join("expected"))?;
-        let escaped_expected_info = regex::escape(&expected_info);
-        let expected_info_expression =
-            escaped_expected_info.replace("MATCH_UNTIL_END_OF_LINE", ".*");
-        let re = Regex::new(format!(r#"{}"#, expected_info_expression).as_str()).unwrap();
-        if !re.is_match(&info_str) {
-            assert_eq!(info_str, expected_info);
+        let info_u8 = strip_ansi_escapes::strip(&info_str)?;
```
atluft, line 409: @spenserblack  and @o2sh 
The function `strip_ansi_escapes::strip` strips the colors and allows a simpler regex which is more readable.

I selected to use a multi-line regex so that the order of fields can be verified in the test.

src/info/mod.rs
atluft: lol, fun to code, but not what is needed, resolving in favor of more readable code and info structure created from real repo.

tests/fixtures/language_repo.sh
atluft: makes sense, in 7b26b88 moves the regex into code, by removing the escape characters the regex becomes more readable too.

src/info/mod.rs
```diff
@@ -408,13 +406,31 @@ mod tests {
...
+        ]
+        .join(".");
+        let re = Regex::new(&expected_info)?;
+        if !re.is_match(&simple_info_str) {
+            assert!(
+                false,
+                "OOPS, REGEX\n{}\nDOESNT MATCH\n{}",
+                expected_info, simple_info_str
+            );
         }
```
spenserblack, line 434:
```suggestion
        assert!(
            re.is_match(&simple_info_str),
            "OOPS, REGEX\n{}\nDOESNT MATCH\n{}",
            expected_info, simple_info_str
        );
```
Asserting a hardcoded `false` inside an if statement seems a bit unnecessary IMO.

src/info/mod.rs
```diff
@@ -408,13 +406,31 @@ mod tests {
...
+            r".+Verilog \(100.0 %\).+",
+            r"Author: 100% author 2",
+            r"Last change: 22 years ago",
+            r"Repo: https://github.com/user/repo.git",
+            r"Commits: 2",
+            r"Lines of code: 1",
+            r"Size: 6 B \(1 file\)",
+        ]
+        .join(".");
+        let re = Regex::new(&expected_info)?;
```
spenserblack, line 427:
```suggestion
        let re = Regex::new(&expected_info).unwrap();
```
I think this should actually be unwrapped instead of propagating the error. If the regex fails to compile, that means the test was bad, not that the tested code is failing.

src/info/mod.rs
```diff
@@ -408,13 +406,31 @@ mod tests {
...
+            r"Language:",
+            r".+Verilog \(100.0 %\).+",
+            r"Author: 100% author 2",
+            r"Last change: 22 years ago",
+            r"Repo: https://github.com/user/repo.git",
+            r"Commits: 2",
+            r"Lines of code: 1",
+            r"Size: 6 B \(1 file\)",
+        ]
+        .join(".");
```
spenserblack, line 426: I'd be interested in moving this to a separate file, for example `tests/regex/test_language_repo.stdout.regex` or something, and using `include_str!`. The regex could potentially be more accurate by checking explicitly for newlines (`\r?\n`) instead of using the `s` flag.
Also, [the `x` flag](https://docs.rs/regex/latest/regex/#example-replacement-with-named-capture-groups) may be worth looking into to help make the regex more readable.

spenserblack (changes requested):
Testing against a regex, and stripping out the ANSI escape sequences, looks much better IMO! :+1:
Almost there! Just a few more comments.

src/info/mod.rs
atluft: Yes your are right, makes me laugh to see it now. Thank you for pointing it out.

src/info/mod.rs
atluft: Learning a lot in that last post, thanks for pointing those methods out.

src/info/mod.rs
```diff
@@ -408,27 +409,16 @@ mod tests {
...
-            r"Repo: https://github.com/user/repo.git",
-            r"Commits: 2",
-            r"Lines of code: 1",
-            r"Size: 6 B \(1 file\)",
-        ]
-        .join(".");
-        let re = Regex::new(&expected_info).unwrap();
+        #[cfg(windows)]
+        let expected_regex = include_str!("..\\..\\tests\\regex\\test_verilog_repo.stdout.regex");
+        #[cfg(unix)]
```
spenserblack, line 414:
```suggestion
```
I'm pretty sure the path will be resolved correctly on Windows.

tests/regex/test_verilog_repo.stdout.regex
```diff
@@ -0,0 +1,13 @@
+(?s)onefetch-committer-name ~ git version ([^\r\n]+)\r?\n(?x) # MATCH LINE 1 CAPTURE GIT VERSION
```
spenserblack, line 1:
```suggestion
onefetch-committer-name ~ git version ([^\r\n]+)\r?\n(?x) # MATCH LINE 1 CAPTURE GIT VERSION
```
I don't think the `s` flag is needed if you're explicitly matching `\n`. Also, instead of disabling and re-enabling `x` on every line, you might just want to change ` ` into `\ `, which should be matched instead of ignored even when using `x`. Not sure which would be more readable TBH :thinking:

src/info/mod.rs
atluft: ok, will give it a try, I read that [the include_str path was operating system dependent](https://doc.rust-lang.org/std/macro.include_str.html#). I see there is a CI test on windows, I don't have any ability to test that, all OSX or linux on my side.

tests/regex/test_verilog_repo.stdout.regex
atluft: Ah, that is how to match spaces in x modified regex, yeah I couldn't figure out how to match spaces.  Thanks for explaining that aspect.

atluft: yep path separator worked, I do not understand what the documentation is talking about.
[dfe994c](https://github.com/o2sh/onefetch/pull/827/commits/dfe994cc7dfb7edf7ef590ef948d8e913f748e08)

src/info/mod.rs
spenserblack: Yeah TBH those docs seem a bit misleading (maybe worth a PR? :wink:), but I think it's just a warning that Windows path separators will *only* work on Windows. I think modern Windows OSs always treat `/` just like `\` in paths (at least that happens when I accidentally use the wrong separator on Windows :lauging:). Additionally, AFAIK most modern programming languages will treat `/` as OS-independent.

src/info/mod.rs
```diff
@@ -364,4 +371,52 @@ mod tests {
...
+    #[test]
+    fn test_bare_repo() -> Result {
+        let repo = test_repo(&"bare_repo.sh")?;
+        match Info::init_repo_path(&repo) {
+            Ok(_info) => assert!(false, "oops, info was returned on a bare git repo"),
+            Err(error) => assert_eq!(
+                "please run onefetch inside of a non-bare git repository",
+                error.to_string()
+            ),
+        };
```
spenserblack, line 392: Another nitpick: instead of asserting false, maybe assert `.is_err()`, then unwrap the error for the following `assert_eq!`.

src/info/mod.rs
```diff
@@ -364,4 +371,52 @@ mod tests {
...
+    fn test_language_repo() -> Result {
+        let repo_path = git_testtools::scripted_fixture_repo_read_only_with_args(
+            "language_repo.sh",
+            ["verilog"],
+        )
+        .unwrap();
+        let safe_repo =
+            ThreadSafeRepository::open_opts(repo_path.join("verilog"), open::Options::isolated())?;
+        let repo = safe_repo.to_thread_local();
+        let mut config = Config::parse_from(&["."]);
```
spenserblack, line 406: Is this comment chain still relevant? https://github.com/o2sh/onefetch/pull/822#discussion_r994138849
Perhaps this should be an absolute path (defined at compile-time with macros). Or some refactoring to the implementation so that there's an easy way to create a `Config` struct besides parsing from command-line args.

spenserblack (commented):
Thanks for putting up with all of my nitpicks!
This is looking pretty good, I just have a few more notes.

src/info/mod.rs
atluft: I only have one perspective and experience with this so am reluctant to do any refactoring without significant coaching ;-) It would be nice to create a new Config and then just create a new Info using the path returned from `git-testtools`.

@Byron would be able to speak about using absolute paths versus non absolute paths.

The rest of this is some review that I think helps:

This PR uses `git-testtools` approach of using scripts to create known repositories.
The directory `tests/fixtures/generated-do-not-edit` is a  new part of that method.

How to create and see a clean generated test directory:
```text
rm -fr tests/fixtures/generated-do-not-edit
cargo test -- --nocapture 
```

What does the generated directory look like?
`tests/fixtures/generated-do-not-edit` directory looks like this (only directories are shown, no files):
```text
tree -d tests/fixtures/generated-do-not-edit/
tests/fixtures/generated-do-not-edit/
├── bare_repo
│   └── 3723886778-unix
│       ├── hooks
│       ├── info
│       ├── objects
│       │   ├── info
│       │   └── pack
│       └── refs
│           ├── heads
│           └── tags
├── basic_repo
│   └── 2065824803-unix
└── language_repo
    └── 320037018-unix
        └── verilog
```

@spenserblack 
Maybe using the shell script to generate files is helpful because the generated path has some dependencies not related to an absolute path?

Here is a script created `verilog` test repo:

```text
tree tests/fixtures/generated-do-not-edit/language_repo
tests/fixtures/generated-do-not-edit/language_repo
└── 320037018-unix
    └── verilog
        ├── code.vg
        ├── expected_json
        └── expected_regex
```

The `language_repo` shell script was used on `unix` using a specific `320037018` script.

Here is what the tree looks like after adding another language, in this case I added `vhdl`. The test code for this likely belongs at the top level and might be useful to add a `expected_ansi_art` file to the scripts abilities.

```text
tree tests/fixtures/generated-do-not-edit/language_repo
tests/fixtures/generated-do-not-edit/language_repo
└── 320037018-unix
    ├── verilog
    │   ├── code.vg
    │   ├── expected_json
    │   └── expected_regex
    └── vhdl
        ├── code.vhdl
        ├── expected_json
        └── expected_regex
```

src/info/mod.rs
atluft: Sweet, so much  better.
```rust
        let res = Info::init_repo_path(&repo);
        assert!(res.is_err(), "oops, info was returned on a bare git repo");
        assert_eq!(
            res.unwrap_err().to_string(),
            "please run onefetch inside of a non-bare git repository"
        );
```

src/info/mod.rs
```diff
@@ -417,6 +417,13 @@ mod tests {
             expected_regex,
             simple_info_str
         );
+
+        let v = serde_json::to_value(info).unwrap();
+        let expected_json: serde_json::Value = serde_json::from_str(
+            &include_str!("../../tests/json/test_verilog_repo.serialize.json"),
```
spenserblack: Super-small nitpick: since this is basically a snapshot test, so maybe use this name?
```suggestion
            &include_str!("../../tests/json/test_verilog_repo.snapshot.json"),
```

src/info/mod.rs
```diff
@@ -417,6 +417,13 @@ mod tests {
...
+        ).unwrap();
+        assert_eq!(v, expected_json);
```
spenserblack: I think it makes sense to actually test against the JSON *string*, not the JSON struct. This would truly make it a snapshot test, and it would be easier to diff the two IMO.

spenserblack (commented):
Now I want to write a macro to better support snapshot testing :laughing:

src/info/mod.rs
spenserblack: > so am reluctant to do any refactoring without significant coaching ;-)

So this function currently only exists in a test mod: 
https://github.com/o2sh/onefetch/blob/3a3b1c6c04907d5461deb6ed609a2543f5be3776/src/cli.rs#L259-L287

But I think there is value in moving this logic to an implementation of the `Default` trait, making `Default::default()` available throughout the source code where a `Config` needs to be constructed (especially tests).

With `Default` implemented, you can set only the config values that you *need* to set, and leave everything else as a "do not care" default value.

Hopefully [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=%23!%5Ballow(dead_code)%5D%0A%23%5Bderive(Debug)%5D%0Aenum%20Color%20%7B%0A%20%20%20%20Black%2C%0A%20%20%20%20White%2C%0A%20%20%20%20RGB(u8%2C%20u8%2C%20u8)%2C%0A%20%20%20%20%2F%2F%20etc.%0A%7D%0A%0A%23%5Bderive(Debug)%5D%0Astruct%20Output%20%7B%0A%20%20%20%20foreground%3A%20Color%2C%0A%20%20%20%20background%3A%20Color%2C%0A%20%20%20%20text%3A%20String%2C%0A%7D%0A%0Aimpl%20Default%20for%20Output%20%7B%0A%20%20%20%20fn%20default()%20-%3E%20Output%20%7B%0A%20%20%20%20%20%20%20%20Output%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20foreground%3A%20Color%3A%3AWhite%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20background%3A%20Color%3A%3ABlack%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20text%3A%20Default%3A%3Adefault()%2C%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%7D%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20output%20%3D%20Output%7Btext%3A%20%22Hello%2C%20World!%22.into()%2C%20..Default%3A%3Adefault()%7D%3B%0A%20%20%20%20println!(%22output%20struct%3A%20%7B%3A%3F%7D%22%2C%20output)%3B%0A%7D) shows how this can be used.

spenserblack:
Here I was about to make a snapshot testing library, and in my research found that one exists :laughing:
https://crates.io/crates/insta

It might be worth using this library for the JSON output test.

src/info/mod.rs
atluft: Sure thing, will change the name to snapshot.

src/info/mod.rs
atluft: Agree, string based testing makes a lot more sense.
See 8ab0bb2

atluft:
Hav a look at  [6486d50](https://github.com/o2sh/onefetch/pull/827/commits/6486d506e9e4b39dcce1f8038bf4a1ef191f0305)
uses insta snapshots for the JSON
Neato.

src/info/mod.rs
atluft: Nice idea, removes the need for the `.` weirdness when constructing the Config object.
[2f04c83](https://github.com/o2sh/onefetch/pull/827/commits/2f04c8354fe046e3e436f33cd8e9debbce36e64e)

src/info/mod.rs
```diff
@@ -364,4 +372,85 @@ mod tests {
...
+        let info_str = format!("{}", info);
+        let info_u8 = strip_ansi_escapes::strip(&info_str).unwrap();
+        let simple_info_str = std::str::from_utf8(&info_u8).unwrap();
+        assert!(
+            re.is_match(&simple_info_str),
+            "OOPS, REGEX\n{}\nDOESNT MATCH\n{}",
+            re.to_string(),
+            simple_info_str
+        );
+    }
```
spenserblack, line 404: `test_language_repo_again` is a bit vague, maybe something like `assert_info_str_matches`, or `foo_helper`.

spenserblack (commented):
Great work!

>The .gitignore file has *.snap in it to ignore snapshots.

We might want to adjust the `.gitignore` so that it ignores only the files it *should* ignore.

@o2sh Do you remember why `*.snap` is in the `.gitignore`?

src/info/mod.rs
```diff
@@ -364,4 +372,85 @@ mod tests {
...
+        let repo = safe_repo.to_thread_local();
+
+        // TEST JSON SERILIZER FIRST
+        let mut config: Config = Config {
+            input: repo.path().to_path_buf(),
+            ..Default::default()
+        };
+        let info = Info::new(&config).unwrap();
+        assert_json_snapshot!(
+            serde_json::to_value(info).unwrap(),
```
o2sh, line 424:
```suggestion
            info,
```

The info struct already implements `Serialize`

src/info/mod.rs
atluft: see [69c6229](https://github.com/o2sh/onefetch/pull/827/commits/69c62296af280e0b7bc05ea2eb00ea2564e0ff92)

src/info/mod.rs
atluft: see [69c6229](https://github.com/o2sh/onefetch/pull/827/commits/69c62296af280e0b7bc05ea2eb00ea2564e0ff92)

atluft:
Change [69c6229](https://github.com/o2sh/onefetch/pull/827/commits/69c62296af280e0b7bc05ea2eb00ea2564e0ff92) caused a snapshot change, testing gitignore change ;-)

Also, gave another chance to use `cargo insta review` that is so cool how it "fixes" up the snapshot file for the test.

spenserblack (approved):

spenserblack:
>that is so cool how it "fixes" up the snapshot file for the test.

Yeah, snapshot testing is pretty interesting. It leans a bit towards "acceptance". That is, instead of a human writing a test for the machine to run and the code to pass, a human reviews the output and tells the machine if they like it.

PR #829, opened 2022-10-18
Add tests for `author.rs` (#700)

gallottino:
Hi guys!

Oniryu95 and me want to collaborate with you to develop onefetch. We started from this simple issue #700.
In future we wants to do something more consistent for this project.

Let us know!

Signed-off-by: gallottino <d.gallotti@outlook.it>
Co-authored-by: Oniryu95  <oniryu95@yahoo.it>

### DIFF
--- a/src/info/repo/author.rs
+++ b/src/info/repo/author.rs
@@ -135,4 +135,97 @@ mod test {
 
         assert_eq!(author.to_string(), "75% John Doe 1500");
     }
+
+    #[test]
+    fn test_authors_info_title_with_one_author() {
+        let author = Author::new(
+            "John Doe".into(),
+            "john.doe@email.com".into(),
+            1500,
+            2000,
+            true,
+        );
+
+        let authors_info = AuthorsInfo {
+            info_color: DynColors::Rgb(255, 0, 0),
+            authors: vec![author],
+        };
+
+        assert_eq!(authors_info.title(), "Author");
+    }
+
+    #[test]
+    fn test_authors_info_title_with_two_authors() {
+        let author = Author::new(
+            "John Doe".into(),
+            "john.doe@email.com".into(),
+            1500,
+            2000,
+            true,
+        );
+
+        let author_2 = Author::new(
+            "Roberto Berto".into(),
+            "bertolone2000@email.com".into(),
+            240,
+            300,
+            false,
+        );
+
+        let authors_info = AuthorsInfo {
+            info_color: DynColors::Rgb(255, 0, 0),
+            authors: vec![author, author_2],
+        };
+
+        assert_eq!(authors_info.title(), "Authors");
+    }
+
+    #[test]
+    fn test_author_info_value_with_one_author() {
+        let author = Author::new(
+            "John Doe".into(),
+            "john.doe@email.com".into(),
+            1500,
+            2000,
+            true,
+        );
+
+        let authors_info = AuthorsInfo {
+            info_color: DynColors::Rgb(255, 0, 0),
+            authors: vec![author],
+        };
+
+        assert!(authors_info
+            .value()
+            .contains("75% John Doe <john.doe@email.com> 1500"));
+    }
+
+    #[test]
+    fn test_author_info_value_with_two_authors() {
+        let author = Author::new(
+            "John Doe".into(),
+            "john.doe@email.com".into(),
+            1500,
+            2000,
+            true,
+        );
+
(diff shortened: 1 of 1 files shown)


src/info/repo/author.rs
```diff
@@ -135,4 +135,80 @@ mod test {
...
+        };
+
+        let authors_info_2 = AuthorsInfo {
+            info_color: DynColors::Rgb(255, 0, 0),
+            authors: vec![author2, author],
+        };
+
+        assert_eq!(authors_info.title(), "Author");
+        assert_eq!(authors_info_2.title(), "Authors");
+    }
```
spenserblack, line 169: If you're testing `title()`, you might want to make it 2 tests -- one like `test_author_info_singular_title` and another like `test_author_info_plural_title`.

src/info/repo/author.rs
```diff
@@ -135,4 +135,80 @@ mod test {
...
+            info_color: DynColors::Rgb(255, 0, 0),
+            authors: vec![author.clone()],
+        };
+        let authors_info_2 = AuthorsInfo {
+            info_color: DynColors::Rgb(255, 0, 0),
+            authors: vec![author, author2],
+        };
+
+        println!("{}", authors_info);
+        println!("{}", authors_info_2);
```
spenserblack, line 199:
```suggestion
```
Let's keep these print statements out of the tests.

src/info/repo/author.rs
```diff
@@ -135,4 +135,80 @@ mod test {
...
+
+        assert_eq!(
+            authors_info.value(),
+            "\u{1b}[38;2;255;0;0m75% John Doe <john.doe@email.com> 1500\u{1b}[39m"
+        );
```
spenserblack, line 204: See comments like https://github.com/o2sh/onefetch/pull/827#discussion_r996912875

Test that the expected *plain text* is there with `.contains` instead of including the color formatting in the expected value.

src/info/repo/author.rs
```diff
@@ -135,4 +135,80 @@ mod test {
...
+
+        let pad = authors_info_2.title().len() + 2;
+        let mut pad_str = String::from("");
+        let perc = (240 as f32 * 100. / 300 as f32).round() as usize;
+        for i in 0..pad {
+            pad_str.push_str(" ");
+        }
+        assert_eq!(authors_info_2.value(), format!("\u{1b}[38;2;255;0;0m75% John Doe <john.doe@email.com> 1500\u{1b}[39m\n{}\u{1b}[38;2;255;0;0m{}% Roberto Berto 240\u{1b}[39m", pad_str, perc))
```
spenserblack, line 212: Also use `.contains` here.

This test would be more readable if you just hardcoded the expected percentage, too. Because you control the factors (author's name, etc.) it's OK, even preferable, to hardcode the expected values.

So just make `perc` a string and test that `{perc}%` is contained.

spenserblack (changes requested):
Appreciate your interest in making `onefetch` a well-tested tool! Here are some notes.

In general, try to split up the tests a bit.

## Example for writing tests

```rust
fn is_odd(n: u32) -> bool {
    if is_even(n) {
        unimplemented!();
    }
    true
}

// good tests
#[test]
fn test_is_odd_true_on_odd_number() {
    assert!(is_odd(1));
}
#[test]
fn test_is_odd_false_on_even_number() {
    assert!(!is_odd(2));
    // when this fails, the first thing we see is that is_odd
    // doesn't handle even numbers
}

// Not so good
#[test]
fn is_odd_works() {
    assert!(is_odd(1));
    assert!(!is_odd(2));
    // when this fails, the first thing we see is that is_odd
    // doesn't work, but we have to dig into the test results
    // to see what isn't working
}
```

gallottino:
Followed your suggestion :)

spenserblack (approved):
The tests are much more readable now, thank you!

src/info/repo/author.rs
o2sh: BTW, if you wanna to check the formatting without losing in readability - by adding a sequence of ANSI escape codes - you could just do :

```rust
  assert_eq!(
      authors_info.value(),
      "75% John Doe <john.doe@email.com> 1500"
          .color(DynColors::Rgb(255, 0, 0))
          .to_string()
  );

```

spenserblack: Great idea, thanks for bringing that up!

Issue #832, opened 2022-10-24
Fail when there is no remote in git repo


forieux:
When I execute onefetch in a local git repo where no remote is set I have a failure. onefetch is installed with cargo and updated to the last version

thread 'main' panicked at 'non-empty path', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/onefetch-2.13.1/src/info/repo/project.rs:31:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

spenserblack:
Thanks for raising this! Since onefetch is supposed to work without fetching any remote data, it makes sense to me that we should support repositories that don't even have a remote.

Perhaps as a reasonable fallback, we can use the name of the directory as the project's name when there is no remote URL.

TBH I was a bit surprised that the error occurred [here](https://github.com/o2sh/onefetch/blob/3a3b1c6c04907d5461deb6ed609a2543f5be3776/src/info/repo/project.rs#L31) instead of somewhere before.

Digging into the code, it's because [we just use an empty string when there are no remotes](https://github.com/o2sh/onefetch/blob/3a3b1c6c04907d5461deb6ed609a2543f5be3776/src/info/repo/url.rs#L19). I think it makes sense to avoid converting `None` into `""`, though some refactoring may be in order to achieve that.

I guess this counts as an enhancement since we don't have any intended behavior for repos without remotes as far as I remember :shrug:

o2sh:
This seems to be a regression

v2.12.0
![image](https://user-images.githubusercontent.com/13710835/197625447-45c99a15-b23c-4433-94de-e4bc44995990.png)

v2.13.1
![image](https://user-images.githubusercontent.com/13710835/197625578-69ff5ecc-e755-4731-aed2-7cd394549c6b.png)

spenserblack:
Thanks for pointing that out, @o2sh! Didn't realize we had previously supported remoteless repositories. Updated labels to match.

o2sh:
Should be fixed in the [latest release](https://github.com/o2sh/onefetch/releases/tag/v2.13.2)

o2sh:
>Perhaps as a reasonable fallback, we can use the name of the directory as the project's name when there is no remote URL.

>I think it makes sense to avoid converting `None` into `""`, though some refactoring may be in order to achieve that to `None` an actual expected possible value without it being an error).

Do we keep this open to implement @spenserblack's suggestion?

o2sh:
[...]

PR #841, opened 2022-10-30
Repo without remote should not fail

o2sh:
Fix for #832


Byron (approved):
Thanks for checking in with me, all looks good. I always like seeing `git-testtools` coming to good use here as well.

PR #851, opened 2022-11-05
Add manifest crate

o2sh:
As a follow up to https://github.com/o2sh/onefetch/discussions/845, this PR adds a new library to onefetch `manifest` that detects and parses package manifests, only `Cargo.toml` and `package.json` are supported for now.

As a result:

- New info line -> Description
- Fallback to manifest entries for:version number, repo's name and license -> if not found with gitoxide or askalono, we display the data extracted from the manifest (if present)
- Remove `DependencyDetector` now part of `manifest` lib.

Left to do:

- [x] Add tests
- [x] Handle long description


manifest/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+use npm_package_json::Package;
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::{ffi::OsStr, fs};
+use strum::{Display, EnumIter};
+
+#[macro_use]
+extern crate lazy_static;
+
+lazy_static! {
```
Byron, line 11: I'd replace this with a function and a `match` on `&str`. 

If it really has to be a static map, I think [`OnceCell::sync::Lazy`](https://docs.rs/once_cell/1.16.0/once_cell/#lazy-initialized-global-data) is the standard (now).

manifest/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+                ManifestType::Cargo => parse_cargo_manifest(manifest_path),
+                ManifestType::Npm => parse_npm_manifest(manifest_path),
+            }
+        })
+        .filter_map(std::result::Result::ok)
+        .collect::<Vec<_>>();
+    Ok(manifests)
+}
+
+fn get_manifest_paths<P: AsRef<Path>>(path: P) -> Result<Vec<PathBuf>> {
```
Byron, line 56: It seems like this function can be combined with `get_manifests()` to do all work in one swoop. Generally, `contains_key` checks can be avoided in favour of `map.get()` which yields an `Option<V>`.

manifest/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    Ok(Manifest {
+        manifest_type: ManifestType::Cargo,
+        number_of_dependencies: m.dependencies.len(),
+        name: package.name.clone(),
+        description,
+    })
+}
+
+fn parse_npm_manifest(path: &PathBuf) -> Result<Manifest> {
+    let package = Package::from_path(path)?;
```
Byron, line 90: I'd not import `Package` here but refer to it by full path, similar to what's done above with `cargo_toml::Manifest`, for consistency and readability.

src/info/repo/dependencies.rs
```diff
@@ -0,0 +1,35 @@
...
+
+pub struct DependenciesInfo {
+    pub dependencies: String,
+}
+
+impl DependenciesInfo {
+    pub fn new(manifest: &Option<Manifest>) -> Self {
+        let dependencies = match manifest {
+            Some(m) => {
+                if m.number_of_dependencies == 0 {
```
Byron, line 12: As a fan of `<bool>.then()` I'd do: `(m.number_of_dependencies != 0).then(|| format!(…)).unwrap_or_default()`. Combined with the above I'd get `manifest.and_then(|m| (m.number_of_dependencies != 0).then(|| format!(…))).unwrap_or_default()`.

src/info/repo/dependencies.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl InfoField for DependenciesInfo {
+    const TYPE: InfoType = InfoType::Dependencies;
+
+    fn value(&self) -> String {
+        self.dependencies.clone()
+    }
+
+    fn title(&self) -> String {
+        String::from("Dependencies")
```
Byron, line 33: I'd do `"Dependencies".into()`

src/info/repo/project.rs
```diff
@@ -12,8 +13,8 @@ pub struct ProjectInfo {
 }
 
 impl ProjectInfo {
-    pub fn new(repo: &Repository, repo_url: &str) -> Result<Self> {
-        let repo_name = get_repo_name(repo_url)?.unwrap_or_default();
+    pub fn new(repo: &Repository, repo_url: &str, manifest: &Option<Manifest>) -> Result<Self> {
```
Byron, line 16: `&Option<T>` should be avoided in favour of `Option<&Manifest>`, which is way more usable. The caller owning `Option<T>` can do `my_option.as_ref()` to pass it along.

Byron (commented):
I think I have been auto-selected for review by GitHub, but thought I'd leave some comments about parts that I think can be more idiomatic, which usually means 'objectively preferable', but maybe not always 😅. That said, I hope it's useful, otherwise, please ignore.

src/info/repo/dependencies.rs
o2sh: Love it, 1ec711e to apply your suggestion everywhere for consistency.

src/info/repo/project.rs
o2sh: done, thanks

manifest/src/lib.rs
o2sh: You're right, done

manifest/src/lib.rs
o2sh: a function it is 👍

manifest/src/lib.rs
spenserblack: A separate `get_manifest_paths` function might be helpful to others that want to use this manifest crate as a dependency.

manifest/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+authors = ["o2sh <ossama-hjaji@live.fr>"]
+name = "manifest"
```
spenserblack, line 3: Looks like [manifest is already taken](https://crates.io/crates/manifest). Perhaps rename this to `onefetch-manifest`? If this sub-crate can't be published to crates.io, I don't think that onefetch itself can be published to crates.io. Also, with multiple crates in the same repo, should this project be converted to a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)?

o2sh: >Looks like [manifest is already taken](https://crates.io/crates/manifest). Perhaps rename this to onefetch-manifest?

🤦 Good catch, thanks @spenserblack 

> should this project be converted to a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)?

done, maybe we should do the same thing  to the `image_backends` module 🤔

spenserblack: > maybe we should do the same thing ... to the `image_backends` module

Thanks makes sense to me :+1:
onefetch probably has a lot of usage that can go into separate crates that others may want to use (kind of related: https://github.com/o2sh/onefetch/pull/851#discussion_r1016683417). I think this `onefetch-manifest` crate is a really cool tool that could evolve into its own project (I'm thinking of it like a tokei for your manifests).

manifest/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+authors = ["o2sh <ossama-hjaji@live.fr>"]
+name = "onefetch-manifest"
+version = "0.1.0"
```
spenserblack, line 4:
```suggestion
version = "2.13.2"
```
What do you think about keeping the version of `onefetch-remote` and `onefetch` the same?

This is a pattern I see in a lot of NPM workspaces/monorepos (jest, typescript-eslint, playwright). I think the gains are
- only one release/changelog is needed for the whole workspace
- it's honestly just easier than managing semver for each individual crate
- we signal that this sub-crate is tied to the main crate, and, while the library is publicly available, it is not stable independently of onefetch itself

Although, a con is that, as a user of some of these monorepos, it can be annoying when you get alerted of a new version yet nothing has changed in the package you're using.

manifest/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+        "package.json" => Some(ManifestType::Npm),
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn test_get_manifest() {}
```
spenserblack, line 80: Is this a to do?

src/info/repo/description.rs
```diff
@@ -21,7 +24,24 @@ impl InfoField for DescriptionInfo {
 
     fn value(&self) -> String {
         match &self.description {
-            Some(v) => v.into(),
+            Some(value) => {
+                if !value.is_empty() && value.contains(' ') {
```
spenserblack, line 28:
```suggestion
                if value.contains(' ') {
```
Just because I'm obsessed with micro-optimizations :laughing:
I'm pretty sure a string would never be empty if it contains `' '`.

o2sh: 🤦

manifest/Cargo.toml
o2sh: if keeping the same version throughout the workspace means less maintenance on our part,  we should go for it.
Thanks @spenserblack 👍

manifest/src/lib.rs
o2sh: I've just added some integration tests for the `manifest` crate

[01ec7a0](https://github.com/o2sh/onefetch/pull/851/commits/01ec7a0b75ad67ed85ba52d1bd525194f6debc01)
[55b0228](https://github.com/o2sh/onefetch/pull/851/commits/55b022877609c3dd3a1022ffb2dbf288372f39c3)

PR #856, opened 2022-11-10
upgrade gitoxide

Byron:
Standard maintenance, which adjust for a breaking change in cargo features of `git-repository`.


o2sh:
Thanks for the heads up @Byron, and for running `clippy` 👍

PR #857, opened 2022-11-12
Add devcontainer/codespace config

spenserblack:
This should allow contributors to use GitHub codespaces to contribute to onefetch.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,9 @@
+{
+  "name": "Rust",
+  "build": {
+    "dockerfile": "Dockerfile",
+    "args": { "VARIANT": "buster" }
+  },
+  "extensions": ["EditorConfig.EditorConfig"],
+  "remoteUser": "vscode"
+}
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
 # NOTE: see https://hub.docker.com/_/microsoft-vscode-devcontainers
 ARG VARIANT="buster"
 FROM mcr.microsoft.com/vscode/devcontainers/rust:1-${VARIANT}
-RUN sudo apt update && sudo apt install cmake
+RUN sudo apt update && sudo apt install -y cmake
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,3 +1,4 @@
 # NOTE: see https://hub.docker.com/_/microsoft-vscode-devcontainers
 ARG VARIANT="buster"
 FROM mcr.microsoft.com/vscode/devcontainers/rust:1-${VARIANT}
+RUN apt install cmake
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,2 +1,3 @@
+# NOTE: see https://hub.docker.com/_/microsoft-vscode-devcontainers
 ARG VARIANT="buster"
 FROM mcr.microsoft.com/vscode/devcontainers/rust:1-${VARIANT}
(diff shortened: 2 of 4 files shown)


o2sh:
Thanks @spenserblack,

As a next step, we may need to add a few things to have fully self-sufficient dev container template for onefetch, I'm thinking of the tooling for [onefetch.dev](https://onefetch.dev): node, npm, svelte extensions...

spenserblack:
> We may need to add a few things to make it a more "self-sufficient" dev container, I'm thinking of the tooling for [onefetch.dev](https://onefetch.dev) -> node, npm, svelte extensions...

Great idea! I still have a lot more to learn, but I think it's possible to have multiple dev container configurations per repo. So potentially one can have a codespace for onefetch, and one for onefetch.dev.

spenserblack:
Oh, BTW, might want to check out the [prebuild setting](https://github.com/o2sh/onefetch/settings/codespaces).

More reading: https://docs.github.com/en/codespaces/prebuilding-your-codespaces

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "Rust",
+  "build": {
+    "dockerfile": "Dockerfile",
+    "args": { "VARIANT": "buster" }
```
o2sh, line 5: Is this line necessary?  It seems like it is already set via an `ARG` instruction in the `Dockerfile`.

o2sh:
>So potentially one can have a codespace for onefetch, and one for onefetch.dev.

That's right, I'll create a  separate dev container config under `docs/vercel` with the right tooling.

>Oh, BTW, might want to check out the [prebuild setting](https://github.com/o2sh/onefetch/settings/codespaces).

I'll definitely set that up 👍

spenserblack:
Does the simplified devcontainer (https://github.com/o2sh/onefetch/commit/50bb474a98a41f4b0b601792af4f87fc2a9d08ed) work for you? For me it fails on libz-sys

```
...
--- stderr
  thread 'main' panicked at '
  failed to execute command: No such file or directory (os error 2)
  is `cmake` not installed?
```

o2sh:
Yes, I'm really sorry about that, I should have tested it before pushing it --> I just reverted it 7cc7db5

BTW, another issue when running `cargo run`  
```
Error: config value 'safe.directory' was not found; class=Config (7); code=NotFound (-3)
```
We may need to add "git config --global --add safe.directory /workspaces/onefetch" to the Dockerfile

spenserblack:
> We may need to add `git config --global --add safe.directory /workspaces/onefetch` to the Dockerfile

I think that's probably best put in a `postCreate` command. Probably don't need an image layer for a config change.
```jsonc
// .devcontainer/devcontainer.json
// ...
"postCreateCommand": ".devcontainer/postCreate.sh",
// ...
```
```bash
#!/bin/bash
# .devcontainer/postCreate.sh
git config --global --add safe.directory $PWD
```

>I should have tested it before pushing it

I didn't know this at first (which is why I was pushing everything to my fork's `main` branch), but you can create codespaces from configs on other branches. You just need to have switched to that branch in the web UI before creating the new workspace. Seems so simple in hindsight :laughing:

o2sh:
Thanks for the help, I think  we can skip the shell script and put the command directly in the .json file c9e8223

spenserblack:
FYI: https://github.com/orgs/community/discussions/39697

One thing I noticed is that prebuilds count against your storage. Perhaps you want to disable and delete prebuilds (knowing that they're billable, they only seem worth it for orgs TBH :shrug:)

o2sh:
>One thing I noticed is that prebuilds count against your storage.

I guess that explains all the notifications I was getting from GH telling me to increase my Codespaces storage usage 😅  

I'll delete them, thanks for letting me know @spenserblack

PR #866, opened 2022-11-18
Fix secondary devcontainer setup

spenserblack:
At least from the docs I've read, alternate dev container configs must be in `/.devcontainer/SUBFOLDER/`.

To use, when creating a codespace, instead of clicking `+`, they can click `...` and then "New with options".

There's still a bit more that can be done (I think it would be reasonable for `npm run dev` to be a postAttachCommand), but I wasn't sure if that should be a different PR or part of this one 🙂

### DIFF
--- a/.devcontainer/website/devcontainer.json
+++ b/.devcontainer/website/devcontainer.json
@@ -2,7 +2,6 @@
   "name": "Node.js & TypeScript",
   "image": "mcr.microsoft.com/devcontainers/typescript-node:16-bullseye",
   "workspaceFolder": "/workspaces/onefetch/docs/vercel",
-  "workspaceMount": "/",
   "extensions": ["esbenp.prettier-vscode", "svelte.svelte-vscode"],
   "postCreateCommand": "npm ci",
   "settings": {
--- a/.devcontainer/website/devcontainer.json
+++ b/.devcontainer/website/devcontainer.json
@@ -1,6 +1,8 @@
 {
   "name": "Node.js & TypeScript",
   "image": "mcr.microsoft.com/devcontainers/typescript-node:16-bullseye",
+  "workspaceFolder": "/workspaces/onefetch/docs/vercel",
+  "workspaceMount": "/",
   "extensions": ["esbenp.prettier-vscode", "svelte.svelte-vscode"],
   "postCreateCommand": "npm ci",
   "settings": {
--- a/.devcontainer/website/devcontainer.json
+++ b/.devcontainer/website/devcontainer.json
@@ -1,6 +1,7 @@
 {
   "name": "Node.js & TypeScript",
   "image": "mcr.microsoft.com/devcontainers/typescript-node:16-bullseye",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/onefetch,type=bind,consistency=cached",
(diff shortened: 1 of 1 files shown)


o2sh:
>alternate dev container configs must be in /.devcontainer/SUBFOLDER/.

Should have known that, sorry 😓 

>To use, when creating a codespace, instead of clicking +, users can click ... and then "New with options".

I'll make sure to setup the prebuild action for that config.

>There's still a bit more that can be done (I think it would be reasonable for npm run dev to be a postAttachCommand), but I wasn't sure if that should be a different PR or part of this one 🙂

Up to you, `npm run dev` does seem convenient especially with the port forwarding already in place 👍

spenserblack:
> Up to you, `npm run dev` does seem convenient especially with the port forwarding already in place

Done :+1:

.devcontainer/website/devcontainer.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Node.js & TypeScript",
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:16-bullseye",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/onefetch,type=bind,consistency=cached",
+  "workspaceFolder": "/workspaces/onefetch/docs/vercel",
+  "extensions": ["esbenp.prettier-vscode", "svelte.svelte-vscode"],
+  "postCreateCommand": "npm ci",
+  "postAttachCommand": "npm run dev -- --port 3000",
```
o2sh, line 8: why not merge the two into : `"postCreateCommand": "npm ci && npm run dev -- --port 3000"`?

.devcontainer/website/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+  "settings": {
+    "editor.formatOnSave": true
+  },
+  "portsAttributes": {
+    "3000": {
+      "label": "onfetch.dev",
+      "onAutoForward": "notify"
+    }
+  }
```
o2sh, line 17: Why not keep the default port 5173?

.devcontainer/website/devcontainer.json
spenserblack: Post create only happens once, when first creating the codespace, while post attach runs each time you connect to the codespace. This way you can stop the codespace and have the dev server run again next time you connect without needing to rebuild the container.

.devcontainer/website/devcontainer.json
spenserblack: I've been surprised before by changes to the default port, so this is just to keep things a bit more consistent.

o2sh:
There seems to be an issue when using prebuilds with this config, it completely ignores the `workspaceMount` and `workspaceFolder` setting which makes the `npm run dev` command fail as it is ran  in the root folder instead of onefetch/docs/vercel.

I deleted the prebuild setup for now.

spenserblack:
>I deleted the prebuild setup for now.

That makes sense. Codespaces initially didn't support `workspaceFolder`/`workspaceMount` *at all,* so I guess now it's just partial support :shrug:

It's no big deal if there isn't a prebuild, it just means that initial container creation might take a bit more time. We aren't using any complex docker setup or anything, either, so I don't think there's much of a gain.

PR #869, opened 2022-11-20
Add README translation for Korean

abiriadev:
First of all, thanks for creating this helpful tool :)

I thought it would be more useful for Korean developers to have Korean README, so I added it.

Unfortunately, I have no idea how to add a flag SVG for the Korean language. should I draw it myself, or there is some source where I can find one?

And is there any point left that I missed updating? (I am not so good at English)

Thank you in advance :)


o2sh:
I've just added the Korean flag to your PR, everything should be good now.

Thanks a lot for your contribution @abiriadev, I'm sure this will be very helpful for Korean developers 👍

PR #881, opened 2022-11-29
Yet another cloud development configuration

spenserblack:
gitpod.io is very similar codespaces, but with different pricing models. [It has some interesting GitHub-specific configuration options](https://www.gitpod.io/docs/references/gitpod-yml#github), but I'll leave that up to you if you want to configure those or not (I think they require the gitpod GitHub app).

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -12,4 +12,4 @@ ports:
   - name: onefetch.dev
     description: Web preview of onefetch ASCII art
     port: 3000
-    onOpen: notify
+    onOpen: open-preview
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,6 +1,9 @@
 vscode:
   extensions:
     - 'EditorConfig.EditorConfig'
+    - 'svelte.svelte-vscode'
+    - 'rust-lang.rust-analyzer'
+    - 'tamasfe.even-better-toml'
 tasks:
   - name: Build Onefetch
     before: sudo apt install -y cmake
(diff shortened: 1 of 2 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,15 @@
+vscode:
+  extensions:
+    - 'EditorConfig.EditorConfig'
```
o2sh, line 3:
```suggestion
    - 'EditorConfig.EditorConfig'
    - 'svelte.svelte-vscode'
    - 'rust-lang.rust-analyzer'
    - 'tamasfe.even-better-toml'
```

spenserblack:
:thinking: Maybe the workflows should ignore this file :sweat_smile:

o2sh (approved):
It's indeed a nice alternative to GH codespaces with some cool features (more IDE options, prebuilds...). 

I wanted to enable the prebuilds but I got scared by the level of access rights requested 😩

In the meantime, this PR can be merged.

Thanks a lot @spenserblack, this sure will improve the developer experience.

spenserblack:
> I wanted to enable the prebuilds but I got scared by the level of access rights requested 😩

I don't blame you! Better to be on the safe side 🙁 

Anyway, gitpod usage is pretty simple: just paste `gitpod.io#/` in front of your repo's URL. So contributing instructions could be
>Fork this repository, then go to `gitpod.io#/https://github.com/<YOUR-USERNAME>/onefetch`

(This works for PR URLs, too, BTW 🙂 )

Issue #706, opened 2022-07-25
Format numbers


H4ckerxx44:
### Summary 💡

On larger code bases, the amount of `files` and the amount of `lines of code` could become a bit unreadable due to the lacking formatting, my feature request would be to format the `lines of code`, `commits` as well as the `files` datapoint like, `15614` -> `15,614` or `3255204307` -> `3,255,204,307`

### Motivation 🔦

That's 3 billlion lines of code, quite unreadable
![grafik](https://user-images.githubusercontent.com/56711385/180836532-5103146d-88e4-4c77-8279-2236934efaf0.png)

H4ckerxx44:
> Just throwing out another option: what would you think about formats like `3K` or `3.3M` instead? Similar to the Size field.

While I have nothing against that suggestion, I would not choose this due to it stripping away some "exactness" of the numbers at hand. (Let's ignore the Size in GiB)

o2sh:
This feature would require to handle multiple formatting options (e.g. which thousands separator to use) and be able to detect the system local of the user (or use a cli flag to ask for the user's preference), [`num_format`](https://github.com/bcmyers/num-format) seems like a good candidate. Unfortunately the library doesn't seem to be maintained https://github.com/bcmyers/num-format/issues/27

spenserblack:
TBH I'm not sure how necessary exactness is. This is just my opinion, but once billions of lines of code are reached, I wouldn't care if it's 3 billion, 3 billion and 1 hundred, or 3 billion and 1 thousand, I'd just care about the 3 or 4 most significant digits. If I'm showing off my repo, I'd be saying "look, over 3 billion lines of code!" not "look, 3 billion, 255 million, ..." :laughing:
Of course, for other outputs, like JSON, the exact number would be desired, but we wouldn't need to format it.

As @o2sh different locales have different thousands separators. [And even the placement of the separators can differ](https://en.wikipedia.org/wiki/Indian_numbering_system) (not all cultures historically group numbers into thousands), so any formatting with separators would need to come from a library.

o2sh:
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

H4ckerxx44:
Well, the activity depends on what you ultimately decide on for this idea.
I would just - as said previously - go with the `123,489,645 lines` formatting of those numbers.

spenserblack:
[...]

PR #892, opened 2022-12-11
Add `--number-separator` CLI flag

o2sh:
refers to #706 

image
image

Impacted by the CLI flag:

* Number of branches and number of tags
* Number of dependencies
* Number of commits per author
* Number of commits
* LOC
* Number of files

### DIFF
--- a/tests/snapshots/repo__repo.snap
+++ b/tests/snapshots/repo__repo.snap
@@ -7,8 +7,8 @@ expression: info
   "gitVersion": "git version",
   "project": {
     "repo_name": "repo",
-    "number_of_branches": 0,
-    "number_of_tags": 1
+    "number_of_branches": "0",
+    "number_of_tags": "1"
   },
   "description": "Amazing tool",
   "head": {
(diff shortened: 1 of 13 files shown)


spenserblack:
Some things to keep in mind:
- [Not all separators are at the thousands place](https://en.wikipedia.org/wiki/Indian_numbering_system#Use_of_separators), so we may get a request to support other groupings, later.
- There are some numerical systems that look like `1.234,5` instead of `1,234.5`.

So, if `--format-numbers` defines the thousands separator, it can be a breaking change if we need to rename that arg to accommodate other options. We could
- rename `--format-numbers` to `--number-separator` or something, or
- change `--format-numbers` totake *names* of formats (we might want to look into what other tools call various number formats).

We could even take a number format like Excel does, `#,###.00`, so that users can define *any* placement of separators, and what types of separators to use, without having to raise future issues. That might be overengineering for our needs, though :laughing:

o2sh:
You're right, my first implementation actually took a localization as an input, but I reverted in 7c6cbba when I found out that [num_crate did not support floating numbers](https://github.com/bcmyers/num-format/issues/1) which would be misleading to users if the decimal character wasn't affected.

I think the `--number-separator` - renamed after your suggestion - CLI flag is a good first step and covers the original issue raised in #706.

tests/snapshots/repo__repo.snap
```diff
@@ -7,8 +7,8 @@ expression: info
   "gitVersion": "git version",
   "project": {
     "repo_name": "repo",
-    "number_of_branches": 0,
-    "number_of_tags": 1
+    "number_of_branches": "0",
+    "number_of_tags": "1"
```
spenserblack, line 11: I guess a change in JSON types makes this a breaking change :thinking:

spenserblack (approved):

tests/snapshots/repo__repo.snap
spenserblack: Thinking about this some more, maybe the format should only affect how the stats are *printed*, but keep the struct's types the same, so data outputs don't change their type.

tests/snapshots/repo__repo.snap
```diff
@@ -49,7 +49,7 @@ expression: info
   "lastChange": "22 years ago",
   "contributors": 4,
   "repoUrl": "https://github.com/user/repo.git",
-  "numberOfCommits": "4",
+  "numberOfCommits": 4,
```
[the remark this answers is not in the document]
spenserblack: Yeah, that makes sense to me, too.

spenserblack (approved):

PR #907, opened 2022-12-27
Bump git-testtools

### LINKED ISSUE #906: Running `cargo test` can result in a prompt for a tag message that blocks tests
### Duplicates

### Current behavior 😯

When running `cargo test` in a codespace, the test stops and I'm prompted to write a tag message in my editor.

### Expected behavior 🤔

Tests don't take any input from the user during runtime.

### Steps to reproduce 🕹

This might be because I set `git config tag.gpgSign true` in my dotfiles when `CODESPACES=true`.

spenserblack:
This prevents the test from potentially prompting for a tag message when generating the repo fixture, depending on the global `tag.gpgSign` config setting.

Resolves #906

******

This config setting could have been added to the other fixtures, but I opted for the smallest possible change to fix #906.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,15 +101,6 @@ dependencies = [
 
 [[package]]
 name = "atoi"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7c57d12312ff59c811c0643f4d80830505833c9ffaebd193d819392b265be8e"
-dependencies = [
- "num-traits",
-]
-
-[[package]]
-name = "atoi"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f28d99ec8bfea296261ca1af174f24225171fea9664ba9003cbebee704810528"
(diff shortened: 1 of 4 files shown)


spenserblack:
@Byron Could you confirm if this is acceptable usage? Not really sure why the other config settings used `--add` in the fixture, but it didn't seem necessary.

Byron (changes requested):
That was a good wakeup call to [finally release a new version of `git-testtools`](https://github.com/Byron/gitoxide/blob/adjustments-for-cargo/tests/tools/CHANGELOG.md#0100-2022-12-28) which fixes this issue automatically.

```patch
diff --git a/Cargo.toml b/Cargo.toml
index 70488e9..4bc10b8 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ tokei = "12.1.2"
 yaml-rust = "0.4.5"
 
 [dev-dependencies]
-git-testtools = "0.9.0"
+git-testtools = "0.10.0"
 insta = { version = "1.22.0", features = ["json", "redactions"] }
 pretty_assertions = "1.3.0"
 ```

PR #911, opened 2022-12-30
Info struct to holds a Vec<InfoField>

o2sh:
Part 2 of #755, 

Now that all info fields implement the `InfoField` trait, we can put them inside a Vector and iterate over to simplify the `fmt()` method of `std::fmt::Display`.

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -65,15 +65,17 @@ impl std::fmt::Display for Info {
         }
 
         //Info lines
-        for info_field in self.info_fields.iter() {
-            if let Some(info_field_value) = info_field.get(&self.disabled_fields) {
-                self.write_styled_info_line(
-                    f,
-                    &info_field.title(),
-                    &info_field_value,
-                    info_field.should_color(),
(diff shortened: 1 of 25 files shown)


src/info/mod.rs
```diff
@@ -73,93 +65,15 @@ impl std::fmt::Display for Info {
...
-            self.write_styled_info_line(f, &self.license.title(), &license_info_value, true)?;
+        for info_field in self.info_fields.iter() {
+            if let Some(info_field_value) = info_field.get(&self.disabled_fields) {
+                self.write_styled_info_line(
+                    f,
+                    &info_field.title(),
+                    &info_field_value,
+                    info_field.should_color(),
+                )?;
+            }
```
spenserblack, line 76: Just a nitpick to consider: `.filter_map` could probably be used to avoid an indentation level

spenserblack (approved):
This is awesome! It looks like it will be much easier to maintain stat lines, now :rocket:

spenserblack (approved):

PR #922, opened 2023-01-07
feat(vercel): add section links

ozwaldorf:
# What

- Adds section links to the vercel website 
- didn't see a npm script for running prettier, so I added one

# Demo

Jumping via section link https://vercel-hw8vz6ref-o2sh.vercel.app/#Makefile

image

`npm run format`:

image

### DIFF
--- a/docs/vercel/public/sakura.css
+++ b/docs/vercel/public/sakura.css
@@ -99,6 +99,12 @@ a:hover {
   border-bottom: 2px solid #222222;
 }
 
+h3 a,
+h3 a:hover,
+h3 a:visited {
+  color: #222222;
+}
+
 ul {
   padding-left: 1.4em;
   margin-top: 0px;
--- a/docs/vercel/public/sakura.css
+++ b/docs/vercel/public/sakura.css
@@ -99,12 +99,6 @@ a:hover {
   border-bottom: 2px solid #222222;
 }
 
-h3 a,
-h3 a:hover,
-h3 a:visited {
-  color: #222222;
-}
-
 ul {
   padding-left: 1.4em;
   margin-top: 0px;
--- a/docs/vercel/src/lib/AsciiPreview.svelte
+++ b/docs/vercel/src/lib/AsciiPreview.svelte
@@ -74,7 +74,7 @@
   .title-link,
   .title-link:hover,
   .title-link:visited {
-      color: #222222;
+    color: #222222;
   }
 
   pre {
--- a/docs/vercel/src/lib/AsciiPreview.svelte
+++ b/docs/vercel/src/lib/AsciiPreview.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import Chip from './Chip.svelte';
+  import TitleLink from './TitleLink.svelte';
 
   export let name: string;
   export let ansi: string[];
@@ -35,7 +36,7 @@
 <div class="title-row">
   <div class="language-name">
     <Chip id={name} color={chip} width={24} height={24} />
-    <h3 id={name}><a href="#{name}" class="title-link">{name}</a></h3>
+    <TitleLink {name} />
   </div>
   <div class="checkbox">
     <input id="dark-checkbox-{name}" type="checkbox" bind:checked={dark} />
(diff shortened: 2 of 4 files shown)


docs/vercel/public/sakura.css
```diff
@@ -99,6 +99,12 @@ a:hover {
   border-bottom: 2px solid #222222;
 }
 
+h3 a,
+h3 a:hover,
+h3 a:visited {
+  color: #222222;
+}
+
```
spenserblack, line 107: I think any custom CSS should go in a separate file, so that there won't be any conflict to resolve if we decide to update `sakura.css`. Also, maybe this should be a class instead of set on the HTML tags, so we won't have to update the CSS if we ever change the HTML element (e.g. `h3` -> `h4`).

docs/vercel/src/lib/AsciiPreview.svelte
```diff
@@ -35,7 +35,7 @@
 <div class="title-row">
   <div class="language-name">
     <Chip id={name} color={chip} width={24} height={24} />
-    <h3>{name}</h3>
+    <h3 id={name}><a href="#{name}">{name}</a></h3>
```
spenserblack, line 38:
```suggestion
    <h3 id={name}><a class="lang-link" href="#{name}">{name}</a></h3>
```
Per other comment.

spenserblack (changes requested):
Great idea!
I'm tempted to attempt to auto-generate PR comments with links to the affected languages, but I think that would be a case of over-engineering :rofl:

Anyway, some minor nitpicks about the CSS and HTML.

ozwaldorf:
I also thought about autogenerating links too, it'd be a great feature for the future :smile:

docs/vercel/src/lib/AsciiPreview.svelte
ozwaldorf: https://github.com/o2sh/onefetch/pull/922/commits/8b0dd3b25d8418150667b918ea1e1a0e78269ef6

docs/vercel/public/sakura.css
ozwaldorf: https://github.com/o2sh/onefetch/pull/922/commits/8b0dd3b25d8418150667b918ea1e1a0e78269ef6

docs/vercel/src/lib/AsciiPreview.svelte
```diff
@@ -35,7 +35,7 @@
 <div class="title-row">
   <div class="language-name">
     <Chip id={name} color={chip} width={24} height={24} />
-    <h3>{name}</h3>
+    <h3 id={name}><a href="#{name}" class="title-link">{name}</a></h3>
```
spenserblack, line 38: Now that I think about it, a title link could probably be a component.
```svelte
<!-- TitleLink.svelte -->
<script lang="ts">
export let name: string;
</script>

<h3 id={name}<a href="#{name}" class="title-link">{name}</a></h3>

<style>
.title-link,
.title-link:hover,
.title-link:visited {
  color: #222;
}
</script>
```
```svelte
<!-- AsciiPreview.svelte -->
<!-- ... -->
<TitleLink {name} />
```

spenserblack (approved):
Left one thought to consider, but I don't consider it as something that would prevent my approval. LGTM!

docs/vercel/src/lib/AsciiPreview.svelte
ozwaldorf: Tbh, it's probably overthinking to separate it :pray:

spenserblack: You're probably right. I'm a bit over-obsessed with how git diffs potentially look :laughing: I think the only gain I was thinking of is that, if updating this title (e.g. slugifying the `id`) the diff would be prettier if this was already a separate component.

So, yeah, as I said these changes look good to me. That suggestion is just a matter of opinion, and I'll leave it up to you if you want to implement it or not :slightly_smiling_face:

ozwaldorf: https://github.com/o2sh/onefetch/pull/922/commits/ca5b1ba94af8c034ba5d018bd412070a59e51e76 :smile:

spenserblack (approved):

Issue #931, opened 2023-01-17
Standalone ascii parser crate


ozwaldorf:
### Summary 💡

Similar to `onefetch_image`, it'd be awesome to also release an additional crate `onefetch_ascii` that contains the template logic for the ascii 

### Motivation 🔦

https://github.com/ozwaldorf/punfetch :slightly_smiling_face:

spenserblack:
`onefetch` being a library that can facilitate the authoring of other `xyzfetch` tools is an interesting idea! I can see how a lot of the logic in `onefetch` could be used for other fetchers. [The ASCII art module](https://docs.rs/onefetch/2.14.2/onefetch/ui/ascii_art/index.html) is already public, so I'm not sure if it needs to be a separate crate right now. Biggest reason to make it a separate crate would probably be to allow a much smaller dependency tree for those who just want to access this module.

Otherwise, it might make more sense to just make the `ascii_art` module more usable for dependent crates.

o2sh:
Hi @ozwaldorf, 
with #934, `AsciiArt.rs` has now its own crate and will be released alongside onefetch in its next release.

Sorry for the lack of documentation. Please feel free to contribute if you believe it will assist others in using this crate.

Good luck with  [punfetch](https://github.com/ozwaldorf/punfetch)! 👍

PR #934, opened 2023-01-18
Turn `AsciiArt.rs` into its own crate

o2sh:
refers to #931

### DIFF
--- a/ascii/README.md
+++ b/ascii/README.md
@@ -1,0 +1,7 @@
+# ascii
+
+Provides the primary interface to display ascii art to the terminal.
+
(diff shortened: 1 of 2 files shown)


o2sh:
>Looks like ascii/README.md is an empty file. Should it be a symlink?

Sorry about that, I forgot to marked it as a draft 😓 

I added a very short readme. This crate is definitely lacking documentation - same applies for `image` and `manifest` BTW.

spenserblack (approved):
> Even though the code is quite clear, this crate is lacking documentation - same applies to `image` and `manifest` BTW.

That gives the community a chance to contribute :laughing:

PR #937, opened 2023-01-19
Fix Markdown / Jupyter markup not getting counted

### LINKED ISSUE #933: Markdown is ignored if it doesn't contain code blocks
### Duplicates

### Current behavior 😯

Running `onefetch -T prose` on a repo that contains Markdown results in
```console
Error: Could not find any source code in this repository
```

### Expected behavior 🤔

Markdown should be detected if prose is enabled, even if the Markdown doesn't contain code blocks.

### Steps to reproduce 🕹

````bash
TMPDIR=$(mktemp -d)
cd $TMPDIR
git init
git commit --allow-empty -m "init"  # just adding a commit for some stats
cat <<EOF > file.md
# hello, world

This is a file without code blocks
EOF
onefetch -T prose  # error

cat << "EOF" > file.md
```bash
chmod +x install.sh
./install.sh
```
EOF
onefetch -T prose  # renders Markdown ASCII
````

This looks similar to https://github.com/o2sh/onefetch/issues/627, but `-T prose` alone doesn't resolve this.

### Possible Solution 💡

The issue seems to be that tokei marks lines of Markdown as *comments* (then what is `<!-- -->`? :laughing:) except for code blocks and blank lines. IMO *all* lines of Markdown should be counted (even blank lines, since they can have significance in the rendered output).

Unfortunately, this is a bit hard to do since all languages are defined in a data file. Perhaps an optional field called something like `tokei-fields` can be used, which would allow one to list what types of lines should be used (`[Code, Comments, Blanks]`).

### PR DESCRIPTION
**Pull request by spenserblack**
This allows an optional definition of the line type for a language. If it is defined, it will sum them together. Defaults to using `.code`. By using this definition, comments are included in the type of lines of code to be summed for Markdown and Jupyter Notebooks.

This also changes the summing of child languages to be recursive, so that deeply nested code (e.g. Bash in Markdown in Jupyter) can be counted.

Fixes #933

### DIFF
--- a/languages.yaml
+++ b/languages.yaml
@@ -1311,7 +1311,6 @@ Jupyter:
       - "#9E9E9E"
     chip: "#DA5B0B"
   serialization: jupyter-notebooks
-  line_types: [code, comments]
 Kotlin:
   type: programming
   ascii: |
--- a/languages.yaml
+++ b/languages.yaml
@@ -1311,7 +1311,7 @@ Jupyter:
       - "#9E9E9E"
     chip: "#DA5B0B"
   serialization: jupyter-notebooks
-  tokei_line_types: [code, comments]
+  line_types: [code, comments]
 Kotlin:
   type: programming
   ascii: |
@@ -1484,7 +1484,7 @@ Markdown:
       - white
       - red
     chip: "#083FA1"
-  tokei_line_types: [code, comments]
+  line_types: [code, comments]
 Nim:
   type: programming
   ascii: |
--- a/languages.yaml
+++ b/languages.yaml
@@ -1311,6 +1311,7 @@ Jupyter:
       - "#9E9E9E"
     chip: "#DA5B0B"
   serialization: jupyter-notebooks
+  tokei_line_types: [code, comments]
 Kotlin:
   type: programming
   ascii: |
@@ -1483,6 +1484,7 @@ Markdown:
       - white
       - red
     chip: "#083FA1"
+  tokei_line_types: [code, comments]
 Nim:
   type: programming
   ascii: |
--- a/src/info/langs/language.tera
+++ b/src/info/langs/language.tera
@@ -5,7 +5,6 @@ use owo_colors::{
 use std::fmt;
 use std::fmt::Write;
 use strum::EnumIter;
-use tokei;
 
 pub struct Colors {
     basic_colors: Vec<DynColors>,
@@ -110,25 +109,18 @@ impl Language {
     }
 }
 
-/// Counts the lines-of-code of a tokei `Language`. Takes into
-/// account that a prose language's comments *are* its code.
-pub fn loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
-    let loc = match language_type {
+fn __loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
+    match language_type {
         {% for language, attrs in languages -%}
             {%- set line_types = attrs.line_types | default(value=['code']) -%}
             tokei::LanguageType::{{ language }} => language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %},
         {% endfor %}
         _ => unimplemented!("Language Type {:?}", language_type),
-    };
-    loc + language.children.iter().fold(0, |sum, (lang_type, reports)| {
-        sum + reports.iter().fold(0, |sum, report| sum + stats_loc(lang_type, &report.stats))
-    })
+    }
 }
 
-/// Counts the lines-of-code of a tokei `Report`. This is the child of a
-/// `tokei::CodeStats`.
-pub fn stats_loc(language_type: &tokei::LanguageType, stats: &tokei::CodeStats) -> usize {
-    let stats = stats.summarise();
+
+fn __stats_loc(language_type: &tokei::LanguageType, stats: &tokei::CodeStats) -> usize {
     match language_type {
         {% for language, attrs in languages -%}
             {%- set line_types = attrs.line_types | default(value=['code']) -%}
--- a/src/info/langs/language.tera
+++ b/src/info/langs/language.tera
@@ -115,7 +115,7 @@ impl Language {
 pub fn loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
     match language_type {
         {% for language, attrs in languages -%}
-            {%- set line_types = attrs.tokei_line_types | default(value=['code']) -%}
+            {%- set line_types = attrs.line_types | default(value=['code']) -%}
             tokei::LanguageType::{{ language }} => language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %},
         {% endfor %}
         _ => unimplemented!("Language Type {:?}", language_type),
--- a/src/info/langs/language.tera
+++ b/src/info/langs/language.tera
@@ -128,14 +128,14 @@ pub fn loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> u
 /// Counts the lines-of-code of a tokei `Report`. This is the child of a
 /// `tokei::CodeStats`.
 pub fn stats_loc(language_type: &tokei::LanguageType, stats: &tokei::CodeStats) -> usize {
-    let loc = match language_type {
+    let stats = stats.summarise();
+    match language_type {
         {% for language, attrs in languages -%}
             {%- set line_types = attrs.line_types | default(value=['code']) -%}
             tokei::LanguageType::{{ language }} => stats.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + stats.{{ line_type }}{% endfor %},
         {% endfor %}
         _ => unimplemented!("Language Type {:?}", language_type),
-    };
-    loc + stats.blobs.iter().fold(0, |sum, (lang_type, stats)| sum + stats_loc(lang_type, stats))
+    }
 }
 
 {% for language, attrs in languages -%}
--- a/src/info/langs/language.tera
+++ b/src/info/langs/language.tera
@@ -5,6 +5,7 @@ use owo_colors::{
 use std::fmt;
 use std::fmt::Write;
 use strum::EnumIter;
+use tokei;
 
 pub struct Colors {
     basic_colors: Vec<DynColors>,
@@ -106,6 +107,20 @@ impl Language {
                 Language::{{ language }} => Rgb({{ rgb.r }}, {{ rgb.g }}, {{ rgb.b }}),
             {% endfor %}
         }
+    }
+}
+
+/// Compiles the top-level lines-of-code of a tokei `Language`. Takes into
+/// account that a prose language's comments *are* its code.
+pub fn compile_tokei_loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
+    match language_type {
+        {% for language, attrs in languages -%}
+            tokei::LanguageType::{{ language }} => {
+                {% set line_types = attrs.tokei_line_types | default(value=['code']) %}
+                language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %}
+            },
+        {% endfor %}
+        _ => unimplemented!("Language Type {:?}", language_type),
     }
 }
 
--- a/src/info/langs/language.tera
+++ b/src/info/langs/language.tera
@@ -110,16 +110,32 @@ impl Language {
     }
 }
 
-/// Counts the top-level lines-of-code of a tokei `Language`. Takes into
+/// Counts the lines-of-code of a tokei `Language`. Takes into
 /// account that a prose language's comments *are* its code.
 pub fn loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
-    match language_type {
+    let loc = match language_type {
         {% for language, attrs in languages -%}
             {%- set line_types = attrs.line_types | default(value=['code']) -%}
             tokei::LanguageType::{{ language }} => language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %},
         {% endfor %}
         _ => unimplemented!("Language Type {:?}", language_type),
-    }
+    };
+    loc + language.children.iter().fold(0, |sum, (lang_type, reports)| {
+        sum + reports.iter().fold(0, |sum, report| sum + stats_loc(lang_type, &report.stats))
+    })
+}
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by spenserblack** [comment]
I can add a test or two if needed.

File: src/info/langs/mod.rs [hunk 1]
```diff
@@ -31,7 +31,7 @@ fn get_language_distribution(languages: &tokei::Languages) -> Option<HashMap<Lan
     let mut language_distribution = HashMap::new();
 
     for (language_name, language) in languages.iter() {
-        let mut code = language.code;
+        let mut code = language::compile_tokei_loc(language_name, language);
```
o2sh, line 34:
```suggestion
        let mut loc = language::loc(language_name, language);
```
minor: it may be confusing because from my understanding the "compile" part (tera template -> rust) is done at build time and not at run time.

src/info/langs/language.tera
```diff
@@ -109,6 +110,20 @@ impl Language {
...
+pub fn compile_tokei_loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
+    match language_type {
+        {% for language, attrs in languages -%}
+            tokei::LanguageType::{{ language }} => {
+                {% set line_types = attrs.tokei_line_types | default(value=['code']) %}
+                language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %}
+            },
+        {% endfor %}
+        _ => unimplemented!("Language Type {:?}", language_type),
+    }
```
o2sh, line 124:
```suggestion
    match language_type {
        {% for language, attrs in languages -%}
            {% set line_types = attrs.tokei_line_types | default(value=['code']) -%}
            tokei::LanguageType::{{ language }} => language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %},
        {% endfor %}
        _ => unimplemented!("Language Type {:?}", language_type),
    }
```

  minor: use an inline match syntax to reduce clutter:
  
  ```rust
      match language_type {
        tokei::LanguageType::ABNF => language.code,
        tokei::LanguageType::Abap => language.code,
        tokei::LanguageType::Ada => language.code,
        tokei::LanguageType::Assembly => language.code,
        tokei::LanguageType::AutoHotKey => language.code,
  ```

languages.yaml
```diff
@@ -1311,6 +1311,7 @@ Jupyter:
       - "#9E9E9E"
     chip: "#DA5B0B"
   serialization: jupyter-notebooks
+  tokei_line_types: [code, comments]
```
o2sh, line 1314:
```suggestion
  line_types: [code, comments]
```
minor: I don't think we should expose internal terminology in this file 🤔

src/info/langs/language.tera
```diff
@@ -109,6 +110,20 @@ impl Language {
...
 
+/// Compiles the top-level lines-of-code of a tokei `Language`. Takes into
+/// account that a prose language's comments *are* its code.
+pub fn compile_tokei_loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
+    match language_type {
+        {% for language, attrs in languages -%}
+            tokei::LanguageType::{{ language }} => {
+                {% set line_types = attrs.tokei_line_types | default(value=['code']) %}
+                language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %}
+            },
```
o2sh, line 121:
```suggestion
            {% set line_types = attrs.tokei_line_types | default(value=['code']) -%}
            tokei::LanguageType::{{ language }} => language.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + language.{{ line_type }}{% endfor %},
```

  minor: use an inline match syntax to reduce clutter:
  
  ```rust
      match language_type {
        tokei::LanguageType::ABNF => language.code,
        tokei::LanguageType::Abap => language.code,
        tokei::LanguageType::Ada => language.code,
        tokei::LanguageType::Assembly => language.code,
        tokei::LanguageType::AutoHotKey => language.code,
....
  ```

spenserblack: Good point, I'm a bit out of practice with Rust and forgot that the block was optional 😆

languages.yaml
spenserblack: :thinking: I kind of like exposing "tokei" (at least for now) so that potential contributors will know that they need to reference tokei to see the available "line types". Though perhaps it would be better to add to `CONTRIBUTING.md` instead.

src/info/langs/mod.rs
spenserblack: >from my understanding the "compile" part

Oh, yeah, maybe "compile" wasn't the best word to use. Simple `loc` is much more clear :+1:

o2sh (changes requested):
Nicely done!

Left a few remarks.

Don't forget to apply the same logic to the `get_total_loc` function.

spenserblack:
Thanks for the review! I'll get to `get_total_loc` over the weekend.

src/info/langs/mod.rs
```diff
@@ -31,23 +31,23 @@ fn get_language_distribution(languages: &tokei::Languages) -> Option<HashMap<Lan
...
-        let mut code = language.code;
+        let mut loc = language::loc(language_name, language);
 
         let has_children = !language.children.is_empty();
 
         if has_children {
             for reports in language.children.values() {
                 for stats in reports.iter().map(|r| r.stats.summarise()) {
-                    code += stats.code;
+                    loc += stats.code;
```
o2sh, line 41: `language::loc` should probably also applied here in case one of the children language happens to be Markdown.

To handle this case for example:

image

languages.yaml
```diff
@@ -1311,6 +1311,7 @@ Jupyter:
       - "#9E9E9E"
     chip: "#DA5B0B"
   serialization: jupyter-notebooks
+  line_types: [code, comments]
```
o2sh, line 1314: Actually, we may not need it for `Jupyter`. Seems like tokei treats jupyter notebooks only as a holder for other languages with 0 as a count for all line types:

image

src/info/langs/mod.rs
```diff
@@ -65,11 +65,9 @@ fn get_language_distribution(languages: &tokei::Languages) -> Option<HashMap<Lan
 }
 
 fn get_total_loc(languages: &tokei::Languages) -> usize {
-    languages
-        .values()
-        .collect::<Vec<&tokei::Language>>()
-        .iter()
-        .fold(0, |sum, val| sum + val.code)
+    languages.iter().fold(0, |sum, (lang_type, lang)| {
```
o2sh, line 68: here is unit test for this method:

```rust
    #[test]
    fn get_total_loc_counts_md_comments() {
        let js = tokei::Language {
            blanks: 25,
            comments: 50,
            code: 100,
            ..Default::default()
        };
        let js_type = tokei::LanguageType::JavaScript;

        let md = tokei::Language {
            blanks: 50,
            comments: 200,
            code: 100,
            ..Default::default()
        };
        let md_type = tokei::LanguageType::Markdown;

        let mut languages = tokei::Languages::new();
        languages.insert(js_type, js);
        languages.insert(md_type, md);

        assert_eq!(get_total_loc(&languages), 400);
    }
```

src/info/langs/mod.rs
spenserblack: That's a good point. I don't use Jupyter notebooks, so I thought Markdown was the "top level" code. Could that Markdown then contain its own code blocks? If so, perhaps this function should be recursive for deeply nested code blocks?

o2sh: Indeed, tokei's parser is indeed recursive, as stated in its [changelog](https://github.com/XAMPPRocky/tokei/blob/master/CHANGELOG.md)

image

spenserblack:
Oops, I thought that I would be labelled the committer and you would be the author of https://github.com/o2sh/onefetch/pull/937/commits/f41bc02eb1c70c9abf27927c573eb42bce55d5a3. I guess codespaces behave differently. Hope you don't mind that I made a commit that looks like you.

src/info/langs/mod.rs
spenserblack: Deeply nested stats are now recursively counted with https://github.com/o2sh/onefetch/pull/937/commits/939e9e1f1b5f78cb5a136ccf8dcaf1b932e64d00

spenserblack:
As something to do in the future, I think it would be nice to collect code samples like linguist and tokei do, and perform some snapshot tests on the CLI output using those samples. As a lazy test, I used
```jupyter-notebook
{
 "cells": [
  {
   "attachments": {},
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "## Setup\n",
    "\n",
    "```bash\n",
    "echo 'Hello, World!'\n",
    "```"
   ]
  }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.4.3"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 0
}
```
And got 4 lines as expected (Jupyter contains Markdown, Markdown contains 3 comment lines and 1 line of Bash).

src/info/langs/mod.rs
```diff
@@ -31,23 +31,13 @@ fn get_language_distribution(languages: &tokei::Languages) -> Option<HashMap<Lan
...
 
     for (language_name, language) in languages.iter() {
-        let mut code = language.code;
+        let mut loc = language::loc(language_name, language);
 
-        let has_children = !language.children.is_empty();
-
-        if has_children {
-            for reports in language.children.values() {
-                for stats in reports.iter().map(|r| r.stats.summarise()) {
```
o2sh, line 40: Looking at tokei's documentation for `summarise `:

>     /// Creates a new `CodeStats` from an existing one with all of the child
>     /// blobs merged.

If I'm not mistaken this method already handles the recursion for use 🤔

spenserblack: Oh :laughing: I was so focused on the struct fields I missed that. Thanks for checking!

spenserblack:
Just checked the source code of `.summarise`, and it looks like we'll lose some granularity.
Before https://github.com/o2sh/onefetch/pull/937/commits/aa31a72a583a1ad975d2106c81638a6d272c4654, a comment in a bash code block in Markdown in Jupyter would be ignored, since only code is counted for Bash. Summarise would combine the `CodeStats` from the Bash with the Markdown, making *all* comments get counted, as long as the 1st-level child is Markdown.

Which may be preferable behavior (it makes sense to me IMO), but I just want to throw that out there.

src/info/langs/language.tera
```diff
@@ -109,6 +110,34 @@ impl Language {
...
+pub fn stats_loc(language_type: &tokei::LanguageType, stats: &tokei::CodeStats) -> usize {
+    let stats = stats.summarise();
+    match language_type {
+        {% for language, attrs in languages -%}
+            {%- set line_types = attrs.line_types | default(value=['code']) -%}
+            tokei::LanguageType::{{ language }} => stats.{{ line_types.0 }}{% for line_type in line_types | slice(start=1) %} + stats.{{ line_type }}{% endfor %},
+        {% endfor %}
+        _ => unimplemented!("Language Type {:?}", language_type),
+    }
+}
```
o2sh, line 139: Don't you think it would be better to keep the code logic outside of the tera template? I'm refering to lines 123-125 and 131. 

We could probably extract those lines and put them in `get_language_distribution()` between line 34 and 36:

```
loc += language
  .children
  .iter()
  .fold(0, |sum, (lang_type, reports)| {
      sum + reports.iter().fold(0, |sum, report| {
          sum + language::stats_loc(lang_type, &report.stats.summarise())
      })
  });
```

this would simplify the template but more importantly avoid obfuscating part of the code logic.

In the template, we would then have two methods `loc` and `stats_loc` which would be almost identical except for one of their parameters. I am wonder if there is a way we could combine the two methods 🤔 (maybe in the future)

spenserblack: :thinking: Yeah, it does obfuscate it a bit by hiding the logic in the template. But my thinking was that I'd like to avoid needing to repeat similar logic (adding the LOC of the children) for both language distributions and total line counts. Since these are kind of private methods, maybe they should be `#[doc(hidden)]` and called `__loc()`, and have more visible methods called `loc()` and `stats_loc()` in `language.rs` that call these methods and act as wrappers.

```rust
fn loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
    __loc(language_type, language) + language
        .children 
        // ...
```
What do you think?

o2sh: As you mentioned in your comment https://github.com/o2sh/onefetch/pull/937#issuecomment-1399553995
We'll have to add some unit tests of more complex cases (language with children) to make sure we don't break things.

o2sh: > But my thinking was that I'd like to avoid needing to repeat similar logic (adding the LOC of the children) for both language distributions and total line counts.

You're right, I missed that 🤦 

> Since these are kind of private methods, maybe they should be #[doc(hidden)] and called __loc(), and have more visible methods called loc() and stats_loc() in language.rs that call these methods and act as wrappers.

I'm sorry, but I'm lost - too many `loc` 😅 - maybe you can commit your suggestion to illustrate your idea. But yeah, if we could avoid putting logic inside the tera template, that would be great. we could probably factorize this part inside a method:

```
let mut loc = language::loc(language_name, language);

let has_children = !language.children.is_empty();

if has_children {
    for (lang_type, reports) in language.children {
        loc += reports
            .iter()
            .map(|r| r.stats.summarise())
            .fold(0, |sum, stats| sum + language::stats_loc(&lang_type, &stats));
    }
}
```

and call it for both language distributions and total line counts.

But it's your call, do what you think is best 👍 I think the PR is already pretty good.

o2sh (approved):

src/info/langs/language.tera
spenserblack: > maybe you can commit your suggestion to illustrate your idea.

Sure! https://github.com/o2sh/onefetch/pull/937/commits/3e14718ac30022257e837fbc1e36bd24b91962df is the basic idea is that `__<function>` contains the repetitive code that the template manages, and `<function>` is a wrapper that calls that function and includes some helper code.

spenserblack: I'll probably add a test for Jupyter code and call it on `get_total_loc`. That way `loc` and `stats_loc` is 100% covered. `__loc`'s *template* would be 100% covered, but the *generated code* would probably be <10% with each language being a branch :laughing:
I think I'll just hardcode the structs necessary to simulate Jupyter code.

o2sh: Perfect, thanks @spenserblack

PR #939, opened 2023-01-20
Fix path to language template

spenserblack:
Codeowners file was pointing to an old path that doesn't exist.


o2sh (approved):
Good catch!
Thanks.

PR #948, opened 2023-01-28
Refactoring of `info/langs/mod.rs`

o2sh:
Folllowing #937, I couldn't help but notice that [`info/langs/mod.rs`](https://github.com/o2sh/onefetch/blob/main/src/info/langs/mod.rs) could need some refactoring for better readability but also for performance reasons.

Covered in this PR:

- `language::loc`  is called only in `get_loc_by_language` and reused in `get_total_loc`
- Better naming for the methods and simplified signatures - up for discussion.
- Extract some of the code logic from [`info/langs/mod.rs`](https://github.com/o2sh/onefetch/blob/main/src/info/langs/mod.rs) into their respective info fields: [`LanguageInfo`](https://github.com/o2sh/onefetch/blob/main/src/info/langs/language.rs) and [`LocInfo`](https://github.com/o2sh/onefetch/blob/main/src/info/loc.rs)

### DIFF
--- a/src/info/langs/mod.rs
+++ b/src/info/langs/mod.rs
@@ -12,6 +12,9 @@ pub fn get_main_language(loc_by_language: &[(Language, usize)]) -> Language {
     loc_by_language[0].0
 }
 
+/// Returns a vector of tuples containing all the languages detected inside the repository.
+/// Each tuple is composed of the language and its corresponding loc (lines of code).
+/// The vector is sorted by loc in descending order.
 pub fn get_loc_by_language_sorted(
     dir: &Path,
     ignored_directories: &[PathBuf],
--- a/src/info/langs/mod.rs
+++ b/src/info/langs/mod.rs
@@ -8,27 +8,34 @@ use strum::IntoEnumIterator;
 
 pub mod language;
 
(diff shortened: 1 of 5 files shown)


spenserblack (approved):
Looks good!

src/info/langs/mod.rs
```diff
@@ -8,27 +8,34 @@ use strum::IntoEnumIterator;
...
 pub mod language;
 
-pub fn get_dominant_language(languages_stat_vec: &[(Language, f64)]) -> Language {
-    languages_stat_vec[0].0
+pub fn get_main_language(loc_by_language: &[(Language, usize)]) -> Language {
+    loc_by_language[0].0
 }
 
-pub fn get_language_statistics(
+pub fn get_loc_by_language_sorted(
```
spenserblack, line 15: >Better naming for the methods and simplified signatures - up for discussion.

I guess, on this subject, `_loc_` could potentially be confusing to new contributors, since "loc" is frequently an abbreviation for "location". `lines_by_language_sorted` might be slightly more understandable to newcomers.

o2sh: Good point, I added some documentation to clarify the terminology and explain the method's intent.

PR #963, opened 2023-02-19
upgrade `git-repository` 0.30 to `gix` 0.36

Byron:
This upgrade is mainly to benefit from the new names of all `git-*` crates which now share the `gix-` prefix, with `gix` naturally being what was formerly `git-repository`.

Now there is no need anymore to rename crates in order to get the `git` namespace, instead we refer to `gitoxide` as `gix` from the get-go.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -724,6 +724,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bd4b30a6560bbd9b4620f4de34c3f14f60848e58a9b7216801afcb4c7b31c3c"
+
+[[package]]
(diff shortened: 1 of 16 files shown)


o2sh (approved):
Thanks @Byron 👍 

BTW, do you have an ETA for the "git status/pending changes" support in gitoxide? Once we have this final piece, we'll be able to completely abandon `git2` at last.

Byron:
Interestingly I thought exactly the same when doing this PR, just one more feature to go!

Unfortunately I have to follow priorities dictated by `cargo` for this half of the year which doesn't include this feature. Fortunately though, we don't have to wait that long as @pascalkuthe has already indicated interest in the task and I couldn't be more excited - `helix` is waiting for this feature too and I know we are going to get a high-performance, state-of-the-art implementation once he gives it a go 🎉.

PR #971, opened 2023-02-28
Add Coldfusion and ColdfusionScript support

theemanofsteele:
Adds support for ColdFusion.
In Linguist and Tokei, they have "ColdFusion" and "ColdFusionScript" that collectively cover all ColdFusion files. 
Figured same ASCII art for both is ok. Open to suggestions and feedback though!

Supports this issue: https://github.com/o2sh/onefetch/issues/490

### DIFF
--- a/languages.yaml
+++ b/languages.yaml
@@ -291,6 +291,56 @@ CoffeeScript:
     ansi:
       - red
     chip: "#244776"
+ColdFusion:
+  type: programming
+  ascii: |
+    {0}CfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCf
+    {0}Cf                                    Cf
+    {0}Cf                                    Cf
+    {0}Cf                         CfCfCfCf   Cf
+    {0}Cf       CfCfCfCfCf     CfCfCfCfCf    Cf
+    {0}Cf      CfCfCfCfCf     CfCf           Cf
+    {0}Cf     CfCf           CfCf            Cf
+    {0}Cf    CfCf         CfCfCfCfCfCf       Cf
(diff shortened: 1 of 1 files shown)


spenserblack (approved):
Thanks for the PR!

While I'm personally OK with this, we often try to challenge our contributors to come up with something a bit unique.

A couple of ideas to play around with:
1. A gradient for truecolor (hex).
2. Instead of negative space, a colored background. Maybe a dark blue background, and black for ansi.

Also, I'm unfamiliar with ColdFusion/ColdFusionScript, so I don't know how valid this is, but maybe CFS's logo could be different, unless they truly are the same language.

o2sh (approved):
Very clean design @theemanofsteele, thanks a lot for your contribution.

I left a suggestion, feel free to tell me what you think 👍

languages.yaml
```diff
@@ -291,6 +291,56 @@ CoffeeScript:
...
+    {0}Cf       CfCfCfCfCf   CfCf            Cf
+    {0}Cf                                    Cf
+    {0}Cf                                    Cf
+    {0}CfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCfCf
+  colors:
+    ansi:
+      - Cyan
+    hex:
+      - "#DFF0FD"
+    chip: "#ed2cd6"
```
o2sh, line 343: I agree with @spenserblack's suggestion that you could try creating a spin-off of the original logo (similar to what was done with Fortran) for your submission. However, if you're not feeling inspired at the moment, it's perfectly fine to submit only one entry for Coldfusion and leave the logo variation for another time.

Issue #980, opened 2023-03-12
Does not look well on small terminal


pickfire:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

When running onefetch on a small terminal 80x24, it does not seemed to work well.

### Expected behavior 🤔

The image and the stuff should not overflow. Maybe less people use 80x24 nowadays but it is probably not that rare given that quite a few terminals defaults to this size when starting.

### Steps to reproduce 🕹

1. onefetch

### Additional context/Screenshots 🔦

![Screenshot_20230312_195054](https://user-images.githubusercontent.com/4687791/224542706-64f4bb67-873d-4c3a-8131-7fe806f7764b.png)

### Possible Solution 💡

Reduce the ascii image size.

spenserblack:
Just checked what `neofetch` does, and it looks like it simply cuts off text past the right boundary of the terminal.
[`\e[?7l` seems to be the relevant escape sequence](https://github.com/dylanaraps/neofetch/blob/ccd5d9f52609bbdcd5d8fa78c4fdb0f12954125f/neofetch#L11563).

o2sh:
We've already covered this issue in the past #327 and what came out of it is the following CLI flag:

```
  --show-logo <WHEN>
      Specify when to show the logo

      If set to auto: the logo will be hidden if the terminal's width < 95

      [default: always]
      [possible values: auto, never, always]
```

But I agree with @spenserblack, it would be better to programmatically disable line wrap with `print!("\x1B[?7l")`. Unfortunately, last time I tried I failed: https://github.com/o2sh/onefetch/issues/327#issuecomment-736880866

spenserblack:
@o2sh Thanks for the reminder! Should we consider changing the default value of the flag from always to auto? Technically a breaking change, but it should make it less likely that this issue gets raised again.

pickfire:
Wouldn't it be good to as well reduce the logo size?

spenserblack:
Resizing is certainly an interesting idea, but, giving that ASCII art doesn't have many "pixels", I fear that resized ASCII art could quickly become unreadable and/or diverge from the artist's intent. For example, a resizing algorithm would very likely garble the `NASA` from [Fortran's ASCII art](https://github.com/o2sh/onefetch/blob/4bab48b7f443d92fcc60b704b5de336989b98088/languages.yaml#L716-L734).

o2sh:
>Wouldn't it be good to as well reduce the logo size?

[...]

PR #983, opened 2023-03-15
Disable line wrap

o2sh:
#980

### DIFF
--- a/src/ui/printer.rs
+++ b/src/ui/printer.rs
@@ -18,7 +18,11 @@ pub enum SerializationFormat {
     Yaml,
 }
 
-pub struct Printer {
+pub struct Printer<W>
+where
+    W: Write,
+{
+    writer: W,
     info: Info,
     output: Option<SerializationFormat>,
     art_off: bool,
@@ -30,8 +34,8 @@ pub struct Printer {
     ascii_language: Option<Language>,
 }
 
-impl Printer {
-    pub fn new(info: Info, config: Config) -> Result<Self> {
+impl<W: Write> Printer<W> {
+    pub fn new(writer: W, info: Info, config: Config) -> Result<Self> {
         let art_off = match config.show_logo {
             When::Always => false,
             When::Never => true,
@@ -59,6 +63,7 @@ impl Printer {
         };
 
         Ok(Self {
+            writer,
             info,
             output: config.output,
             art_off,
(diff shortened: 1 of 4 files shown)


src/ui/printer.rs
```diff
@@ -135,7 +135,13 @@ impl<W: Write> Printer<W> {
                     }
                 }
 
+                // Disable line wrapping
+                write!(self.writer, "\x1B[?7l")?;
+
                 write!(self.writer, "{buf}")?;
+
+                // Re-enable line wrapping
+                write!(self.writer, "\x1B[?7h")?;
```
o2sh, line 144: @spenserblack 

I think it's fairly safe to do it like this. My concern was to not leave the user's terminal in an altered state in case of an error with the line wrap still enabled. But with this diff the only scenario in which it could happen is if the program panics at line 141.

What do you think ?

An easy alternative would be to implement the `Drop` trait to `Printer<W>` in which we would add `write!(self.writer, "\x1B[?7h")?;` but it feels over the top 🤔

spenserblack: No, this looks good to me! Since `Drop::drop` doesn't return a `Result`, the call to `write!` would have to unwrap, right? Probably best to keep potentially panicking code out of `drop` :slightly_smiling_face:

Any reason not to make it one call to `write!` instead of 3?

spenserblack (approved):

spenserblack:
Didn't mean to duplicate comments. That's an odd behavior of GitHub :thinking: :laughing:

src/ui/printer.rs
o2sh: >No, this looks good to me! Since Drop::drop doesn't return a Result, the call to write! would have to unwrap, right? Probably best to keep potentially panicking code out of drop 🙂

Very well, that reassures me 😩 

>Any reason not to make it one call to write! instead of 3?

You're right it's better if we merge them into one -> ddba00b

o2sh:
Thanks for the review @spenserblack 

I didn't expect this issue to be resolved with a one-line PR. 😅

spenserblack:
Just thought of something: how does Ctrl+C handle this? If onefetch is taking a long time (like in a massive repo) a user might try to terminate the process. Could that potentially leave the terminal in a "no wrap" state?

o2sh:
That's a good point. But since we disable line wrapping only after generating the output, we should be fine. If the user decides to halt the program, it is likely to happen during the commit traversal or language stat computation stages.

PR #936, opened 2023-01-19
feat(ascii): parse multi-byte unicode chars correctly + docs

ozwaldorf:
# Why

resolves #935 

# What

- take next char and return the remainder, instead of pulling next char and returning a slice of the input
- add unicode render assertion to tests
- add `AsciiArt` usage example
- add badges to `ascii/README.md`

# Demo

## Usage

image

## Docs

image

PR #989, opened 2023-03-22
Add Pascal support

rchastain:


spenserblack (approved):
The 3D effect looks really cool!

I think that this is the first language that uses block characters. @o2sh what do you think? Does this clash with the other artstyles?

o2sh (approved):

o2sh:
I agree with @spenserblack, it looks really cool! 

> I think that this is the first language that uses block characters.

That's right, thanks to @ozwaldorf's contribution at https://github.com/o2sh/onefetch/pull/936, ASCII logos can now incorporate block characters. @rchastain is the first contributor to utilize this new feature.

Thanks a lot @rchastain

spenserblack:
Hi @rchastain! I'm pinging you because you've previously contributed to onefetch to support Pascal. We're considering [changing our language detection tool to gengo](https://github.com/o2sh/onefetch/pull/1152). Gengo needs a bit more language support, so, if you're interested, [we'd love your help](https://github.com/spenserblack/gengo/issues/34)!

PR #996, opened 2023-04-01
feat: remove github token from url field

### LINKED ISSUE #855: Remove github token from repo url
### Summary 💡

If GH token is used to access repository, then its displayed url looks like `https://<token>@github.com/user/repo.git`.  
Displaying this url as is can easily lead to a token leak.  
It makes sense to cut the login and password fields from the url by default and add a special key to display full urls for those who need it.  

### Motivation 🔦

I almost leaked my token by running onefetch while recording the screen.

jim4067:
tries to resolve #855 

get rid of personal access tokens from displaying with onefetch.


spenserblack:
Thanks for getting this started!

This looks very GitHub-specific, though. Other tokens might still leak. Not sure if GitLab, Bitbucket, etc. support tokens in the clone URL, but I'd assume so.

I think it makes sense to remove everything between the `https://` and `@` symbol. E.g.
`https://foo@github.com/bar/baz.git` :arrow_right: `https://github.com/bar/baz.git`

jim4067:
thanks for the heads-up.

jim4067:
@spenserblack do you think it would make sense to have an option for displaying the personal tokens in the cli?

spenserblack:
> do you think it would make sense to have an option for displaying the personal tokens in the cli?

No, they should probably always be hidden.

Could you switch to using regex? IMO it could be a bit more readable.
```regex
https?:\/\/([^@]+@)\w+\.\w+
```
Removing the captured group should do it.

spenserblack:
Also, we'd appreciate if a test was added for this 🙂

tests/fixtures/repo_with_token_url.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
spenserblack, line 1: :thinking: For some reason, this test failed locally with `/bin/sh` for me (yet other tests passed :shrug:). Some quick googling said that `set -eu -o pipefail` is a bashism.

tests/snapshots/repo__repo_with_token_url.snap
```diff
@@ -0,0 +1,15 @@
+---
```
spenserblack, line 1: Tried to disable colors, but I forgot that using empty custom colors is equivalent to passing nothing and letting the default colors get used. But a colored snapshot test isn't terrible, I guess. Looks pretty with `cat path/to/snapshot` :laughing:

spenserblack (approved):

src/info/url.rs
```diff
@@ -35,7 +36,11 @@ fn get_url(repo: &Repository) -> Result<String> {
     }
 
     let remote_url = match remote_url {
-        Some(url) => url,
+        Some(url) => {
+            let pattern = Regex::new(r"(https?://)([^@]+@)").unwrap();
+            let replaced_url = pattern.replace(&url, "$1").to_string();
+            replaced_url
```
o2sh, line 42: You could maybe extract this logic into a separate method with a good name.

It will be easier to unit test it

src/info/url.rs
```diff
@@ -35,7 +36,11 @@ fn get_url(repo: &Repository) -> Result<String> {
     }
 
     let remote_url = match remote_url {
-        Some(url) => url,
+        Some(url) => {
+            let pattern = Regex::new(r"(https?://)([^@]+@)").unwrap();
```
o2sh, line 40: can we have URIs with HTTP what about ssh?

Will this work?

src/info/url.rs
jim4067: @spenserblack, is it okay to remove everything that you added in that commit?

spenserblack: Yes, please do :smile:

o2sh (approved):
Thanks @jim4067

PR #1002, opened 2023-04-02
replace `--show-logo` with `--no-art`

o2sh:
Given the recent addition of #983, perhaps we could simplify the `--show-logo` option into a more straightforward Boolean flag, such as `--no-art`.

```
VISUALS:
      --no-color-palette  Hides the color palette
      --no-art            Hides the ascii art or image if provided
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2190,7 +2190,6 @@ dependencies = [
  "serde_yaml",
  "strum",
  "tera",
- "terminal_size",
  "time",
  "time-humanize",
  "tokei",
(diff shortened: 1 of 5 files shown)


spenserblack (approved):
Yeah, that makes sense to me! `--no-foo` flags are pretty common.
If we implement a config, though, it might be good to having an accompanying `--art` flag, so the usage would be `[--no-]art`. Just that, whenever there's a config option, there's usually also an off and on flag. E.g. `git commit [--no-]gpg-sign`. I wonder if this is a clap builtin :thinking:

PR #1003, opened 2023-04-03
Set snapshot language to plain text

spenserblack:
This is a small change that removes the unusual syntax highlighting on snapshots. GitHub assumes that snapshots are *Jest* snapshots, which are similar to JavaScript.


o2sh (approved):
Thanks @spenserblack

PR #1011, opened 2023-04-09
Improve code coverage of src/info/mod.rs

changhc:
As part of #700. I checked the current code coverage and tried to cover these two methods:
image


o2sh (approved):
Nicely done! Thank a lot @changhc

PR #1012, opened 2023-04-10
Improve code coverage of src/ui/mod.rs

changhc:
Part of #700. The following lines should be covered by the new test cases:

image

Note that the test case `num_to_color_code_match` is almost meaningless. It's only for boosting test coverage a bit.

Result with `tarpaulin`:
```
|| Tested/Total Lines:
|| src/ui/mod.rs: 31/31
55.77% coverage, 681/1221 lines covered
```

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -115,29 +115,4 @@ mod test {
             ]
         );
     }
-
-    #[test]
-    fn num_to_color_code_match() {
-        // This test case is only for test coverage
-        assert_eq!(num_to_color(&0), DynColors::Ansi(AnsiColors::Black));
-        assert_eq!(num_to_color(&1), DynColors::Ansi(AnsiColors::Red));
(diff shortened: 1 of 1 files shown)


src/ui/mod.rs
```diff
@@ -99,4 +99,45 @@ mod test {
...
+                num_to_color(&0),
+                DynColors::Ansi(AnsiColors::Default),
+                DynColors::Ansi(AnsiColors::Yellow)
+            ]
+        );
+    }
+
+    #[test]
+    fn num_to_color_code_match() {
+        // This test case is only for test coverage
```
o2sh, line 121: Not a big fan of test coverage for the sake of task coverage, this is a perfect illustration of [Goodhart’s law](https://en.wikipedia.org/wiki/Goodhart%27s_law)

o2sh (approved):
Thanks again @changhc, this is much appreciated!

I left a small remark

PR #1021, opened 2023-04-23
Added fish git repository greeter script to wiki

TheSast:
Preview [here](https://github.com/TheSast/onefetch/wiki/getting-started)

### DIFF
--- a/.github/wiki/Getting-Started.md
+++ b/.github/wiki/Getting-Started.md
@@ -44,6 +44,25 @@ cd() {
 check_directory_for_new_repository
 ```
 
+By @TheSast
+
+A fish adaptation of the previous script to put in your `$XDG_CONFIG_HOME/fish/functions/cd.fish`:
+```fish
+function cd -w='cd'
+  builtin cd $argv || return
+  check_directory_for_new_repository
+end
+
+function check_directory_for_new_repository
+  set current_repository (git rev-parse --show-toplevel 2> /dev/null)
+  if [ "$current_repository" ] && \
+    [ "$current_repository" != "$last_repository" ]
+    onefetch
+  end
+  set -x last_repository $current_repository
+end
+```
+
 By @mbrehin
 ```sh
 # Add Git alias for onefetch.
--- a/.github/wiki/Getting-Started.md
+++ b/.github/wiki/Getting-Started.md
@@ -46,7 +46,7 @@ check_directory_for_new_repository
 
 By @TheSast
 
-A fish adaptation of the previous script to put in your `$XDG_CONFIG_HOME/fish/functions/cd.fish`:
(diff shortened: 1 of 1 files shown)


.github/wiki/Getting-Started.md
```diff
@@ -44,6 +44,25 @@ cd() {
 check_directory_for_new_repository
 ```
 
+By @TheSast
+
+A fish adaptation of the previous script to put in your `$XDG_CONFIG_HOME/fish/functions/cd.fish`:
```
spenserblack, line 49: Couldn't you use fish's built-in `funcsave` to save the functions to the correct path for each OS?

.github/wiki/Getting-Started.md
```diff
@@ -44,6 +44,25 @@ cd() {
...
+```fish
+function cd -w='cd'
+  builtin cd $argv || return
+  check_directory_for_new_repository
+end
+
+function check_directory_for_new_repository
+  set current_repository (git rev-parse --show-toplevel 2> /dev/null)
```
spenserblack, line 57: Should these be `set --local`?

spenserblack (commented):
As a fish user, this is much appreciated!

.github/wiki/Getting-Started.md
```diff
@@ -44,6 +44,25 @@ cd() {
...
+  if [ "$current_repository" ] && \
+    [ "$current_repository" != "$last_repository" ]
+    onefetch
+  end
+  set -x last_repository $current_repository
```
o2sh, line 62:
```suggestion
  set -gx last_repository $current_repository
```

When I tested locally, it seems to always run `onefetch` when navigating inside a repository.

To fix this issue, I added the -g flag to the set command to mark last_repository as a global variable, so its value is remembered between function calls

.github/wiki/Getting-Started.md
TheSast: The variable is local to the currently executing function, I think `--local ` is used to restrict scope to the innermost block.
 AFAIK this should only be an issue if this was not inside a function, which it is,

.github/wiki/Getting-Started.md
TheSast: Thanks for catching the typo!

.github/wiki/Getting-Started.md
TheSast: I didn't think of that, I could and I should.
[ I did as you suggested](https://github.com/TheSast/onefetch/commit/b27e54080faa3a5e60f7ef7621d737000c77750a)

spenserblack (approved):

o2sh (approved):
Thanks a lot for your first contribution @TheSast 🎉 and to @spenserblack for taking the time to review it 😊

PR #1034, opened 2023-05-01
Fix 050568f1: run unit tests in generated git repo

antecrescent:
test_info_style_info() and test_info_style_subtitle() fail with `Error: Could not find a git repository in '.' or in any of its parents` if the current working directory is not a git repository. This breaks workflows of e.g. Linux distributions which download tarballs from Github.

Currently, the `fn repo(name: &str) -> Result<Repository>` git repository setup function is duplicated across benchmarking, integration and unit tests. This pull requests unifies it under an opt-in library function via the `test-utils` feature and fixes the aforementioned tests.

I chose to handle this with a feature, because when compiling integration tests, `#[cfg(test)]` [is not active ](https://stackoverflow.com/a/68877141) and the normally compiled lib gets linked against them.
[Using a workaround](https://github.com/rust-lang/cargo/issues/2911#issuecomment-1464060655), the `test-utils` feature and all tests depending on it, are enabled by default for testing and benchmarking.

As of now, it is not possible to add dependencies [based on optional features](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies), which is why I had to move `gix-testtools` to [dependencies].

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ gix-features-for-configuration-only = { package = "gix-features", version = "0.2
 gix = { version = "0.44.1", default-features = false, features = [
     "max-performance-safe",
 ] }
+gix-testtools = { version = "0.11.0", optional = true }
 git2 = { version = "0.17.1", default-features = false }
 human-panic = "1.1.4"
 image = "0.24.6"
@@ -52,13 +53,18 @@ yaml-rust = "0.4.5"
 
 [dev-dependencies]
 criterion = "0.4.0"
-gix-testtools = "0.11.0"
 insta = { version = "1.29.0", features = ["json", "redactions"] }
+onefetch = { path = ".", features = ["test-utils"] }
 pretty_assertions = "1.3.0"
 
 [[bench]]
 name = "repo"
 harness = false
+required-features = ["test-utils"]
+
+[[test]]
+name = "repo"
+required-features = ["test-utils"]
 
 [build-dependencies]
 lazy_static = "1"
@@ -75,3 +81,4 @@ enable-ansi-support = "0.2.1"
 
 [features]
 fail-on-deprecated = []
+test-utils = ["gix-testtools"]
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,0 +1,9 @@
+use anyhow::Result;
+use gix::{open, Repository, ThreadSafeRepository};
+
+pub fn repo(name: &str) -> Result<Repository> {
+    let name = name.to_string();
+    let repo_path = gix_testtools::scripted_fixture_read_only(name).unwrap();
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -52,13 +53,18 @@ yaml-rust = "0.4.5"
 
 [dev-dependencies]
 criterion = "0.4.0"
-gix-testtools = "0.11.0"
 insta = { version = "1.29.0", features = ["json", "redactions"] }
+onefetch = { path = ".", features = ["test-utils"] }
```
o2sh, line 57: I don't think is line is needed, the `required-features` field you added in [[bench]] and [[test]] tables should be enough

src/utils.rs
o2sh, line 1: We should probably give this module a more specific name, `test_utils`? 🤔

o2sh:
IMO, the `style_info` and `style_subtitle` functions could be improved by refactoring them into standalone helper functions that are independent of the `Info` struct. This would simplify the unit testing process and eliminate the need for integration testing tools like `gix-testtools`. By isolating these functions, they can be tested independently without having to instantiate the entire `Info` struct.

What do you think @antecrescent ?

spenserblack:
Agree with @o2sh that we're jumping through too many hoops to even set up our tests to be run.

We could also possibly refactor `Info` to be easier to construct. Right now it's one `new` method that is, frankly, massive (or at least that's my feeling after using linters that penalize methods with more than 10 lines :laughing:)

Seems like it's very idiomatic in Rust to have a `FooBuilder` struct, so maybe that's an option?
```rust
// Example:
let info = Info::builder() // returns InfoBuilder::default()
    .no_color_palette()
    .title(Title::builder().build()) // Maybe a title builder since it's a complex struct?
    .build();
// No color palette, no info fields, so now this test can focus solely on the title
assert_eq!(info.to_string(), "Name ...");
```
`Info::new` could be refactored to use this builder, and tests could also use the builder.

o2sh:
>Right now it's one new method that is, frankly, massive (or at least that's my feeling after using linters that penalize methods with more than 10 lines 😆)

I think I've been staring at it for so long that it doesn't even shock me anymore 😄 , but yeah, I agree this constructor needs some serious refactoring.

>Seems like it's very idiomatic in Rust to have a FooBuilder struct, so maybe that's an option?

Very cool idea 👍 With this in place, we should be able to add some unit tests on the `Info` struct.

@antecrescent It's your call if you want to take care of this refactor suggested by @spenserblack . IMO, this may be out of scope considering your initial issue. Just updating `style_info` and `style_subtitle` into standalone helper functions should be enough.

antecrescent:
Thank you for the thorough reviews. Unfortunately, I am very new to Rust and quite unfamiliar with the code base. I will try to update `style_info` and `style_subtitle` until the end of next week.

o2sh:
Close as part of #1047 1047

PR #1047, opened 2023-05-06
refacto: add info builder pattern

o2sh:
As suggested here:  https://github.com/o2sh/onefetch/pull/1034#issuecomment-1531937437


spenserblack (approved):
This is awesome! It looks like this will make it a lot easier to test onefetch by granularly building an `Info` struct.

PR #1058, opened 2023-05-17
docs: fix `cd` snippet for `cmd.exe`

mataha:
This addresses some of the issues I've run into while using the snippet for some time:

 - handle misplaced quotes in arguments (and some other things)
 - silence `git` output when navigating to a directory junction
 - don't trigger the hook when staying in the same directory or when required environment variables are not present
 - set last repository regardless of executing onefetch or not
 - add primitive error handling

I've "formatted" it now since it was barely readable when inlined.


o2sh (approved):
🎉 🎉

Issue #1092, opened 2023-06-22
Errs when invoked on partial clones


icp1994:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

```sh
$ git clone --filter=blob:none https://github.com/sharkdp/fd
$ onefetch fd/
Error: sending on a closed channel
```

### Expected behavior 🤔

I don't know if [partial clones](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/) have some limitations which prevents onefetch from working so maybe erring is the expected behavior.

icp1994:
Possible duplicate/related https://github.com/o2sh/onefetch/issues/978. Also, onefetch 2.18.0

spenserblack:
Would you be willing to try out an earlier version? I'm curious if this was introduced in the recently released version.

icp1994:
Works with v2.17.1

spenserblack:
Thanks for checking!

The fact that it was introduced in v2.18, along with this error

> ```
> Error: sending on a closed channel
> ```

Makes me think that this was introduced with our new file churn metric.

icp1994:
https://github.com/o2sh/onefetch/commit/d5f86216c596b56940f80c97fdb8635c3be640bf is the last "good" commit. With the next commit https://github.com/o2sh/onefetch/commit/1955153a9f998780b78bc391df1a94464c57cc80 I get:
```
Error: Could not find blob for similarity checking

Caused by:
    An object with id 0da0e9c3963daf5e99f99d2adbf825ce15090c58 could not be found
```
Same with https://github.com/o2sh/onefetch/commit/fd1d5f04bbf1fbf160506fad4a05dd2b1dd35fd9 and finally after the following commit https://github.com/o2sh/onefetch/commit/09c4dc9df6e1201eeed54b01f3432f4097ff54dc it shows the current error.

spenserblack:
Thanks for bisecting and finding the exact commits!

o2sh:
@icp1994 new release with a fix -> https://github.com/o2sh/onefetch/releases/tag/2.18.1

icp1994:
Thanks! Tested locally, works as intended.

PR #1093, opened 2023-06-24
don't fail when computing diff on partial clones

o2sh:
fix for #1092 

I tested it manually on the repo that caused the issue `git clone --filter=blob:none https://github.com/sharkdp/fd` and it seems to work fine. 

However, I couldn't reproduce the bug with an integration test (cf. comment below)😢

### DIFF
--- a/tests/fixtures/make_partial_repo.sh
+++ b/tests/fixtures/make_partial_repo.sh
@@ -18,6 +18,14 @@ mkdir base
     git add code.rs
     git commit -q -m c4
     echo more >> code.rs
+    git mv code.rs renamed.rs
+    echo change >> renamed.rs
+    git commit -q -am c5
+
+    git config uploadpack.allowfilter true
 )
 
 git clone --filter=blob:none file://$PWD/base partial
+(cd partial
+    git config diff.renames true
+)
--- a/tests/fixtures/make_partial_repo.sh
+++ b/tests/fixtures/make_partial_repo.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+set -eu -o pipefail
+
+mkdir base
+(cd base
+    git init -q
+    git checkout -b main
+    touch code.rs
+    git add code.rs
+    git commit -q -m c1
+    echo hello >> code.rs
+    git add code.rs
+    git commit -q -m c2
+    echo world >> code.rs
+    git add code.rs
+    git commit -q -m c3
+    echo something >> code.rs
+    git add code.rs
+    git commit -q -m c4
+    echo more >> code.rs
+)
+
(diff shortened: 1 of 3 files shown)


tests/fixtures/make_partial_repo.sh
o2sh, line 1: I can't reproduce the bug with this script. It seems like the blobs are still fetched for previous commits even with `--filter=blob:none`. @Byron @spenserblack Any idea how I could setup a repository to get the following [error](https://github.com/o2sh/onefetch/issues/1092#issuecomment-1602969382)

Byron (approved):
I could reproduce the issue and fix it so that the diff won't try to access blobs in the first place. It's a bit tricky and I missed this behaviour as well. I have improved the documentation accordingly.

tests/fixtures/make_partial_repo.sh
Byron: There was an issue with the configuration that is fixed - by trying the lines by hand in the respective `fixtures/generated-do-not-edit` folder I observed a warning which led to tracking down the option to enable filtering when `git upload-pack` is invoked.r

tests/fixtures/make_partial_repo.sh
```diff
@@ -0,0 +1,31 @@
...
+    git mv code.rs renamed.rs
+    echo change >> renamed.rs
+    git commit -q -am c5
+
+    git config uploadpack.allowfilter true
+)
+
+git clone --filter=blob:none file://$PWD/base partial
+(cd partial
+    git config diff.renames true
```
Byron, line 30: This is just to be explicit about this option as it is the one that matters. If unset, it would still default to `true` though, but I thought it's nice to see all options that are important here.

tests/fixtures/make_partial_repo.sh
```diff
@@ -0,0 +1,31 @@
...
+    git commit -q -m c3
+    echo something >> code.rs
+    git add code.rs
+    git commit -q -m c4
+    echo more >> code.rs
+    git mv code.rs renamed.rs
+    echo change >> renamed.rs
+    git commit -q -am c5
+
+    git config uploadpack.allowfilter true
```
Byron, line 25: This makes `--filter=blob:none` work, otherwise we see a warning when cloning, saying that the option is ignored as the server didn't support it.

tests/fixtures/make_partial_repo.sh
spenserblack: Great idea! Yeah, I think it's a good philosophy that tests' setup code describes what's being tested, even if it's unnecessarily explicit.

spenserblack (approved):

PR #1100, opened 2023-06-29
add a test for negative dates and see how onefetch handles it

Byron:
Currently there is an 'empty' error which probably is related to a failure to parse commits with negative dates.

This should be fixed in future versions of `gitoxide`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.1"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Byron:
It seems that the cache is getting in the way of even compiling successfully, and I don't know how to fix that.

In theory, this PR already works and adds support for commits with early commit dates, so [this repository](https://archeo-lex.fr/codes/code_civil) could be visualized fi there was any code in it. Maybe that would also be an avenue of improvement for another day.

spenserblack:
>It seems that the cache is getting in the way of even compiling successfully, and I don't know how to fix that.

Perhaps a call to `cargo update` to update the sub-dependencies? Although if one of our dependencies tries to enable a feature that doesn't exist, that's a sign that perhaps its version constraint isn't strict enough.

Byron:
A great idea, let me try that. GitHub also just had some general CI issues, I hope that's not interfering now.

spenserblack (approved):
Looks like the CI is passing :tada:

Byron:
Thanks a lot for the test, much better now!

PR #1101, opened 2023-07-01
Group clap dependency updates

spenserblack:
[Dependabot now supports grouping dependency updates](https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/)

Since `clap_complete` is likely to be dependent on `clap`, it seemed like a good idea to group them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,11 @@ version: 2
 updates:
   - package-ecosystem: cargo
     directory: "/"
+    groups:
+      clap:
+        patterns:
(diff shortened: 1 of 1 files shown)


o2sh (approved):
That's a cool new feature! Thanks @spenserblack 

I'm now tempted to put all the `npm` dependencies in a single group 😅

spenserblack:
Good idea! The `@typescript-eslint/*` dependencies should probably be grouped at least, since their versions should be in sync.

PR #1110, opened 2023-07-07
Group all NPM dependency updates

spenserblack:
https://github.com/o2sh/onefetch/pull/1101#pullrequestreview-1508986895

I don't see any harm in grouping them all, but we could potentially break them out into typescript-eslint, svelte, etc. if we want.

Although this has gotten me thinking how much we should even care about keeping the site's dependencies up-to-date :thinking:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,5 +15,9 @@ updates:
       interval: "weekly"
   - package-ecosystem: npm
     directory: /docs/vercel
+    groups:
+      site:
+        patterns:
+          - "*"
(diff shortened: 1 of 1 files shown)


o2sh:
Thanks @spenserblack 

>Although this has gotten me thinking how much we should even care about keeping the site's dependencies up-to-date

You're right, but It doesn't hurt to be aware of them - especially security/performance patches - and merge them once in a while.  Grouping them in a single PR is great for that.

o2sh (approved):

Issue #1091, opened 2023-06-20
Add option to exclude files from churn


spenserblack:
### Summary 💡

Nice to see this feature in action!

Proposing an option `--churn-exclude` to exclude certain files from the churn metrics. To pair with `--exclude`, it should probably take 1 or more glob patterns in the format `--churn-exclude '**.foo' 'assets/*.png'`

Alternatively, `--exclude` can be used to ignore files for churn metrics. Some users might be surprised to `--exclude` a file and still see it in the churn metric.

### Motivation 🔦

When trying out this new feature, I noticed that `.snap` files for snapshot tests were getting picked up. These can be expected to change a lot, and as a result it's not necessarily useful to see a churn metric for these files. The same is probably true for other generated assets. So it might be useful to be able to ignore certain files.

spenserblack:
Sorry for requesting something that will require a refactor to pass around the config so soon after a release :cry:

o2sh:
>Sorry for requesting something that will require a refactor to pass around the config so soon after a release 😢

No worries! The reason why I was so eager to push a new release was to get feedback, which seems to be working 😄 

>Proposing an option --churn-exclude to exclude certain files from the churn metrics.

That's a great idea, and I agree that it should behave the same way as the `--exclude` flag and accept patterns.

>Alternatively, `--exclude` can be used to ignore files for churn metrics. 

I don't know, maybe we could decouple the two and explain the difference between them in a help message, or by finding a more specific name. Some users may want to exclude certain files/folders from the churn while still having them taken into account in the language distribution 🤔

spenserblack:
>Some users may want to exclude certain files/folders from the churn while still having them taken into account in the language distribution

Sounds good to me! That was my initial instinct, but then I had trouble thinking of a real-world example where I'd want to exclude files from churn metrics but include them for everything else, since my use case is generated files.

o2sh:
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

PR #1120, opened 2023-07-16
exclude files from churn

o2sh:
As suggested in #1091 

I went for the simpler option of reusing the `--exclude` flag for the churn files.

### DIFF
--- a/src/info/langs/mod.rs
+++ b/src/info/langs/mod.rs
@@ -19,7 +19,6 @@ pub fn get_loc_by_language_sorted(
     language_types: &[LanguageType],
     include_hidden: bool,
 ) -> Result<Vec<(Language, usize)>> {
-    println!("{}", dir.to_str().unwrap());
     let stats = get_statistics(dir, globs_to_exclude, language_types, include_hidden);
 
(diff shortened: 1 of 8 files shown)


src/info/langs/mod.rs
```diff
@@ -90,23 +88,6 @@ fn filter_languages_on_type(types: &[LanguageType]) -> Vec<tokei::LanguageType>
...
-            if re.is_match(&dir) {
-                let prefix = if dir.starts_with('/') { "**" } else { "**/" };
-                ignored_directories.push(format!("{prefix}{dir}"));
-            } else {
-                ignored_directories.push(dir);
-            }
-        }
-    }
-    ignored_directories
-}
```
o2sh, line 108: Was meant to resolve this issue https://github.com/XAMPPRocky/tokei/issues/564.

But it turns out that [gitoxide's work_dir function returns a relative path ](https://github.com/Byron/gitoxide/blob/9fbed4b97cd31785d11cbc98c44a8332776a847c/gix/src/repository/location.rs#L44)

spenserblack (approved):

Issue #1033, opened 2023-05-01
Support analysis of tracked files


spenserblack:
### Summary 💡

Converting this comment chain into an issue for discussion/triage: https://github.com/o2sh/onefetch/issues/26#issuecomment-1523524399

Onefetch currently analyzes the filesystem via tokei. However, it seems like some users expect behavior more like `github-linguist`, which analyzes the language stats of *tracked* changes (AFAIK it analyzes a rev, defaulting to HEAD).

Utilizing `git ls-files` can be a partial solution, but it will be buggy:
- Deleted files where the deletion isn't tracked would still show up in `git ls-files`.
- We'd analyze tracked files, but not the tracked *contents* of those files, which can be confusing.

We could potentially hack it by stashing changes and restoring them after analysis. But I'm hesitant to have onefetch change the contents of a repository, even temporarily.

### Motivation 🔦

There was some initial confusion brought up in https://github.com/o2sh/onefetch/issues/26#issuecomment-1522403060, where a large number of untracked files significantly changed the results compared to what the user was expecting.

spenserblack:
cc @dsully

spenserblack:
IMO onefetch should do one or the other -- analyze filesystem *or* git rev. Supporting both would be too complex IMO. So, before thinking of how this could be done, I think the first question is *should* this be done? This would be a significant change and most likely mark a v3 release.

dsully:
Thanks, I got busy over the weekend.

Agreed on not changing the contents of the repository. 

FWIW, my expectation was for `onefetch` to analyze the tracked files at HEAD.

o2sh:
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

PR #1152, opened 2023-08-25
Switch to gengo

### LINKED ISSUE #26: Improve language detection system to recognize C++ headers
https://github.com/github/linguist

Linguist is a tool developed by GitHub for the specific purpose of detecting languages. It's a very mature tool that gets it right the majority of the time by using complex rules.

spenserblack:
This makes some changes to our language support, which I'll be happy to discuss.

Some notable things:
- ~~IDK about the gix side, but~~ gengo supports bare repos, so maybe we can remove that error message
- Some ASCII art has been removed. Currently, that's JSX, TSX, Fish, and Bash.
- Turns out having a dependency that hasn't been released for a while that depends on clap 2 was seriously bloating our dependency tree :laughing: Gengo won't have anything like that since the library and binary are separate crates.

# To Do

- [ ] Gengo needs to support the remaining languages (or at least have reasons to *not* support them) (https://github.com/spenserblack/gengo/issues/34)
- [x] Change pretty much everything that previously matched on `Language` and `tokei::LanguageType` enums.

Resolves #26


spenserblack:
Should the `Language` type in `src/info/langs` be dropped completely? I'm a bit Object-oriented obsessed, so I like calling `language.get_ascii_art` over `get_ascii_art(language)`, but I'm not really seeing a point to this type anymore. Gengo returns a string instead of an enum (e.g. `"C++"` instead of `CPlusPlus`). So you'd really just match on the strings.

```rust
match language {
    "C++" => c_plus_plus_ascii
}
```

The other option is to make language a wrapper for `String` (`struct Language(String)`).

spenserblack:
Hey, looks like the language titles in our vercel deployment are prettier now :tada:

spenserblack:
@o2sh How much do we care about breaking changes in the `onefetch` library?
https://github.com/o2sh/onefetch/pull/1152/commits/6ac0ee001f411cc243c18621b89770b50f614add is trying to be non-breaking by continuing to use a language enum so [this usage doesn't break](https://docs.rs/onefetch/latest/onefetch/info/langs/language/enum.Language.html), but TBH they could probably become just become functions with no implementing type that takes a `&str`, and also probably `pub(crate)` since I can't imaging these functions would be reasonable in a public API.

o2sh:
>@o2sh How much do we care about breaking changes in the onefetch library?

No one seems to be using onefetch as a dependency (https://crates.io/crates/onefetch/reverse_dependencies), so we can do whatever we want 😅 

>Should the Language type in src/info/langs be dropped completely? I'm a bit Object-oriented obsessed, so I like calling language.get_ascii_art over get_ascii_art(language)

Is there a reason why `gengo` doesn't use an `enum` to represent the list of possible languages? I'm a bit surprised, to be honest. 🤔 It seems like a natural choice.

spenserblack:
> Is there a reason why `gengo` doesn't use an `enum` to represent the list of possible languages? I'm a bit surprised, to be honest. 🤔 It seems like a natural choice.

Yeah, I went back and forth a bit on this one. If you look in the commit history [when I was initializing the project](https://github.com/spenserblack/gengo/pull/1), you'll see that I did start of with an enum, building out Rust code with tera like onefetch and tokei does. [I also experimented with a macro](https://github.com/spenserblack/gengo/commit/564a34589ec61b28ee7665c546e36755c1e432ad).

But after thinking about it a bit, I chose `Language { name: String }` over `LanguageType::Foo` for a few reasons:
- I can drop the dependency on a templating library (I also wasn't happy with the macro because the results didn't look good in generated docs)
- Simplifies contributing since you don't have to worry about the language name being valid Rust syntax
- Allows definitions to be defined at runtime. A small delay to interpret those, and the variants aren't known at compile-time, but it makes the code more flexible and testable by allowing for arbitrary definitions.

But mostly, and this might be a bias from taking a break from Rust for a while, but a branch for every single language honestly just seemed like overkill. With several methods on an enum that each match on themselves to return a literal, it just kind of felt like a struct with extra steps :laughing:

In terms of testing: I recognize that coverage on its own isn't indicative of quality, but I don't really like a pattern where a simple addition to a data file will raise or lower test coverage. It's kind of cheating when you get +90% coverage only because you have 100+ branches and a repetitive test for each one :laughing: Though, to be fair, all those tests would certainly guarantee that a language will never be *removed.*

languages.yaml
```diff
@@ -17,7 +16,6 @@ Abap:
       - "#EEEEEE"
     chip: "#E8274B"
```
o2sh, line 17: Maybe the chip `color` could also be given by `gengo` ?

o2sh:
Understood. It's unfortunate that this makes the interface with onefetch somewhat awkward. Additionally, we might encounter runtime errors if we misspell a language name.

languages.yaml
o2sh: Actually, if that's the only reason to keep it, we might as well remove it 🤔

spenserblack: Yeah, that sounds good to me. I didn't add it initially since gengo's colors are different (onefetch's are all from linguist, right?)

Guess we could still get gengo's colors into the vercel UI with WASM, but that feels like major overkill :laughing:

spenserblack:
:thinking: With tokei we had to reference the docs or `languages.json` in order to find the correct way to name a language. Perhaps there's a way to include the contents of `languages.yaml` in the generated docs so it's more visible.

languages.yaml
spenserblack: This is just how I prefer how PRs/diffs are made, but I'm thinking we could remove chip colors and use gengo's in a separate PR.

spenserblack:
Sorry for the *massive* amount of changes in https://github.com/o2sh/onefetch/pull/1152/commits/0d9a7f99781adb0b64dded43e7f26c3db5e12c03, but finally it at least *compiles*. I need to remove the `unimplemented` panics, because running onefetch on itself will panic on `not implemented: GitHub Workflow`.

Some of the changes might be controversial, since I've kind of removed some features. I'd be happy to discuss and get them back in.

One thing to note is that the `--exclude` option might be kind of confusing. Gengo's exclusions are done with
```gitattributes
# .gitattributes
# or you could label this as vendored, generated, or documentation
foo -gengo-detectable
```

spenserblack:
So the snapshot test fails, and my dev environment doesn't play nicely with gix test tools :cry:

I'll see what I can do, but a quick `cargo insta test --accept` would be appreciated :slightly_smiling_face:

spenserblack:
Aaaand no more panic.

I'm going to un-draft this for now. Even if gengo can use a few more updates, the code changes here are ready for review.

I'm thinking it would be cool to make some prereleases to give users a change to play with the unstable version. Something like `onefetch 3.0.0-rc.1` or `-alpha`. Perhaps this PR's target shouldn't be main, but a new branch like `next`.

spenserblack:
Some more things I just noticed:
- Gengo reports blob size, not LOC. So now there are two "size" info lines :shrug:
- Gengo reports if a file is generated. Perhaps that could be used to exclude files from the churn report? `Cargo.lock` and `package-lock.json` have a lot of churn.

src/info/langs/language.tera
```diff
@@ -2,142 +2,77 @@ use owo_colors::{
...
-
-impl From<tokei::LanguageType> for Language {
-    fn from(language: tokei::LanguageType) -> Self {
-        match language {
-            {% for language, _ in languages -%}
-                tokei::LanguageType::{{ language }} => Self::{{ language }},
-            {% endfor %}
-            _ => unimplemented!("Language {:?}", language),
-        }
+pub(crate) fn is_supported(language: &str) -> bool {
```
o2sh, line 11: this method could probably be extracted from the template and use make use of `get_supported_languages` internally

src/info/langs/language.tera
```diff
@@ -2,142 +2,77 @@ use owo_colors::{
...
+    };
+    match colors.true_colors {
+        Some(true_colors) if true_color => true_colors,
+        _ => colors.basic_colors,
     }
 }
 
-fn __loc(language_type: &tokei::LanguageType, language: &tokei::Language) -> usize {
-    match language_type {
+pub fn get_circle_color(s: &str) -> DynColors {
```
o2sh, line 52: maybe rename to `get_chip_color`

src/info/langs/language.tera
```diff
@@ -2,142 +2,77 @@ use owo_colors::{
...
 
-impl From<Language> for tokei::LanguageType {
-    fn from(language: Language) -> Self {
-        match language {
-            {% for language, _ in languages -%}
-                Language::{{ language }} => tokei::LanguageType::{{ language }},
-            {% endfor %}
-        }
+pub(crate) fn get_ascii_art(s: &str) -> &'static str {
+    match s {
```
o2sh, line 21: won't that make it case sensitive, for example if I do `onefetch -a rust`, will it work ?

src/info/langs/language.tera
spenserblack: >and use make use of get_supported_languages internally

I haven't benchmarked this, but its intent is to be a bit of a performance optimization, since `get_supported_langauges` returns a simple iterable, Although I haven't researched how `match` is implemented, I'm assuming it's more performant to match than iterate. Perhaps they could be combined by using a static [`IndexSet`](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html) :thinking:

src/info/langs/language.tera
spenserblack: Good point, I'll normalize the strings to lowercase for this one.

src/info/code_size.rs
```diff
@@ -9,30 +8,30 @@ use crate::{
 
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct LocInfo {
-    pub lines_of_code: usize,
+pub struct CodeSizeInfo {
+    pub size: usize,
```
o2sh, line 12: Can you explain what is the difference between this `size` (which replaces loc) and the other already existing `size` info line.

src/info/langs/mod.rs
```diff
@@ -1,211 +1,36 @@
...
 pub mod language;
 
 pub fn get_main_language(loc_by_language: &[(Language, usize)]) -> Language {
-    loc_by_language[0].0
+    loc_by_language[0].0.clone()
 }
 
 /// Returns a vector of tuples containing all the languages detected inside the repository.
 /// Each tuple is composed of the language and its corresponding loc (lines of code).
 /// The vector is sorted by loc in descending order.
```
o2sh, line 15: comment needs to be updated.

src/info/langs/mod.rs
```diff
@@ -1,211 +1,36 @@
...
-    let mut languages = tokei::Languages::new();
-    let filtered_languages = filter_languages_on_type(language_types);
-
-    let tokei_config = tokei::Config {
-        types: Some(filtered_languages),
-        hidden: Some(include_hidden),
-        ..tokei::Config::default()
+pub fn get_size_by_language_sorted(dir: &Path) -> Result<Vec<(Language, usize)>> {
+    let builder = gengo::Builder::new(dir);
+    let gengo = match builder.build() {
```
o2sh, line 18: can't you use `builder.build()?;`

src/info/langs/mod.rs
```diff
@@ -1,211 +1,36 @@
...
-            (Language::Rust, 1200),
-            (Language::Ada, 300),
-            (Language::Java, 40),
-            (Language::Go, 8),
-        ];
-        let actual_order: Vec<_> = sorted_map.into_iter().collect();
-
-        assert_eq!(expected_order, actual_order);
-    }
+    let analysis = match gengo.analyze("HEAD") {
```
o2sh, line 22: can't you use gango.analyze("HEAD")?;

src/info/code_size.rs
spenserblack: Yeah, this is one of the things that I wanted to discuss since, as it is, it's going to be really confusing to see two size stats for the typical user. But `CodeSizeInfo` is the total size of *blobs at `HEAD`.* E.g. "how big is `src/*` for the latest commit?" `SizeInfo` is, per my understanding, *all* entries, so that would include things like commit messages. @Byron could probably explain it better.

But if you, for example, delete all your code, then "code size" will drop much more significantly than "repo size." And "code size" will be limited to *detected* code, obviously.

I guess a simple example would be that an assets repository, like a collection of images, could have a very large repository size, but a small code size.

Off-topic: regarding "assets" repositories, it might be nice to have an "unknown" ASCII image for repositories that don't have any source code.

src/info/langs/mod.rs
spenserblack: I'd love to :laughing:

Right now gengo returns `Result<T, Box<dyn std::error::Error>>`, and `anyhow::Result` didn't place nicely trying to convert between the two. Maybe I'll give it another go, I was at a point where I was braindead and just begging the compiler to compile :laughing:

src/info/langs/mod.rs
spenserblack: Same as https://github.com/o2sh/onefetch/pull/1152#discussion_r1313835411

src/info/langs/mod.rs
spenserblack: Good catch!

src/info/langs/mod.rs
```diff
@@ -1,211 +1,36 @@
...
```
o2sh, line 22: In which would we want something different than "HEAD"?

src/cli.rs
```diff
@@ -84,18 +84,6 @@ pub struct InfoCliOptions {
...
-    pub include_hidden: bool,
-    /// Filters output by language type
-    #[arg(
-        long,
-        num_args = 1..,
-        default_values = &["programming", "markup"],
-        short = 'T',
-        value_enum,
-    )]
-    pub r#type: Vec<LanguageType>,
```
o2sh, line 98: Will this option be available eventually?

src/info/mod.rs
```diff
@@ -133,22 +134,12 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
...
-    let loc_by_language_sorted_handle = std::thread::spawn({
-        let globs_to_exclude = cli_options.info.exclude.clone();
-        let language_types = cli_options.info.r#type.clone();
-        let include_hidden = cli_options.info.include_hidden;
+    let size_by_language_sorted_handle = std::thread::spawn({
         let workdir = repo_path.clone();
-        move || {
-            langs::get_loc_by_language_sorted(
-                &workdir,
-                &globs_to_exclude,
```
o2sh, line 144: will gengo be able to use `globs_to_exclude` eventually

src/info/code_size.rs
```diff
@@ -9,30 +8,30 @@ use crate::{
...
         }
     }
 }
 
 #[typetag::serialize]
-impl InfoField for LocInfo {
+impl InfoField for CodeSizeInfo {
     fn value(&self) -> String {
-        format_number(&self.lines_of_code, self.number_separator)
+        format_number(&self.size, self.number_separator)
```
o2sh, line 30: Is there a unit? Bytes?

src/info/code_size.rs
o2sh: Feels like we might want to drop this info line alltogether 🤔

o2sh: In which case, maybe `gengo` could directly output the percentage per language instead of the `size`

src/info/langs/mod.rs
o2sh: BTW, it's still hard for me to accept that users have to commit changes for them to be analysed. 😢

Personally, I like to "play" with the language distribution sometimes by deleting files, which is reflected real time in both the `languages` and `pending changes` info lines.

Byron:
Thanks for letting me chime in @spenserblack ! When looking into how `gengo` works I noticed that you are the author, congratulations for the new crate :)! 

I suppose that `onefetch` can do even better with its own code-analysis engine, even though I wonder if it can be as fast as `tokei` if it relies on reading data through `git` (just opening a file in the worktree can hardly be any faster, especially if done in parallel like `tokei` does).

However, independently of that, which probably is solvable, I was surprised to see `git2` as a dependency that we worked so hard to remove, with great performance-payoffs. Now I wonder what might be missing in `gix` that made you pick `git2`?

Even if not in this PR, my goal here clearly would be to get `gix` into `gengo` and then optimize repository object access so it truly is a fast as possible.

I am probably missing a lot of context as well, apologies if anything I said is wrong, but please fill me in so I can learn how things really are 🙏.

src/info/langs/mod.rs
Byron: It should be possible to use `git status` information to read changed files from disk instead *or* choose to read the working tree in the first place, if it's so conveniently present. As mentioned [here](https://github.com/o2sh/onefetch/pull/1152#issuecomment-1703871936) I worry about performance issues with the current implementation (but it's really just that as I didn't test anything).

Personally I feel that `gengo` right now is a breaking change if the worktree isn't considered, but also think that it generally is the way to go in order to get beyond `tokei`, so… excitement :).

src/info/code_size.rs
spenserblack: > In which case, maybe `gengo` could directly output the percentage per language instead of the `size`

Sounds good! I was actually originally going to do that, and that's what `gengo::Analysis::Summary` is for. But once I saw that `Vec<(Language, f64)>` was used in so many places, I opted for getting this is a compilable, reviewable state. I'll look into this, and probably refactoring the code, when I have a good opportunity to dive in.

src/info/langs/mod.rs
spenserblack: Yeah I'm also of the opinion that `gengo` is a breaking change. [I'm considering letting `gengo` take multiple sources](https://github.com/spenserblack/gengo/pull/150) and the filesystem (worktree) would be one of them. But I don't know if I'll actually go that route or commit fully to git as the source.

I was considering [this user](https://github.com/o2sh/onefetch/issues/1033#issuecomment-1529922030) -- a dotfiles repo is a great example of when you'll only want to analyze committed changes. But this was definitely an opinionated choice I made, and I'd love to get more user feedback on expectations.

spenserblack:
Thanks for chiming in, @Byron!



> I suppose that `onefetch` can do even better with its own code-analysis engine, even though I wonder if it can be as fast as `tokei` if it relies on reading data through `git` (just opening a file in the worktree can hardly be any faster, especially if done in parallel like `tokei` does).

With the currently implementation performance will slightly drop with each supported language. But last time I compared tokei and gengo on a clean clone of the [linux repo](https://github.com/torvalds/linux), they were both about the same -- they both took a little over 20 seconds, and gengo was maybe 10% slower (Also checked github-linguist, which took 5 minutes :laughing:).

I did actually consider if onefetch (and any other potential user) doesn't want to go the "gengo way," though, and [made the analyzers public](https://docs.rs/gengo/latest/gengo/languages/analyzer/struct.Analyzers.html). So you can use gengo's language detection without needing to follow gengo's opinions on how to *get* the file's data.

> However, independently of that, which probably is solvable, I was surprised to see `git2` as a dependency that we worked so hard to remove, with great performance-payoffs. Now I wonder what might be missing in `gix` that made you pick `git2`?
>
> Even if not in this PR, my goal here clearly would be to get `gix` into `gengo` and then optimize repository object access so it truly is a fast as possible.

I'm fully open to dropping git2 in favor of gix. I chose git2 to start out with because, even though the implementation is different, I used github-linguist as inspiration, which uses the Ruby equivalent to git2. Also, I'm much more familiar with git2 personally. So I wanted to get the API mostly set up, and now that I have I'm happy to change up the internals. I'd also be excited to see if gix improves performance :smiley:

If you check gengo's `Analyzer`, all it needs is a `AsRef<Path>` (filename) and `&[u8]` (file contents). I haven't dug into gix yet, but I'm sure it can provide that. The only other thing is parsing `.gitattributes` for file attribute overrides.

> I am probably missing a lot of context as well, apologies if anything I said is wrong, but please fill me in so I can learn how things really are 🙏.

Let me know if you have any other questions! 🙂

Byron:
> With the current implementation performance will slightly drop with each supported language. But last time I compared tokei and gengo on a clean clone of the [linux repo](https://github.com/torvalds/linux), they were both about the same -- they both took a little over 20 seconds, and gengo was maybe 10% slower

That's amazing! With this speed using `git2` I think it will be significantly faster with `gix`, especially once it's fully parallelized (and if it is already, then `gix` will make that much faster, like releasing the handbreak). I think it should be easy to test the performance on a typical dev machine to see if more cores change the outcome - I'd think so, but again, with `gix` there would be ways to catchup. If it nonetheless is significantly behind `tokei` for example *or* could be a lot faster by using the worktree, adding support for it seems like a real possibility.

> I'm fully open to dropping git2 in favor of gix.

That's great to hear! Everything you mention, i.e. object access and attribute access, is fully implemented and very fast, too, so there won't be an issue migrating. You can even go as far as to apply filters to the objects to get the same state one would get if a worktree was checked out. Maybe it's better though to simply work on the normalized forms of blobs though.

#### Performance

Meanwhile I realized that general performance indicators can already be obtained by pitting `gengo` vs `tokei`. Here is the result on the `gitoxide` repo:

```
gitoxide ( feat/gix-momo) [$] took 6s
❯ hyperfine -N --warmup 1 gengo tokei
Benchmark 1: gengo
  Time (mean ± σ):     379.9 ms ±   0.8 ms    [User: 146.5 ms, System: 231.5 ms]
  Range (min … max):   378.9 ms … 381.8 ms    10 runs

Benchmark 2: tokei
  Time (mean ± σ):      29.2 ms ±   0.8 ms    [User: 70.6 ms, System: 91.0 ms]
  Range (min … max):    27.9 ms …  35.3 ms    102 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  tokei ran
   13.02 ± 0.34 times faster than gengo
```

On the linux kernel:

```
  Time (mean ± σ):     28.433 s ±  1.271 s    [User: 12.033 s, System: 16.076 s]
  Range (min … max):   27.718 s … 31.294 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: tokei
  Time (mean ± σ):      2.229 s ±  0.010 s    [User: 5.690 s, System: 3.543 s]
  Range (min … max):    2.217 s …  2.244 s    10 runs

Summary
  tokei ran
   12.76 ± 0.57 times faster than gengo
```

`gengo` seems to be single-threaded by the way, something that should definitely be fixed to be competitive. `onefetch` takes 15s on the linux kernel currently, and if `gengo` would be run in parallel to everything else it's budget for the linux kernel would be no more than 15s to not regress performance. Very doable even with just a second thread.

In terms of memory, relying on a `git` repository also has the disadvantage of needing more memory, so worktree-support will definitely lower those requirements. `gengo` uses 690MB peak memory, and `tokei` does it with 149MB, and I am convinced `git2` and its caches are responsible for most of that.

One more thing I noticed, which makes switching to `gix` paramount: `git2` isn't able to deal with a range of repositories that will get more common. For instance, `index.skipHash` is not support, so with my `webkit` repo this happened:

```
❯ gengo
failed to analyze repository: invalid data in index - calculated checksum does not match expected; class=Index (10)
```

Once the index file was recreated with trailing hash though, I did run `gengo` successfully on webkit though which shows that it definitely has advantages to control the engine - it took 1m53s on a single thread. `tokei` for instance didn't want to finish on `webkit` with high-CPU, probably due to some of these files having very long lines that kill some engines that use regex.

#### Conclusion

Some more work is needed until it can replace `tokei`, but I think it's very feasible and very rewarding, too. The first step would be to parallelize object access, and my suggestion here is to build an index from a tree (i.e.flatten it), and then chunk up all blobs onto threads which can then retrieve the objects and analyse them one by one.
And just to share some of my excitement here: now that I have seen `tokei` falter on the `webkit` checkout, I believe `onefetch` needs `gengo` even in order to be even more stable and capable :).

spenserblack:
Thanks for the research! I hadn't put much thought into memory usage. Although I was initially surprised, it makes sense. While I was trying to avoid generating code, memory usage can likely be improved further by doing that, so language definitions are interpreted at build time instead of run time.

Byron:
> Thanks for the research! I hadn't put much thought into memory usage. Although I was initially surprised, it makes sense. While I was trying to avoid generating code, memory usage can likely be improved further by doing that, so language definitions are interpreted at build time instead of run time.

I wouldn't spend any time on memory consumption until worktree analysis is supported - with `git2` or even `gix` involved it's a bit hard to tell where the consumption is coming from. I'd think that with the switch to `gix` that won't be an issue anymore in the context of `onefetch` in particular, which also reads all commits.

Would you be interested in a PR with a `gix` transition? I have an itch now 😅.

spenserblack:
> Would you be interested in a PR with a `gix` transition?

More than interested -- I'd love it!

https://github.com/spenserblack/gengo/issues/154

Byron:
There is now [a PR](https://github.com/spenserblack/gengo/pull/157) over at `gengo` which fixes all performance issues to the point where `onefetch` simply cannot get around using `gengo` anymore. As an example, `WebKit` takes `tokei` a lot of time (for some reason), so onefetch takes 2m 4s. However, `gengo` takes only 2s now (1.85s). It can also churn through the Rust repository and all of its submodules in 1.0s, `tokei` takes 4s for that.

It does appear that going through `git` is actually beneficial for performance, but I am looking forward to seeing a working-tree implementation nonetheless. Reading files from disk probably won't be much faster than `tokei` though, but the jury is still out on that. Definitely exciting times!

Does anybody dare to pitch `gengo-git2`, `gengo-gix` and `linguist` against each other on WebKit? If `linguist` takes 5 min on the Linux Repo, then it's hard to imagine it ever finishes there.

spenserblack:
> WebKit takes tokei a lot of time (for some reason)

I'm going to hazard a guess that one of the contributing factors is that tokei, AFAIK, will read the *entire* contents of every file. This is because tokei's goal is to count code: lines of code, lines of comments, etc. It might be fair to say that that's tokei's primary goal rather than language detection. Since gengo *doesn't* count LoC, we're free to read only small portions of the file.

In fact, if gengo ever does take a significant performance penalty for a large file, please raise a bug report :laughing:

src/cli.rs
spenserblack: That's one of the things I wanted to discuss. If we instruct users to use, for example,
```gitattributes
# .gitattributes
my-data-file.json gengo-detectable
```

Then they can have very fine control over what is and isn't reported. Instead of reporting all or no data files, prose files, etc., they can control which specific files are reported. Also it simplifies onefetch's code if onefetch doesn't have to handle this :wink:

Re-implementing the current behavior should be easily possible by using `gengo::Category`, though.

## Pros

- Simplify onefetch's code
- User has fine control over exactly which files to include/exclude

## Cons

- Asking a user to follow the configuration options of a separate tool can be obscure (I personally wasn't fond of asking users to use `.tokeignore` for this reason)

src/info/code_size.rs
spenserblack: Bytes (assuming that's the unit gix returns)

cc @Byron

src/info/code_size.rs
spenserblack: Sorry, think I misread the last comment. I think it makes sense to stop passing around `Vec<(Language, f64)>` everywhere and look into simplifying/refactoring the code. Even though a big inspiration for gengo was improving onefetch's language detection, I don't necessarily want to make it a dedicated "onefetch support" crate, and would prefer more generic implementations, like returning sizes instead of percentages.

src/info/mod.rs
spenserblack: It can be done with `.gitattributes` (see comment in [this thread](https://github.com/o2sh/onefetch/pull/1152/files#r1313841319)). Otherwise it should be pretty trivial to iterate over gengo's analysis results, which maps filepaths to analysis results, and exclude filepaths that match one or more globs.

Again,

## Pros

- Simplify onefetch's code

## Cons

- Obscure usage for those not familiar with gengo

Cargo.toml
```diff
@@ -37,6 +37,7 @@ bytecount = "0.6.3"
 clap.workspace = true
 clap_complete = "4.3.2"
 crossbeam-channel = "0.5.8"
+gengo = "^0.5.0"
```
Byron, line 40: Now that `gengo` is also using `gix` it's worth noting that both need to coordinate their `gix` versions while pre-1.0. Otherwise you would end up with two distinct `gix`-dependency trees which costs double to compile and in the final binary.

o2sh:
Should onefetch report on submodule data?

o2sh:
Sorry for the late answer @spenserblack 

>Should onefetch report on submodule data?

It looks like `linguist` doesn't do that, so maybe we should follow their lead. And yeah, it'd be cool if `gengo` skipped processing submodules by default to speed things up, like mentioned in https://github.com/spenserblack/gengo/pull/164#issuecomment-1706827563

Cargo.toml
spenserblack: I think we would want to create a group in the dependabot config so that `gix` and `gengo` are updated simultaneously. On gengo's side I can try to promptly create a new release for significant `gix` updates.

spenserblack:
> > Should onefetch report on submodule data?
> 
> It looks like `linguist` doesn't do that, so maybe we should follow their lead.

[FYI](https://github.com/spenserblack/gengo/issues/174)

spenserblack:
Heads up that the latest version of gengo supports analyzing directories. You can play with it and compare performance between a fully-generic implementation that analyzes the filesystem vs. a git-specific implementation that analyzes git objects.

o2sh:

spenserblack:
Just want to point out that analyzing a git history can make some fun tricks available. For example, building a history of language info:
```shell
# this example is for the gengo executable, but could easily be converted to a
# onefetch version.
for rev in $(git log --reverse --format="%H"); do
	gengo git --rev "$rev"
	echo  # add a newline
done
```

spenserblack:
v0.11 Now uses an `enum` to define languages

PR #1160, opened 2023-09-02
upgrade `gix` to v0.52

Byron:


Byron:
Just saw this is a dupe of #1156 .

PR #1172, opened 2023-09-18
docs(contributing): Add syntax highlighting to YAML block

spenserblack:
This makes it a bit easier to read the example YAML, and separate the commented-out instructions from the relevant code IMO.


spenserblack:
Looks like the required checks won't run because `**.md` is an ignored pattern :laughing: @o2sh I think only admins can force a merge without required checks running.

IIRC GitHub is considering allowing rulesets to ignore patterns for exactly this issue, but I don't think it's been deployed yet.

PR #1177, opened 2023-09-21
add release.yml file

o2sh:
as suggested in https://github.com/o2sh/onefetch/discussions/1175


spenserblack (approved):
LGTM!

You might want to consider setting up a label `exclude-from-release` or something, to explicitly hide some PRs from the generated changelog. For example, in gengo, I'm planning on using it to ignore refactoring PRs, since that's not user-facing.

PR #1200, opened 2023-11-02
Detect Windows Terminal as a true color terminal

### LINKED ISSUE #1199: Incorrect language color
### Duplicates

### Current behavior 😯

In windows terminal:
QQ截图20231102192629

The dart is red, and the rust is green

### Expected behavior 🤔

But in the VSCode terminal, it displays as expected:

QQ截图20231102192637

### Steps to reproduce 🕹

Open windows terminal and run `onefetch`

_No response_

### Possible Solution 💡

Using VSCode terminal

spenserblack:
Windows terminal does not seem to set `COLORTERM`, but it is capable of
the full range of RGB values. This detects if the `WT_SESSION`
environment variable is set to determine if this session is in Windows
Terminal.

Fixes #1199

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -302,6 +302,8 @@ pub fn is_truecolor_terminal() -> bool {
     env::var("COLORTERM")
         .map(|colorterm| colorterm == "truecolor" || colorterm == "24bit")
         .unwrap_or(false)
+        // Windows Terminal
+        || env::var("WT_SESSION").is_ok()
 }
 
 pub fn get_git_version() -> String {


spenserblack:
So I just dove down a rabbit hole (https://github.com/microsoft/terminal/issues/11057) and what I found interesting is that `WT_SESSION` isn't 100% reliable (https://github.com/microsoft/terminal/issues/11057#issuecomment-1493118152, https://github.com/microsoft/terminal/issues/13006). Basically, if Windows Terminal is not the *first* executable, but `cmd.exe` or `powershell.exe` starts up and then "attaches Windows Terminal to itself" then this environment variable isn't set.

o2sh:
I've found this crate [supports-color](https://crates.io/crates/supports-color) that should cover our need. [Looking at the code](https://github.com/zkat/supports-color/blob/dece38fdbc04d6c29cde18fc9303e6118db58ab4/src/lib.rs#L111), truecolor seems to always be supported on windows. 🤔

o2sh:
This one might be tricky. AFAIK, the only way to determine if a terminal supports truecolor is by using heuristics on environment variables, which can be somewhat 'hacky.'

I've discovered this crate called [supports-color](https://crates.io/crates/supports-color), which is quite popular and should meet our needs. Unfortunately, according to [the code](https://github.com/zkat/supports-color/blob/dece38fdbc04d6c29cde18fc9303e6118db58ab4/src/lib.rs#L111), it states that all terminals in Windows do not support truecolor 🤔, which is not accurate.

This crate is based on another project with the same name: [https://www.npmjs.com/package/supports-color](https://www.npmjs.com/package/supports-color), and it actually does what we need. It returns 'truecolor' for Windows after a certain version.

spenserblack:
> t returns 'truecolor' for Windows after a certain version [chalk/supports-color#71](https://github.com/chalk/supports-color/pull/71)

In that terminal issue one they did recommend just assuming that Windows can support true color as long as we're running on 10 (https://github.com/microsoft/terminal/issues/11057#issuecomment-908496636), so that certainly could be a valid option. But at the same time, I ran the `custom_colors` example from `owo-colors` on "plain" command prompt on Windows 10 and got results that weren't pretty :shrug:

spenserblack:
After a bit of digging, I've found that GitHub checks if the terminal is a virtual terminal, and then falls back to `COLORTERM`:
https://github.com/cli/go-gh/blob/debe718854c4ef26135c2348fd91db360fe5bae8/pkg/term/env.go#L72C47-L72C47
https://github.com/cli/go-gh/blob/debe718854c4ef26135c2348fd91db360fe5bae8/pkg/term/console_windows.go#L12-L18

Whatever the GitHub CLI does could be the way to go. It certainly has plenty of user testing.

o2sh:
FYI, I've opened an issue on `supports-color` 👉  https://github.com/zkat/supports-color/issues/17

o2sh:

spenserblack:
This is a bit of a stretch, but [colored](https://github.com/colored-rs/colored), with built-in color support checking, is now maintained again, if we were interested in switching back and requesting improved Windows support there.

o2sh:
Could be worth it, especially with https://github.com/colored-rs/colored/pull/90 which could simplify the language template.

o2sh:

PR #1207, opened 2023-11-10
Right align authorship percentages

lukehsiao:
This patch right-aligns the percentages for authorship to make them
easier to visually compare.

Before:

    Authors: 53% Ossama Hjaji 1212
             21% dependabot[bot] 485
             4% Spenser Black 88

After:

    Authors: 53% Ossama Hjaji 1212
             21% dependabot[bot] 485
              4% Spenser Black 88

Tested: Cargo test all passes. Also added a new test that checks
  alignment in a simple case, and tested that the output looks as
  intended locally.

### DIFF
--- a/src/info/author.rs
+++ b/src/info/author.rs
@@ -70,6 +70,19 @@ impl AuthorsInfo {
         let authors = git_metrics.authors_to_display.clone();
         Self { authors }
     }
+
+    fn top_contribution(&self) -> usize {
+        if let Some(top_contributor) = self.authors.get(0) {
+            return top_contributor.contribution;
+        }
+        0
+    }
+}
+
+fn digit_difference(num1: usize, num2: usize) -> usize {
+    let count_digits = |num: usize| (num.checked_ilog10().unwrap_or(0) + 1) as usize;
+
+    count_digits(num1) - count_digits(num2)
 }
 
 impl std::fmt::Display for AuthorsInfo {
@@ -77,12 +90,17 @@ impl std::fmt::Display for AuthorsInfo {
         let mut authors_info = String::new();
 
         let pad = self.title().len() + 2;
-
         for (i, author) in self.authors.iter().enumerate() {
             if i == 0 {
                 write!(authors_info, "{author}")?;
             } else {
-                write!(authors_info, "\n{:<width$}{}", "", author, width = pad)?;
+                write!(
+                    authors_info,
+                    "\n{:<width$}{}",
+                    "",
+                    author,
+                    width = pad + digit_difference(self.top_contribution(), author.contribution)
+                )?;
             }
         }
 
@@ -220,5 +238,40 @@ mod test {
         assert!(authors_info
             .value()
             .contains(&"80% Roberto Berto 240".to_string()));
+    }
+    #[test]
+    fn test_author_info_value_alignment_with_three_authors() {
+        let author = Author::new(
+            "John Doe".into(),
+            Some("john.doe@email.com".into()),
+            1500,
+            2000,
+            NumberSeparator::Plain,
+        );
+
+        let author_2 = Author::new(
+            "Roberto Berto".into(),
+            None,
+            240,
+            300,
+            NumberSeparator::Plain,
+        );
+
+        let author_3 = Author::new("Jane Doe".into(), None, 1, 100, NumberSeparator::Plain);
+
+        let authors_info = AuthorsInfo {
+            authors: vec![author, author_2, author_3],
+        };
+
+        assert!(authors_info
+            .value()
+            .contains(&"75% John Doe <john.doe@email.com> 1500".to_string()));
+
+        assert!(authors_info
+            .value()
+            .contains(&"80% Roberto Berto 240".to_string()));
+
+        // Note the extra leading space to right-align the percentages
+        assert!(authors_info.value().contains(&" 1% Jane Doe 1".to_string()));
     }
 }


spenserblack (approved):
Makes sense to me :slightly_smiling_face:

src/info/author.rs
```diff
@@ -42,7 +42,7 @@ impl std::fmt::Display for Author {
         if let Some(email) = &self.email {
             write!(
                 f,
-                "{}% {} <{}> {}",
+                "{:3}% {} <{}> {}",
```
o2sh, line 45:
```suggestion
                "{:>3}% {} <{}> {}",
```
minor, just for readability

src/info/author.rs
```diff
@@ -51,7 +51,7 @@ impl std::fmt::Display for Author {
         } else {
             write!(
                 f,
-                "{}% {} {}",
+                "{:3}% {} {}",
```
o2sh, line 54:
```suggestion
                "{:>3}% {} {}",
```

minor, just for readability

src/info/author.rs
lukehsiao: Great point, done!

src/info/author.rs
lukehsiao: Also done.

src/info/author.rs
```diff
@@ -51,7 +51,7 @@ impl std::fmt::Display for Author {
         } else {
             write!(
                 f,
-                "{}% {} {}",
+                "{:>3}% {} {}",
```
o2sh, line 54: With a systematic left pad of 3, we introduce extra spaces in most cases, misaligning this info line compared to the others.

 I think there are three cases:

1. Only one contributor at 100% -> no left pad
2. Multiple contributors with a top contributor having a 2-digit contribution percentage -> left pad of size 2
3. Less common scenario: multiple contributors, but none of them has a contribution percentage higher than 9% -> no left pad

spenserblack: Ah, I see. In other words, when aligning, the first author shouldn't have extra spaces between it and the stat label?
So, to give another example, with 95% and 5%, the first author shouldn't have any alignment, and the second author should be aligned 1 extra space to fill in for the missing digit?

lukehsiao: While you're point is totally valid, I might argue that eliminating the 1 space is not worth the additional complexity.

In particular, we cannot simply tweak the std::fmt::Display, like this, we'd need a whole different way to dynamically format based on the vector of authors.

Perhaps this could be done in the std::fmt::Display for AuthorsInfo, but definitely would be more complex.

For reference, it looks like this:
```
❯ cargo run --release
    Finished release [optimized] target(s) in 0.15s
     Running `target/release/onefetch`
                 R RR RR                  Luke Hsiao ~ git version 2.42.0
              R RRRRRRRR R          R     -------------------------------
 R RR       R RRRRRRRRRRRRR R      RR     Project: onefetch (5 branches, 44 tags)
rR RRR    R RRRRRRRRRRRRRRRRR R   RRR R   Description: Command-line Git information tool
RRR RR   RRRRRRRRRRRRRRRRRRRRRRR  RRRRR   HEAD: a907d352a019 (main, fork/main)
 RRRRR  RRRRRRRRRRRRRRRRRRRRRRRR  RRRR    Version: 2.18.1
  RRR RRRRRRRRRRRRRRRRRRRRRRRRRRRR RR     Created: 5 years ago
    R  RRRRRRRRRR=  RR = RRRRRRRRRRR      Languages:
     RRRRRRRRRRRR=  RR = RRRRRRRRRR                  ● Rust (97.5 %) ● Shell (2.1 %)
      RRRRRRRRRRR   RR   RRRRRRRRRR                  ● Makefile (0.4 %)
     RR==RRRRRRRRRRRRRRRRRRRRRR===RR      Dependencies: 29 (Cargo)
     RR =  ==RRRRRRR  RRRRRR==  = RR      Authors:  52% Ossama Hjaji 1212
      RR =     ===========     = RR                 21% dependabot[bot] 485
       RR                        R                   4% Spenser Black 88
        R                       R         Last change: 10 hours ago
         R                                Contributors: 136
                                          URL: ssh://git@github.com/o2sh/onefetch.git
                                          Commits: 2309
                                          Churn (246): Cargo.lock 105
                                                       Cargo.toml 76
                                                       …/vercel/package-lock.json 23
                                          Lines of code: 5050
                                          Size: 2.45 MiB (144 files)
                                          License: MIT


```

So, while there is an extra space, IMO it is not visually distracting.

lukehsiao: Nope, I take this back. Turns out the std::fmt::Display of AuthorsInfo makes this pretty trivial to do.

I've updated the commit with the new approach, which now looks like this

```
❯ cargo run --release
   Compiling onefetch v2.18.1 (/home/lukehsiao/repos/github.com/o2sh/onefetch)
    Finished release [optimized] target(s) in 4.95s
     Running `target/release/onefetch`
                 R RR RR                  Luke Hsiao ~ git version 2.42.0
              R RRRRRRRR R          R     -------------------------------
 R RR       R RRRRRRRRRRRRR R      RR     Project: onefetch (5 branches, 44 tags)
rR RRR    R RRRRRRRRRRRRRRRRR R   RRR R   Description: Command-line Git information tool
RRR RR   RRRRRRRRRRRRRRRRRRRRRRR  RRRRR   HEAD: a907d352a019 (main, fork/main)
 RRRRR  RRRRRRRRRRRRRRRRRRRRRRRR  RRRR    Pending: 1+-
  RRR RRRRRRRRRRRRRRRRRRRRRRRRRRRR RR     Version: 2.18.1
    R  RRRRRRRRRR=  RR = RRRRRRRRRRR      Created: 5 years ago
     RRRRRRRRRRRR=  RR = RRRRRRRRRR       Languages:
      RRRRRRRRRRR   RR   RRRRRRRRRR                  ● Rust (97.5 %) ● Shell (2.1 %)
     RR==RRRRRRRRRRRRRRRRRRRRRR===RR                 ● Makefile (0.4 %)
     RR =  ==RRRRRRR  RRRRRR==  = RR      Dependencies: 29 (Cargo)
      RR =     ===========     = RR       Authors: 52% Ossama Hjaji 1212
       RR                        R                 21% dependabot[bot] 485
        R                       R                   4% Spenser Black 88
         R                                Last change: 10 hours ago
                                          Contributors: 136
                                          URL: ssh://git@github.com/o2sh/onefetch.git
                                          Commits: 2309
                                          Churn (213): Cargo.lock 100
                                                       Cargo.toml 71
                                                       …/vercel/package-lock.json 21
                                          Lines of code: 5087
                                          Size: 2.45 MiB (144 files)
                                          License: MIT

```

src/info/author.rs
```diff
@@ -77,12 +77,25 @@ impl std::fmt::Display for AuthorsInfo {
...
             if i == 0 {
                 write!(authors_info, "{author}")?;
             } else {
-                write!(authors_info, "\n{:<width$}{}", "", author, width = pad)?;
+                // Set padding based on top author
+                if alignment_pad == 0 {
+                    alignment_pad = author.to_string().split_once('%').unwrap().0.len();
+                }
+                let author_pad =
+                    alignment_pad - author.to_string().split_once('%').unwrap().0.len();
```
spenserblack, line 91: This could probably be simplified by using `str::find`.

src/info/author.rs
```diff
@@ -77,12 +77,25 @@ impl std::fmt::Display for AuthorsInfo {
...
 
         for (i, author) in self.authors.iter().enumerate() {
             if i == 0 {
                 write!(authors_info, "{author}")?;
             } else {
-                write!(authors_info, "\n{:<width$}{}", "", author, width = pad)?;
+                // Set padding based on top author
+                if alignment_pad == 0 {
+                    alignment_pad = author.to_string().split_once('%').unwrap().0.len();
+                }
```
spenserblack, line 89: I think that, effectively, `alignment_pad == 0` is checking "is this the first iteration in the for loop?" I think this can be moved to the check if the index = 0.

src/info/author.rs
```diff
@@ -221,4 +234,39 @@ mod test {
             .value()
             .contains(&"80% Roberto Berto 240".to_string()));
     }
+    #[test]
+    fn test_author_info_value_alignment_with_three_authors() {
```
spenserblack, line 238: I think that this test is a good candidate for a snapshot test, so it's easier for contributors to visualize the tested behavior.

src/info/author.rs
lukehsiao: Ah, great idea. Made the change.

src/info/author.rs
lukehsiao: Another great idea, made the change.

src/info/author.rs
lukehsiao: I'm afraid I'm not familiar with the snapshot tests, or what this means. Could you give me some pointers?

spenserblack: Sure! You'll want to look into `insta` and `cargo-insta`. We have a few snapshot tests in our codebase using macros from `insta` that you can use as an example.

The basic idea of a snapshot test is that you give it only the actual value, and the snapshot test assertion will generate that value in a file for you to review and commit. I typically find these helpful for complex or styled outputs. So, for a snapshot test here, I'd recommend asserting the entire author info stat converted to a string. The snapshot that gets generated will act as a preview of how it will look in the CLI.

src/info/author.rs
```diff
@@ -77,12 +77,22 @@ impl std::fmt::Display for AuthorsInfo {
...
             } else {
-                write!(authors_info, "\n{:<width$}{}", "", author, width = pad)?;
+                // Set padding based on top author
+                let author_pad = alignment_pad - author.to_string().find('%').unwrap();
+                write!(
+                    authors_info,
+                    "\n{:<width$}{}",
+                    "",
+                    author,
+                    width = pad + author_pad
```
o2sh, line 94: That's better @lukehsiao 👍 

We can try to make it cleaner and more readable, here is a suggestion:

-  Create two methods `top_contribution` in `AuthorsInfo` and `left_pad` in `Author`:
```rust
fn top_contribution(&self) -> usize {
    if let Some(top_contributor) = self.authors.get(0) {
        return top_contributor.contribution;
    }
    0
}
```
```rust
fn left_pad(&self, top_contribution: usize) -> usize {
    if top_contribution > 9 && self.contribution < 10 {
        return 1;
    }
    0
}
```

- in `fmt` of `AuthorsInfo` do `width = pad + author.left_pad(self.top_contribution())` when `i > 0`

spenserblack: `usize` has `checked_ilog10` if you want to simplify this a bit.

lukehsiao: Done.

src/info/author.rs
lukehsiao: I'm afraid I still do not understand this. Can you give me some pointers to these existing snapshot tests, and how I'm supposed to run them?

spenserblack: OK, did a quick search of our codebase and it turns out that snapshot tests aren't as common in this project as I thought. There's a JSON snapshot test [here](https://github.com/o2sh/onefetch/blob/5b1ef678c2b3f86e90a848cf20fd236d230b3110/tests/repo.rs#L52-L58), which automatically generated [this snapshot](https://github.com/o2sh/onefetch/blob/5b1ef678c2b3f86e90a848cf20fd236d230b3110/tests/snapshots/repo__repo.snap). I might try to add the snapshot test myself later, but if you want to take one more look I'd recommend starting with [insta's docs](https://insta.rs/docs/quickstart/). You'd probably want either `assert_snapshot` or `assert_display_snapshot`. You can run them like any other test with `cargo test`, but you might want to install `cargo-insta` so that you can use the `cargo insta test` and `cargo insta review` subcommands.

src/info/author.rs
```diff
@@ -70,19 +70,43 @@ impl AuthorsInfo {
...
+    }
+}
+
+fn digit_difference(num1: usize, num2: usize) -> usize {
+    let count_digits = |num: usize| {
+        if num == 0 {
+            1
+        } else {
+            (num.checked_ilog10().unwrap() + 1) as usize
+        }
```
spenserblack, line 88:
```suggestion
        (num.checked_ilog10().unwrap_or(0) + 1) as usize
```
`checked_ilog10` returns `None` if `num` is `0`.

lukehsiao: thanks!

src/info/author.rs
lukehsiao: My preference would to not block this PR on this new test type.

spenserblack: OK, sorry, I didn't realize that you had basically copied similar tests. You're right that this PR shouldn't be blocked, and all of those tests should be updated in a separate PR instead.

spenserblack (approved):

o2sh (approved):
🎉

lukehsiao:
Thanks for the detailed and thoughtful reviews!

PR #1214, opened 2023-11-13
Refactor and test info field styling

spenserblack:
This PR is to illustrate where we might want to switch to snapshot tests. With the change to alignment in #1207, for example, a snapshot test makes it much easier for a review to confirm proper alignment. "assert contains" isn't really an ideal test -- who knows how many leading or trailing characters there are from the test alone?

~~I'm leaving this as a draft right now because, as you can see, these snapshots look kind of weird. That's because they're testing `.value()`, which assumes that it will be preceded by a title. Thus, the `TODO` comment.~~

Edit: After working on this PR a bit to get more useful snapshot tests, I realized there was an opportunity to refactor a bit. This PR now moves the logic that styles info fields onto the `InfoField` trait, and then uses styled snapshot tests (you can view the styled output with `cat path/to/snapshot`) to test and visualize the outputs.

### DIFF
--- a/src/info/author.rs
+++ b/src/info/author.rs
@@ -126,6 +126,9 @@ impl InfoField for AuthorsInfo {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::ui::text_colors::TextColors;
+    use insta::assert_snapshot;
+    use owo_colors::DynColors;
 
     #[test]
     fn test_display_author() {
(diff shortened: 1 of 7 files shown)


o2sh (approved):
You're absolutely right; it makes more sense to have the formatting logic of info_fields inside info_field trait 😅. Additionally, `get_style` and `format_number` could also be moved out of info/mod.rs into info/utils/mod.rs.

Thanks as well for implementing the snapshot tests; as you mentioned, they are better suited for visual outputs. Only downside being the multiple `snapshots` folders.

spenserblack:
> Only downside being the multiple `snapshots` folders.

We could wait and see where https://github.com/mitsuhiko/insta/issues/416 goes 🙂

o2sh:
No need to block the PR for this. We'll reconsider as we add more snapshot tests and await the upstream issue resolution.

Issue #1208, opened 2023-11-12
Make it easier to discover that data languages (JSON, YAML) and prose (Markdown) are not reported by default


spenserblack:
### Summary 💡

I'm not sure what the best way to do this is, but I'm thinking it can be included in the long help message output by `onefetch --help` to at least be a little easier to discover. Perhaps we can include examples of what "data" and "prose" includes.

Or perhaps we can document some FAQs somewhere.

### Motivation 🔦

We've had many users, most recently @fenio in https://github.com/o2sh/onefetch/issues/26#issuecomment-1806747404, get confused when their project's primary language isn't reported.

o2sh:
On the same note, it may be a good idea to show the language type in [onefetch.dev](https://onefetch.dev/) for each entry 🤔

o2sh:
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

PR #1227, opened 2023-12-10
Filter entries by language type in onefetch.dev 

o2sh:
related to #1208 

Implementing a simple filtering option to make language type information more accessible to users.

@spenserblack, I'm not particularly skilled in UI/UX (or Svelte) 😅 . Feel free to be honest and share your thoughts 🙏

### DIFF
--- a/docs/vercel/src/Index.svelte
+++ b/docs/vercel/src/Index.svelte
@@ -145,6 +145,7 @@
     flex-wrap: wrap;
     gap: 10px;
   }
+
   .checkbox-group label {
     text-transform: capitalize;
   }
--- a/docs/vercel/src/Index.svelte
+++ b/docs/vercel/src/Index.svelte
@@ -37,17 +37,13 @@
   });
 
   onMount(async () => {
-    try {
-      const response = await fetch(
-        'https://api.github.com/repos/o2sh/onefetch/releases/latest'
-      );
-      const data = await response.json();
+    const response = await fetch(
+      'https://api.github.com/repos/o2sh/onefetch/releases/latest'
+    );
+    const data = await response.json();
 
-      tagName = data.tag_name;
-      htmlUrl = data.html_url;
-    } catch (error) {
-      console.error('Error:', error);
-    }
+    tagName = data.tag_name;
+    htmlUrl = data.html_url;
   });
 </script>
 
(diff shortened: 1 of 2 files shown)


docs/vercel/src/Index.svelte
```diff
@@ -82,4 +125,32 @@
...
+    gap: 10px;
+  }
+
+  .checkbox-group label {
+    text-transform: capitalize;
+  }
+
+  .show {
+    display: none;
+  }
```
spenserblack, line 155: The logic is a bit reversed here :slightly_smiling_face:

Perhaps rename this class to `.hide`?

docs/vercel/src/Index.svelte
```diff
@@ -61,14 +82,36 @@
...
+    {#each languageTypes as type}
+      <label for={type}>
+        <input
+          id={type}
+          type="checkbox"
+          value={type}
+          bind:group={$filter.checkboxes} />
+        {type}
+      </label>
+    {/each}
```
spenserblack, line 101: TBH I wonder if the filters really need to be shown/hidden, since they don't take up much space.

docs/vercel/src/Index.svelte
```diff
@@ -1,41 +1,62 @@
...
+
+  const filteredLanguages = derived(filter, ($filter) => {
+    let filtered: Language[] = languages;
+    if ($filter.checkboxes.length > 0) {
+      filtered = filtered.filter((d: Language) =>
+        $filter.checkboxes.includes(d.type)
+      );
+    }
+    return filtered;
+  });
```
spenserblack, line 37: I'm more of a Vue person than a Svelte person, but in Vue I would've probably bound the checkbox values to a `{ [key: string]: boolean }` instead of a `string[]`. Then the filtered languages could be `languages.filter(({ type }) => $filter.checkboxes[type])`.

spenserblack (commented):
Looks good! Just some nitpicks mostly.

The one big thing that I'm wondering is if if makes sense to have "programming" and "markup" be checked by default. This could help hint to what `onefetch`'s defaults are. Perhaps a "reset to default" button could help illustrate this? Though that's a bit of overkill for 4 checkboxes :laughing:

I see you've written the checkboxes to be dynamic by utilizing a `Set<string>` instead of hard-coding. But, to set default checked values, I think you can't get around doing a little bit of hard-coding. But, given how infrequently the language types are likely to change, I think it's OK to do a bit of hard-coding 🙂

docs/vercel/src/Index.svelte
o2sh: done

docs/vercel/src/Index.svelte
o2sh: If I do this I won't be able to use the `bind:group`  on the input element easily. I'm also not very familiar with Svelte.

o2sh:
Hi @spenserblack, thanks a lot for the review 👍

> The one big thing that I'm wondering is if if makes sense to have "programming" and "markup" be checked by default. This could help hint to what onefetch's defaults are.

I see your point but I'm just not sure we should restrict the default output of `onefetch.dev` based on constraints specific to the binary 🤔

spenserblack (approved):

PR #1252, opened 2024-01-11
Added the V programming language

gilramot:


spenserblack (changes requested):
[Doesn't look like tokei supports V](https://github.com/XAMPPRocky/tokei/issues/889). And, honestly, I'm not sure if it ever will, since tokei has a "one language per extension" rule that would cause V to conflict with Coq, which uses the `.v` extension. For more context, Verilog, which also apparently uses `.v`, was forced to use a [non-standard extension](https://github.com/XAMPPRocky/tokei/issues/520). So I don't think we can merge this unless there's either a significant change in how tokei works or we change language detection tools :frowning:

Issue #342, opened 2020-12-12
Size: 0 bytes (24 files) ....  wut?


subnut:
```
❯ onefetch --version
onefetch 2.7.3
❯ onefetch
             :::::::::::::                 Subhaditya Nath ~ git version 2.29.2
           ::::::::::::::::::              ------------------------------------
          :::   ::::::::::::::             Project: nvim-ghost.nvim (7 branches, 20 tags)
          ::::::::::::::::::::             HEAD: e6e8042 (main, github/main)
                    ::::::::::             Version: v0.0.20
    :::::::::::::::::::::::::: ______      Created: 10 weeks ago
  :::::::::::::::::::::::::::: ________    Languages: Python (84.2 %) Shell (15.8 %) 
 ::::::::::::::::::::::::::::: _________   Dependencies: 4 (pip)
:::::::::::::::::::::::::::::  _________   Authors: 85% Subhaditya Nath 100
::::::::::::::::::::::::::    __________            6% Subhaditya Nath 8
:::::::::::  ___________________________            6% Restyled.io 8
:::::::::  _____________________________   Last change: 3 days ago
 :::::::: _____________________________    Repo: https://github.com/subnut/nvim-ghost.nvim.git
 :::::::: _____________________________    Commits: 117
   :::::: __________________________       Lines of code: 368
          ____________________             Size: 0 bytes (24 files)
          ____________________             License: MIT
          ______________   ___             
           __________________                                      
              _____________             
```

Using `onefetch-bin` from AUR

subnut:
The two issues I have -

- The obvious one: Zero file size????
- The less obvious one: Why are there two _"Subhaditya Nath"_ in the **Authors:** section?

subnut:
Why is it showing 24 files? I only have 19 files (and 6 directories (well... 19+6=24))
```
.
├── autoload
│   ├── nvim_ghost
│   │   └── installer.vim
│   └── nvim_ghost.vim
├── binary.py
├── LICENSE
├── plugin
│   └── nvim_ghost.vim
├── README.md
├── requirements.txt
└── scripts
    ├── hooks
    │   └── pre-commit
    ├── install_binary.ps1
    ├── install_binary.sh
    ├── install_hooks.sh
    ├── session_closed.sh
    ├── session_closed.vbs
    ├── start_server.sh
    └── start_server.vbs
.github
└── workflows
    ├── auto_release.yaml
    ├── black.yaml
    ├── build_release_binaries.yaml
[...]

Issue #122, opened 2019-10-27
Size: 0 bytes (90 files)


pawamoy:
Well... it's in the title :smile: 

Running it on a local clone of https://github.com/pawamoy/aria2p.

spenserblack:
Size is calculated by the first line returned from the command `git -C <your-repo> count-objects -vH` Could you show us the output of this command? (`-C <your-repo>` can be omitted if you're running git in the repo in question). Thanks!

pawamoy:
Sure!

```console
$ git count-objects -vH 
count: 1988
size: 9.39 MiB
in-pack: 0
packs: 0
size-pack: 0 bytes
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```

spenserblack:
>It seems `onefetch` is picking the `size-pack` value.

Yes, that's correct. Sorry, I was wrong earlier in saying it was the "first line". Was thinking of a different info field 🤦‍♂ 
Could you run this command in your local repo?
```bash
$ find .git/objects -type f
```

pawamoy:
No problem :slightly_smiling_face: 

The command `find .git/objects -type f` outputs 1995, I don't think you want them all, so here are the first 10:

```
.git/objects/6b/89fcce946b4fd5b63b023207ef7961ddf6e1e6
.git/objects/6b/239a70185d47108034b76f68b2e5bb490b90e4
.git/objects/6b/0a477d3794a6f90754ab7c020b4e697360f53b
.git/objects/6b/30fabfc6cd4c573c1fa41b6d33c5c53a6c1e15
.git/objects/6b/753a40d77a91aa25fd4bb518593bceb7011a46
.git/objects/28/216055080981e6c78eb908769f2e08cea2b9c9
.git/objects/28/3e8af7371cdd1c7492d06699cded2aa5824906
.git/objects/db/c62711c906028e1e3fe6a0218906c7afa91aec
.git/objects/db/81b06182ac75f30a3cd6c6620ca1664b9ba882
.git/objects/db/c705553706229ac4888a5623a8b22e186d7992
```

spenserblack:
Try either of these commands.
```bash
$ find .git/objects -type f | grep pack
$ ls -lh .git/objects/pack/
```
What I'm thinking is that there aren't any pack files in your local repo for some reason. I'll need to read up on git's behavior some more to figure out why that would happen.

pawamoy:
Indeed, first command outputs nothing, second command says `total 0`.

[...]

PR #1305, opened 2024-04-08
Switch to gengo

### LINKED ISSUE #26: Improve language detection system to recognize C++ headers

spenserblack:
See #1152 for previous conversation. With the drastic API changes to gengo it started to feel like more work to modify the existing PR instead of starting over.

# To Do

- [ ] Gengo needs to support the remaining languages (or at least have reasons to *not* support them) (https://github.com/spenserblack/gengo/issues/34)
- [ ] Change pretty much everything that previously matched on `Language` and `tokei::LanguageType` enums.

Resolves #26
Closes #1152

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,15 +43,6 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
-dependencies = [
- "memchr",
-]
-
-[[package]]
-name = "aho-corasick"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
@@ -66,34 +57,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
-
-[[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
 
 [[package]]
 name = "anstream"
@@ -178,17 +145,6 @@ dependencies = [
  "serde_json",
  "unicode-normalization",
  "zstd",
-]
-
-[[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
 ]
 
 [[package]]
@@ -408,40 +364,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
 
 [[package]]
-name = "chrono"
-version = "0.4.26"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec837a71355b28f6556dbd569b37b3f363091c0bd4b2e735674521b4c5fd9bc5"
-dependencies = [
- "android-tzdata",
- "iana-time-zone",
- "num-traits",
- "winapi",
-]
-
-[[package]]
-name = "chrono-tz"
-version = "0.8.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1369bc6b9e9a7dfdae2055f6ec151fe9c554a9d23d357c0237cee2e25eaabb7"
-dependencies = [
- "chrono",
- "chrono-tz-build",
- "phf",
-]
-
-[[package]]
-name = "chrono-tz-build"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e2f5ebdc942f57ed96d560a6d1a459bae5851102a25d5bf89dc04ae453e31ecf"
-dependencies = [
- "parse-zoneinfo",
- "phf",
- "phf_codegen",
-]
-
-[[package]]
 name = "ciborium"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


spenserblack:
@o2sh Before this moves any further we need to determine if we want to move forward with analyzing a git ref (`HEAD`) or a folder (I think preferably we don't support both). I've started listing reasons for each here: https://github.com/o2sh/onefetch/discussions/1303

src/info/langs/mod.rs
o2sh, line 1: *Seems like I cannot add a comment on unchanged lines, so I'm putting this here...* 😓 

You need to update `get_loc_by_language_sorted` method name and description

src/info/blob_size.rs
o2sh, line 1: [As discussed in the past](https://github.com/o2sh/onefetch/pull/1152#discussion_r1313829809), I think we can drop this info line and only keep `repo_size`

src/info/langs/mod.rs
```diff
@@ -66,146 +73,13 @@ fn get_statistics(
...
-    let tokei_config = tokei::Config {
-        types: Some(filtered_languages),
-        hidden: Some(include_hidden),
-        ..tokei::Config::default()
-    };
-    let ignored: Vec<&str> = globs_to_exclude.iter().map(AsRef::as_ref).collect();
-    languages.get_statistics(&[&dir], &ignored, &tokei_config);
-    languages
+) -> Result<gengo::Analysis, Box<dyn Error>> {
+    // TODO Determine best way to ignore files (and if that should continue to be handled by onefetch)
```
o2sh, line 77: It would be nice to have the language type filter and glob filtering for a V1

src/info/langs/mod.rs
```diff
@@ -35,28 +41,29 @@ fn sort_by_loc(map: HashMap<Language, usize>) -> Vec<(Language, usize)> {
     vec
 }
 
-fn get_loc_by_language(languages: &tokei::Languages) -> Option<HashMap<Language, usize>> {
-    let mut loc_by_language = HashMap::new();
+fn get_loc_by_language<Any>(languages: Any) -> Option<HashMap<Language, usize>> {
+    todo!("Determine if this is necessary")
```
o2sh, line 45: minor: use the `size_by_language logic here`

src/info/langs/mod.rs
```diff
@@ -66,146 +73,13 @@ fn get_statistics(
...
-        types: Some(filtered_languages),
-        hidden: Some(include_hidden),
-        ..tokei::Config::default()
-    };
-    let ignored: Vec<&str> = globs_to_exclude.iter().map(AsRef::as_ref).collect();
-    languages.get_statistics(&[&dir], &ignored, &tokei_config);
-    languages
+) -> Result<gengo::Analysis, Box<dyn Error>> {
+    // TODO Determine best way to ignore files (and if that should continue to be handled by onefetch)
+    let file_source = Git::new(dir, "HEAD")?;
```
o2sh, line 78: is this going to take into account the pending changes ?

src/info/langs/mod.rs
o2sh: or is this going to be handled via `.gitattributes`?

spenserblack: That depends on the "file source." Right now the `Git` file source is highly opinionated and would use `.gitattributes` for ignoring files, overrides, etc.

******

I plan `Directory` file source being much less opinionated (can't get much more generic than just reading a folder :slightly_smiling_face:). The functionality isn't there yet, but I was considering adding ways to configure excluding/including files. Right now `.gitignore` and `.ignore` files would ignore files via the `ignore` crate's default behavior. I was hoping to discuss the usage sometime, but it would probably either be like this:
```rust
let file_source = Directory::with_config("./", Config { ignored_files: Some(vec![]) });
```
or like this:
```rust
let file_source = Directory::new("./").extend_ignored_files(&[]);
```

src/info/langs/mod.rs
spenserblack: https://github.com/o2sh/onefetch/discussions/1303#discussioncomment-9057210

src/info/langs/mod.rs
spenserblack: By default Gengo marks Programming, Markup, and Query as detectable, and Prose and Data as not detectable. `detectable` is basically a boolean that answers "should this factor in to the stats?" Also, documentation, generated files, and vendored files are not detectable by default. Assuming we use the `Git` file source with `.gitattributes`, this is example usage:
```gitattributes
# Markdown is Prose, but we still want it in the stats
*.md gengo-detectable

# JavaScript is Programming, but we want to exclude it
*.js -gengo-detectable

# The contents of dist/ are not generated, so they will be detectable
dist/* -gengo-generated
# OR
# Even though dist/* is generated, it should still be included in the stats
dist/* gengo-detectable
```

******

Again, the `Git` file source is highly opinionated and tries to behave a lot like github-linguist. The usage with the `Directory` file source is still undecided, and right now it doesn't implement any overrides. Or we could write our own file source (https://github.com/o2sh/onefetch/discussions/1303#discussioncomment-9058916) if we want very specific behavior that wouldn't be suitable for the `gengo` crate.

But, however we implement it, IMO the best way would be to inform gengo what files are and aren't detectable, and then make use of either the `Analysis` (detailed results) or `Summary` (simplified results).

******

tl;dr Yes, gengo would be the one handling included/excluded types, and we'd just pass configuration to it somehow.

src/info/repo_size.rs
o2sh, line 1: why the renaming?

spenserblack: Originally because there were two "size" stats (https://github.com/o2sh/onefetch/pull/1305#discussion_r1557311545). I could revert the rename, but I kind of like it for clarifying that we're getting the size of the *repository,* not the size of the code, which can be confusing (https://github.com/o2sh/onefetch/issues/122, https://github.com/o2sh/onefetch/issues/342, https://github.com/o2sh/onefetch/issues/338). Though perhaps this changed with the migration to gix?

src/info/langs/mod.rs
```diff
@@ -19,44 +22,35 @@ pub fn get_loc_by_language_sorted(
     language_types: &[LanguageType],
     include_hidden: bool,
```
spenserblack, line 23: How do we want to handle including/excluding hidden files? I checked it out and `tokei` basically forwards the configuration to a configuration for `ignore`. `ignore` [has different behavior on Windows](https://github.com/BurntSushi/ripgrep/blob/bb8601b2bafb5e68181cbbb84e6ffa4f7a72bf16/crates/ignore/src/pathutil.rs#L21-L57). As I understand it on Windows it will be hidden if the file's stats say that it's hidden, and on all operating systems it will be hidden if the filename starts with `.`.

`ignore` is much more general-purpose than our usage, as it should work with any directory, not just a repository. So I wonder if we should even handle Windows' special case. Besides adding code complexity, hiding files via Windows file attributes isn't really applicable to git repositories.

src/info/repo_size.rs
o2sh: maybe out of scope for this PR

src/info/langs/mod.rs
```diff
@@ -9,14 +9,14 @@ use strum::IntoEnumIterator;
 
 pub mod language;
 
-pub fn get_main_language(loc_by_language: &[(Language, usize)]) -> Language {
```
spenserblack, line 9: A bit of refactoring is inevitable to make things make more sense when using `gengo` over `tokei`. Stumbled on this. Don't mean to be harsh, but is a function for the sole purpose of indexing into `[0]` really necessary? Especially now that newer versions of Rust have `.first()` and `.get(0)` to *safely* index into collections.

o2sh: Sure, go ahead. 

But for an easier review, we can keep the refactoring/cleanup in a separate PR, so we can focus on the current subject and don't lose track of the important changes.

src/info/langs/mod.rs
spenserblack: Oh, wait, I just realized the Windows file attributes won't apply anyway if we're reading from the commit instead of the directory.

src/info/langs/mod.rs
spenserblack: Refactored it to handle a OOB panic that was introduced (see https://github.com/o2sh/onefetch/pull/1305/commits/82da923be01d27b7eed51e0835e7e239582ba512 for more info).

spenserblack:
So I think that covers pretty much everything. All that's left is to add more language support (https://github.com/spenserblack/gengo/issues/34) and make a new release to add back the supported language variants and then we're good AFAIK.

spenserblack:
Alright, after a long time procrastinating, I implemented the last of onefetch's languages (besides combining Bash, Zsh, and Sh into Shell).

A couple things to note:

1. The MSRV has been bumped
2. Right now this brings in a duplicate version of gix. Right now there's a dependency resolution bug and I didn't want to release the gengo lib with a higher version until that's resolved.

spenserblack:
Bumping gix resulting in compiler errors (new usage). Just a moment.

spenserblack:
After thinking about this a bit, I think an issue with tokei and other line-counting tools is that they specialize in counting lines, and, while language ID is a priority, not the *top* priority. Which leads to e.g. tokei having poor out-of-the-box support for Verilog and V (their `*.v` extensions were already taken). So they're not great for properly identifying a language.

So I'd like to add line counts back at some point, possibly by using gengo as a dependency of some crate and doing something vaguely like
```rust
match Language {
    JupyterNotebook => my_super_special_jupyter_notebook_counter()
    _ => generic_counter(Language) // would use community-provided data similar to tokei's languages.json
}
```

Issue #1306, opened 2024-04-11
Handle future commit datetimes in the future without panicking


MalteT:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

If the commit time is in the future, onefetch crashes.

### Expected behavior 🤔

It should possibly handle the case or fail gracefully. The crash report blames this on me, but my system time is okay and I don't have much influence over my colleague's system time.

### Steps to reproduce 🕹

Have the most recent git commit in the future and run `onefetch`

### Additional context/Screenshots 🔦

```
name = "onefetch"
operating_system = "NixOS 24.5.0 [64-bit]"
crate_version = "2.20.0"
explanation = """
Panic occurred in file 'src/info/utils/mod.rs' at line 35
"""
cause = "Achievement unlocked: time travel! Check your system clock and commit dates."
method = "Panic"
backtrace = """

   0: 0x56049cc6bb63 - core::option::expect_failed::hcbf5a9b7e22cab9f
   1: 0x56049cc93285 - onefetch::info::utils::format_time::he784b32eb1d439c7
   2: 0x56049ccec879 - onefetch::info::InfoBuilder::last_change::h5aac04e9a53655fa
   3: 0x56049ccea471 - onefetch::info::build_info::h6232c8d9e8572b77
   4: 0x56049cc7dded - onefetch::main::hfc8e895e84fac4d0
   5: 0x56049cc85363 - std::sys_common::backtrace::__rust_begin_short_backtrace::he043df6535079d0d
   6: 0x56049cc86c9d - std::rt::lang_start::{{closure}}::h5f878ceb72151ea0
   7: 0x56049d5883d4 - std::panicking::try::h45c8379f4866ecc7
   8: 0x56049d581eb5 - std::rt::lang_start_internal::hf358ddc7f05da9a0
   9: 0x56049cc7e5f5 - main
  10: 0x7f25f061c0ce - __libc_start_call_main
  11: 0x7f25f061c189 - __libc_start_main@GLIBC_2.2.5
  12: 0x56049cc6be25 - _start
  13:        0x0 - <unresolved>"""
```

Thanks for your awesome work!

OsiPog:
I'm that colleague

spenserblack:
But I like that "time travel" achievement :cry: But you're right that we could handle this more gracefully.

[...]

PR #1307, opened 2024-04-11
Use offset when getting time duration

spenserblack:
Related: #1306

### DIFF
--- a/src/info/utils/mod.rs
+++ b/src/info/utils/mod.rs
@@ -35,7 +35,7 @@ fn to_human_time(time: Time) -> String {
 /// Gets the duration between `now` and `time`. Returns `Err` if this cannot be calculated.
 fn diff_gix_time(now: SystemTime, time: Time) -> Result<Duration, String> {
     let since_epoch_duration = now.duration_since(SystemTime::UNIX_EPOCH).unwrap();
-    let ts = Duration::from_secs(match time.seconds.try_into() {
+    let ts = Duration::from_secs(match (time.seconds + i64::from(time.offset)).try_into() {
         Ok(s) => s,
         Err(_) => return Err("<before UNIX epoch>".into()),
     });
--- a/src/info/utils/mod.rs
+++ b/src/info/utils/mod.rs
@@ -24,20 +24,26 @@ where
 }
 
 fn to_human_time(time: Time) -> String {
-    let since_epoch_duration = SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap();
+    let duration = match diff_gix_time(SystemTime::now(), time) {
+        Ok(d) => d,
+        Err(s) => return s,
+    };
+    let ht = HumanTime::from(-(duration.as_secs() as i64));
+    ht.to_string()
+}
 
+/// Gets the duration between `now` and `time`. Returns `Err` if this cannot be calculated.
+fn diff_gix_time(now: SystemTime, time: Time) -> Result<Duration, String> {
+    let since_epoch_duration = now.duration_since(SystemTime::UNIX_EPOCH).unwrap();
     let ts = Duration::from_secs(match time.seconds.try_into() {
         Ok(s) => s,
-        Err(_) => return "<before UNIX epoch>".into(),
+        Err(_) => return Err("<before UNIX epoch>".into()),
     });
     let duration = since_epoch_duration.checked_sub(ts).expect(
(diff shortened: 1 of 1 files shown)


spenserblack:
cc @Byron just confirming that `gix::date::Time.offset` is getting used correctly here.

src/info/utils/mod.rs
```diff
@@ -24,20 +24,26 @@ where
...
+}
 
-    let ts = Duration::from_secs(match time.seconds.try_into() {
+/// Gets the duration between `now` and `time`. Returns `Err` if this cannot be calculated.
+fn diff_gix_time(now: SystemTime, time: Time) -> Result<Duration, String> {
+    let since_epoch_duration = now.duration_since(SystemTime::UNIX_EPOCH).unwrap();
+    let ts = Duration::from_secs(match (time.seconds + i64::from(time.offset)).try_into() {
         Ok(s) => s,
-        Err(_) => return "<before UNIX epoch>".into(),
+        Err(_) => return Err("<before UNIX epoch>".into()),
```
Byron, line 40: `gitoxide` supports times before unix epoch, but of course, there are dates outside of the valid range of `i64`.

src/info/utils/mod.rs
```diff
@@ -24,20 +24,26 @@ where
...
+    };
+    let ht = HumanTime::from(-(duration.as_secs() as i64));
+    ht.to_string()
+}
 
-    let ts = Duration::from_secs(match time.seconds.try_into() {
+/// Gets the duration between `now` and `time`. Returns `Err` if this cannot be calculated.
+fn diff_gix_time(now: SystemTime, time: Time) -> Result<Duration, String> {
+    let since_epoch_duration = now.duration_since(SystemTime::UNIX_EPOCH).unwrap();
+    let ts = Duration::from_secs(match (time.seconds + i64::from(time.offset)).try_into() {
```
Byron, line 38: I think the handling of `time` is correct as it adds `offset` to `seconds`.

src/info/utils/mod.rs
```diff
@@ -120,6 +126,21 @@ mod tests {
...
+        let current_time = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap();
+        let hour_offset: i32 = 1;
+        let offset = 60 * 60 * hour_offset;
+        let also_now = Time::new(
+            (current_time.as_secs() as gix::date::SecondsSinceUnixEpoch) + i64::from(offset),
+            -offset,
+        );
+        let diff = diff_gix_time(SystemTime::now(), also_now).unwrap();
```
Byron, line 140: I wouldn't repeat the `SystemTime::now()` call, but reuse the value from above. This is not deterministic.

Byron (commented):
Thanks for reeling me in.

Generally I recommend adding way more tests, and also see what happens with unusual times which are absolutely possible. Future dates and pre-epoch dates, which I believe definitely shouldn't panic.

spenserblack:
Thanks for the review @Byron!

> Future dates and pre-epoch dates, which I believe definitely shouldn't panic.

This is currently being discussed in #1306, and will probably be a separate PR.

src/info/utils/mod.rs
o2sh, line 38: I'm a bit perplexed. Why do we need to consider the offset here? This method should simply return the difference between two UTC timestamps (without timezone): `now (since epoch in UTC+0) - commit.time (since epoch in UTC+0`.

Sorry if I'm mistaken.

spenserblack: I made this with the assumption that #1306 was caused by differing timezones, which was an incorrect assumption, which made me assume that we were comparing a timezone-aware time to a naive time. Right now, personally, I am not sure if `commit.time.seconds` is UTC+0 or if it's a naive time that needs `offset` to get the real time.

Byron: Sorry, I also didn't catch it. `commit.time.seconds` is UTC. The [doc-string there](https://docs.rs/gix-date/latest/gix_date/struct.Time.html#structfield.seconds) unfortunately only implies it ("time in seconds since epoch."). I have fixed this [here](https://github.com/Byron/gitoxide/blob/554ec1c448cbf8da07d3a24b22696ab1c7c1dcb0/gix-date/src/lib.rs#L26).

spenserblack: Thanks! I guess this PR should be closed, then.

PR #1314, opened 2024-04-22
add cli option --force-url-http for URL display

### LINKED ISSUE #1313: Add CLI option to force URL format to http
### Summary 💡

Would it be useful to be able to choose the format of the URL that's displayed?

By adding the option to always reformat the URL to an http format, it will be clickable in most terminals and allow quick access to the remote repo. Depending on hyperlink support in different terminals, having a separate link with a shorter phrase or symbol could be another cool option.

### Motivation 🔦

I've often wished that I could click the url to view the repository on Github, but I always use ssh urls for the remote which my terminal can't detect.

0spotter0:
resolves #1313 

* Added cli flag --force-url-http
* Added new formatting functions and tests.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -84,6 +84,9 @@ pub struct InfoCliOptions {
     /// Show the email address of each author
     #[arg(long, short = 'E')]
     pub email: bool,
+    /// Always display the repo url as http
+    #[arg(long)]
+    pub force_url_http: bool,
     /// Count hidden files and directories
     #[arg(long)]
     pub include_hidden: bool,
@@ -243,6 +246,7 @@ impl Default for InfoCliOptions {
             no_bots: Option::default(),
             no_merges: Default::default(),
             email: Default::default(),
+            force_url_http: Default::default(),
             include_hidden: Default::default(),
             r#type: vec![LanguageType::Programming, LanguageType::Markup],
             disabled_fields: Vec::default(),
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -84,9 +84,9 @@ pub struct InfoCliOptions {
     /// Show the email address of each author
     #[arg(long, short = 'E')]
     pub email: bool,
-    /// Always display the repo url as http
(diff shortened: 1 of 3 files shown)


spenserblack:
I wonder if normalizing to HTTP should be the default behavior. 2 users can clone the exact same repo and get different stats based on their preferred clone strategy. I'm not sure if this is ideal. @o2sh thoughts?

src/cli.rs
```diff
@@ -84,6 +84,9 @@ pub struct InfoCliOptions {
     /// Show the email address of each author
     #[arg(long, short = 'E')]
     pub email: bool,
+    /// Always display the repo url as http
+    #[arg(long)]
+    pub force_url_http: bool,
```
o2sh, line 89:
```suggestion
/// Display repository URL as HTTP
#[arg(long)]
pub http_url: bool,
```

src/info/url.rs
```diff
@@ -36,17 +36,32 @@ pub fn get_repo_url(repo: &Repository) -> String {
...
+    }
+}
+
 fn remove_token_from_url(url: &str) -> String {
     let pattern = Regex::new(r"(https?://)([^@]+@)").unwrap();
     let replaced_url = pattern.replace(url, "$1").to_string();
     replaced_url
 }
 
+fn create_http_url(url: &str) -> String {
```
o2sh, line 59:
```suggestion
fn create_http_url_from_ssh(url: &str) -> String {
```

o2sh, line 59: maybe add a unit test specifically for this method

o2sh:
>I wonder if normalizing to HTTP should be the default behavior. 2 users can clone the exact same repo and get different stats based on their preferred clone strategy. I'm not sure if this is ideal. @o2sh thoughts?

I believe Onefetch should display the Git information verbatim unless asked otherwise. Knowing which protocol is used to sync with the remote can be important IMO.

o2sh (approved):
Approved 🎉

PR #1319, opened 2024-04-28
Add CLI flag to hide token from repository URL

o2sh:
Following https://github.com/o2sh/onefetch/pull/1314#issuecomment-2078259382

This PR introduces a CLI flag to enable the option of hiding the token from the repository URL, making it an opt-in mechanism rather than the default behavior.

CC @jim4067 @asciimoth @t-h2o

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -87,6 +87,9 @@ pub struct InfoCliOptions {
     /// Display repository URL as HTTP
     #[arg(long)]
     pub http_url: bool,
+    /// Hide token in repository URL
+    #[arg(long)]
+    pub hide_token: bool,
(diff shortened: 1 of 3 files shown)


spenserblack (approved):
Code looks good.

But I believe that one big draw of "\*fetch" tools is to share info with other people. Given that, I'm not sure that displaying a secret value by default is preferable default behavior TBH.

o2sh:
Yeah, it's a trade-off between security and transparency.

In my opinion, transparency combined with user control is the best option.

Transparency:  ensuring users see the raw data without any modifications.
User Control:  allowing users to opt-in to hide tokens if they wish, rather than enforcing it by default.

Issue #1334, opened 2024-05-04
Churn includes bot commits when `--no-bots` option is provided


o2sh:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

In this example the churn pool matches the total number of commits in the repository (2432)  instead of only accounting for non bot commits (1860)

<img width="752" alt="Screenshot 2024-05-05 at 00 20 01" src="https://github.com/o2sh/onefetch/assets/13710835/11a83fb9-00a9-4dbf-bfb5-96c1b7c95606">

### Expected behavior 🤔

exclude bot commits from churn when `no-bots` option is used

PR #1335, opened 2024-05-04
exclude bot commits from churn when `--no-bots` option is used

o2sh:
Fix for #1334


src/info/git/mod.rs
```diff
@@ -172,20 +174,29 @@ type ChurnPair = (NumberOfCommitsByFilepath, usize);
 
 fn get_churn_channel(
     repo: &gix::Repository,
+    mailmap: &gix::mailmap::Snapshot,
+    bot_regex_pattern: &Option<MyRegex>,
```
Byron, line 178: Passing `&Option<T>` isn't idiomatic, but `Option<&T>` is. That requires `.as_ref()` on the call-site. This is likely the work of an IDE and refactoring tools.

It's a minor detail though and one doesn't have to care, it also seems to be done in other places that aren't touched here.

Byron (commented):
LGTM!

src/info/git/mod.rs
spenserblack: Is there perhaps a blanket `AsRef` implementation that can be used? Something like this?
```rust
bot_regex_pattern: impl AsRef<Option<MyRegex>>
```
(also perhaps `MyRegex` should implement `AsRef<Regex>`)

spenserblack (approved):

src/info/git/mod.rs
Byron: I don't think that adding generics to method signatures would do more than allowing to avoid manual `.as_ref()`, which I'd be more than happy to do for clear compiler errors.
But in the end, it's probably Ok the way it is as it's most natural for IDEs. It's surprising clippy is happy with it though.

spenserblack: TBH looking at this more it is kind of odd IMO to have functions that don't do anything significant unless that option is `Some`.
Basically instead of this
```rust
fn foo(ref: &Option<T1>) -> T2 {
    if bar(ref) {
        do_thing()
    } else {
        do_other_thing()
    }
}

fn bar(ref: &Option<T1>) -> bool {
    let thing = do_calculation();
    baz(thing, ref)
}

fn baz(thing: Thing, ref: &Option<T1>) -> bool {
    ref.as_ref().map_or(false, |t| t.process(thing))
}
```

I think it should be more like this:
```rust
fn foo(ref: &Option<T1>) -> T2 {
    if ref.as_ref().unwrap_or(false, bar) {
        do_thing()
    } else {
        do_other_thing()
    }
}

fn bar(ref: &T1) -> bool {
    let thing = do_calculation();
    ref.process(thing)
}
```

So, like you said, move the `.as_ref` call higher up in the "layers" of functions, but while we're at it these function calls could probably use some other cleanup. I think this non-idiomatic usage is partially caused by an overly complex flow of functions getting called.

o2sh: @Byron @spenserblack [1b682e9](https://github.com/o2sh/onefetch/pull/1335/commits/1b682e91aceb699e7e44fc739cc4e0fa8e486262) Did I do it right?
There seems to be a little more cloning involved.

src/info/git/mod.rs
```diff
@@ -172,20 +174,29 @@ type ChurnPair = (NumberOfCommitsByFilepath, usize);
...
+        let mailmap = mailmap.clone();
+        let bot_regex_pattern = bot_regex_pattern.clone();
         let has_commit_graph_traversal_ended = has_commit_graph_traversal_ended.clone();
         let total_number_of_commits = total_number_of_commits.clone();
         move || -> Result<_> {
             let mut number_of_commits_by_file_path = NumberOfCommitsByFilepath::new();
             let mut number_of_diffs_computed = 0;
             while let Ok(commit_id) = rx.recv() {
                 let commit = repo.find_object(commit_id)?.into_commit();
+                if bot_regex_pattern.is_some()
```
Byron, line 195: This extra check shouldn't be needed if `is_bot_commit()` would exit early. I agree that all the hassle might go away if `is_bot_commit()` would take the regex without an `Option` right away.

src/info/git/mod.rs
o2sh: As a second follow up https://github.com/o2sh/onefetch/pull/1340

src/info/git/mod.rs
o2sh: Done, moved the check inside `is_bot_commit`

src/info/git/mod.rs
```diff
@@ -286,10 +286,10 @@ fn compute_diff_with_parent(
     Ok(())
 }
 
-fn get_no_bots_regex(no_bots: &Option<Option<MyRegex>>) -> Result<Option<MyRegex>> {
-    let reg = if let Some(r) = no_bots.clone() {
+fn get_no_bots_regex(no_bots: Option<&Option<MyRegex>>) -> Result<Option<MyRegex>> {
```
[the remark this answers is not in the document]
o2sh: no more Option<Option...

cf. https://github.com/o2sh/onefetch/pull/1340

spenserblack (approved):

Issue #1363, opened 2024-07-10
LICENSE.md in published crates contains symlink path instead of license text 


musicinmybrain:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

In `onefetch-ascii`, `onefetch-image`, and `onefetch-manifest`, the `LICENSE.md` files in the published crates contain the text

```
../LICENSE.md
```

(with no trailing newline), rather than the expected license text.

### Expected behavior 🤔

In `onefetch-ascii`, `onefetch-image`, and `onefetch-manifest`, the `LICENSE.md` files in the published crates *should* contain the text from https://github.com/o2sh/onefetch/blob/main/LICENSE.md.

### Steps to reproduce 🕹

Download the crates from crates.io, e.g. https://crates.io/api/v1/crates/onefetch-ascii/2.21.0/download, or try `cargo publish --dry-run`, and examine `LICENSE.md`.

### Additional context/Screenshots 🔦

Currently, these crates appear to be proper symbolic links to the top-level license file in the git repository; this is standard practice, and `cargo publish` normally dereferences these. Indeed, in a fresh checkout on Linux, `cd ascii; cargo publish --dry-run; $ tar -tzvf ../target/package/onefetch-ascii-2.21.0.crate` shows that `LICENSE.md` has the expected size for the license text.

I conjecture that the crates may be published from a Windows environment that does not fully support symbolic links, as described in https://github.com/rust-lang/cargo/issues/5664.

This issue was detected in a [`rust-onefetch-manifest` package review for Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=2293787).

### Possible Solution 💡

I suppose the only solutions are to ensure crates are published from an appropriate platform, or to replace the symbolic links in the git repository with duplicate copies of the license text. If duplicate license text is what you want, I can offer a PR.

spenserblack:
> I suppose the only solutions are to ensure crates are published from an appropriate platform

I personally believe the best method is to automate publish via CD so we can have reproducible publish environments. We stopped doing that in 5085c5b038c46ecb9fb16329caf5479b35a963c6. Comments on that commit say it was because there were issues with publishing the workspace. Should be [pretty easy to publish the workspace crates](https://github.com/spenserblack/gengo/blob/d40fd5c2c8f641a8d2a8e793c6145049e932ea75/.github/workflows/release.yml#L17-L26), though.

[...]

PR #1364, opened 2024-07-11
Automate publishing crates to crates.io

musicinmybrain:
This mostly reverts 5085c5b038c46ecb9fb16329caf5479b35a963c6, and attempts to support publishing workspace crates `onefetch-ascii`, `onefetch-image`, and `onefetch-manifest`. This would fix https://github.com/o2sh/onefetch/issues/1363 by ensuring crates are published on a platform that fully supports symbolic links.

I have little to no experience with GitHub Workflows, and I am unable to test this PR, but it still *seems* like it should do the job.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -61,3 +61,24 @@ jobs:
           installers-regex: '\.exe$'
           version: ${{ github.event.release.tag_name }}
           token: ${{ secrets.WINGET_TOKEN }}
+
+  cargo-publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          components: clippy
+
+      - name: Publish to crates.io
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        run: cargo publish
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -81,23 +81,17 @@ jobs:
       - name: Publish to crates.io (ascii)
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        working-directory:
-          ./ascii
-        run: cargo publish
+        run: cargo publish -p onefetch-ascii
 
       - name: Publish to crates.io (image)
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        working-directory:
-          ./image
-        run: cargo publish
+        run: cargo publish -p onefetch-image
 
       - name: Publish to crates.io (manifest)
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        working-directory:
-          ./manifest
-        run: cargo publish
+        run: cargo publish -p onefetch-manifest
 
       - name: Publish to crates.io
         env:
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -78,6 +78,27 @@ jobs:
           profile: minimal
           components: clippy
 
+      - name: Publish to crates.io (ascii)
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        working-directory:
+          ./ascii
+        run: cargo publish
+
+      - name: Publish to crates.io (image)
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        working-directory:
+          ./image
+        run: cargo publish
+
+      - name: Publish to crates.io (manifest)
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        working-directory:
+          ./manifest
+        run: cargo publish
+
       - name: Publish to crates.io
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -69,7 +69,7 @@ jobs:
       CARGO_TERM_COLOR: always
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -59,5 +59,46 @@ jobs:
         with:
           identifier: o2sh.onefetch
           installers-regex: '\.exe$'
-          version: ${{ github.event.release.tag_name }}
```
spenserblack, line 62: Why was this line removed?

spenserblack (changes requested):
This looks mostly good. Could you use `cargo publish -p PACKAGE-NAME` instead of setting the working directory? I know I suggested changing the working directories in https://github.com/o2sh/onefetch/commit/5085c5b038c46ecb9fb16329caf5479b35a963c6, but I think using `cargo publish -p PACKAGE-NAME` is slightly cleaner. As an added bonus one could just copy+paste the steps and run them from the project's root to publish manually if we avoid modifying the working directory.

spenserblack:
> I am unable to test this PR

Honestly when it comes to situations like this I think it's fine to basically test in production. In other words, just trigger the workflow when we're ready and expect it to work. If there's anything to fix we can make a new patch release to re-trigger the workflow. I've even seen big projects like Vue make new patches just to re-trigger a previously broken CD.

Though of course we would *never* merge in a broken CD workflow :wink:

.github/workflows/cd.yml
```diff
@@ -59,5 +59,46 @@ jobs:
...
           token: ${{ secrets.WINGET_TOKEN }}
+
+  cargo-publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+
+    steps:
+      - uses: actions/checkout@v3
```
spenserblack, line 71:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/cd.yml
```diff
@@ -59,5 +59,46 @@ jobs:
...
           identifier: o2sh.onefetch
           installers-regex: '\.exe$'
-          version: ${{ github.event.release.tag_name }}
           token: ${{ secrets.WINGET_TOKEN }}
+
+  cargo-publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
```
spenserblack, line 68: I think that this environment variable can actually be set on the workflow level (root level), not the job level. That way it will be set for all jobs, which should help on build steps for other jobs.

.github/workflows/cd.yml
```diff
@@ -59,5 +59,46 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          components: clippy
+
```
spenserblack, line 79: All of `actions-rs` is deprecated right now. TBH I'm wondering if we actually need a step to install Rust here. GitHub Actions have Rust and `cargo` installed by default, and typically a pretty recent version AFAIK. I never bothered to explicitly install the Rust toolchain in some personal projects, and I've never had a problem.

spenserblack (changes requested):
Sorry, went through this again and had a few more comments.

.github/workflows/cd.yml
musicinmybrain: It was added in 5085c5b038c46ecb9fb16329caf5479b35a963c6, and then removed again when I reverted that commit. Now, my intention was to revert everything *except the addition of this line*, which is why the commit message says “Revert "fix cd" (but keep the added version)”, but obviously I missed actually fixing up the commit. I’ll correct it. Thanks for catching this.

.github/workflows/cd.yml
musicinmybrain: I’ll give it a try! This is how it was done in the “old” implementation, but that doesn’t make it optimal. Documentation supports your claim: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env

.github/workflows/cd.yml
musicinmybrain: 👍

.github/workflows/cd.yml
musicinmybrain: Interesting. Looking more closely, the existing `deploy` job has the following “Install Rust” step:

https://github.com/o2sh/onefetch/blob/ac7e2b8e3313c70f7e7960aefa1296e302ac19e6/.github/workflows/cd.yml#L17-L21

I suppose it makes sense to either match that in `cargo-publish`, or remove *both* “Install Rust” steps. What do you think?

musicinmybrain:
> This looks mostly good. Could you use `cargo publish -p PACKAGE-NAME` instead of setting the working directory? I know I suggested changing the working directories in [5085c5b](https://github.com/o2sh/onefetch/commit/5085c5b038c46ecb9fb16329caf5479b35a963c6), but I think using `cargo publish -p PACKAGE-NAME` is slightly cleaner. As an added bonus one could just copy+paste the steps and run them from the project's root to publish manually if we avoid modifying the working directory.

I’ll follow up with this change next. `$ cargo publish -p onefetch-ascii --dry-run` works, so it should be fine.

musicinmybrain:
I think I handled all the feedback except for https://github.com/o2sh/onefetch/pull/1364#discussion_r1674351277. Let me know what you think about https://github.com/o2sh/onefetch/pull/1364#discussion_r1674414547, and I’ll be back to this in a few hours.

.github/workflows/cd.yml
spenserblack: Let's add dtolnay's action for consistency.

spenserblack:
> I think I handled all the feedback except for [#1364 (comment)](https://github.com/o2sh/onefetch/pull/1364#discussion_r1674351277). Let me know what you think about [#1364 (comment)](https://github.com/o2sh/onefetch/pull/1364#discussion_r1674414547), and I’ll be back to this in a few hours.

Great! Once you resolve https://github.com/o2sh/onefetch/pull/1364#discussion_r1674431906 this will be good to me.

musicinmybrain:
> > I think I handled all the feedback except for [#1364 (comment)](https://github.com/o2sh/onefetch/pull/1364#discussion_r1674351277). Let me know what you think about [#1364 (comment)](https://github.com/o2sh/onefetch/pull/1364#discussion_r1674414547), and I’ll be back to this in a few hours.
> 
> Great! Once you resolve [#1364 (comment)](https://github.com/o2sh/onefetch/pull/1364#discussion_r1674431906) this will be good to me.

Done! Thank you for the thorough and helpful review.

spenserblack (approved):
Looks good! Congrats on getting some experience with GH Workflows!

o2sh (approved):

PR #1375, opened 2024-07-22
Drop unused shebangs from repo test fixture scripts

musicinmybrain:
These are not really doing any harm, but they are not doing any good either, and are perhaps slightly misleading:

- Since the scripts do not have executable filesystem permissions, the shebang (`#!`) line is not used
- For non-executable scripts like this, `gix-testtool` always executes them with `bash`: https://github.com/Byron/gitoxide/blob/a807dd1ffb05efd177700d065095249e6c4b3c68/tests/tools/src/lib.rs#L555-L560

An alternative would be to `chmod a+x tests/fixtures/*.sh` so that the shebang lines *can* be useful – and *are* used, if I read the `gix-testtools` code correctly.

### DIFF
--- a/tests/fixtures/make_bare_repo.sh
+++ b/tests/fixtures/make_bare_repo.sh
@@ -1,4 +1,3 @@
-#!/bin/sh
 set -eu -o pipefail
 
 git init -q --bare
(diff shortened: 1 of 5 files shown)


spenserblack (approved):
Besides running the script, shebangs can be useful as hints for language detection (editors, GitHub stats). But it shouldn't matter since they all have `.sh` extensions.

PR #1376, opened 2024-07-22
Re-generate the man page with --no-info

musicinmybrain:
Re-generate the man page with

```
help2man --no-info --name='Command-line Git information tool' 
```

(adding the `--no-info` option)

so that it does not direct users to a nonexistent Texinfo manual.


o2sh (approved):
Thanks a lot @musicinmybrain

PR #1389, opened 2024-08-08
show future commit dates without panicking

### LINKED ISSUE #1306: Handle future commit datetimes in the future without panicking

MalteT:
Fixes #1306.

Sorry for the long wait! This is the lamest solution I found, the tool will simply display the time as `in 5 days`. The library `time_humanize` can already handle this.

Thanks!

### DIFF
--- a/src/info/utils/mod.rs
+++ b/src/info/utils/mod.rs
@@ -103,24 +103,22 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(
-        expected = "Achievement unlocked: time travel! Check your system clock and commit dates."
-    )]
-    fn should_panic_when_display_human_time_and_commit_date_in_the_future() {
+    fn handle_display_human_time_and_commit_date_in_the_future() {
         let day = Duration::from_secs(60 * 60 * 24);
         let current_time = SystemTime::now()
             .duration_since(SystemTime::UNIX_EPOCH)
             .unwrap();
         let tomorrow = current_time + day;
         let time = Time::new(tomorrow.as_secs() as gix::date::SecondsSinceUnixEpoch, 0);
-        format_time(time, false);
+        let result = format_time(time, false);
+        assert_eq!(result, "in a day");
     }
 
     #[test]
     fn display_time_before_epoch() {
         let time = Time::new(gix::date::SecondsSinceUnixEpoch::MIN, 0);
         let result = to_human_time(time);
-        assert_eq!(result, "<before UNIX epoch>");
+        assert!(result.ends_with(" years ago"));
     }
 
     #[rstest]
--- a/src/info/utils/mod.rs
+++ b/src/info/utils/mod.rs
@@ -2,7 +2,7 @@ use crate::cli::NumberSeparator;
 use gix::date::Time;
 use num_format::ToFormattedString;
 use owo_colors::{DynColors, Style};
-use std::time::{Duration, SystemTime};
+use std::time::SystemTime;
 use time::{format_description::well_known::Rfc3339, OffsetDateTime};
 use time_humanize::HumanTime;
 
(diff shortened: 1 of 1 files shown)


spenserblack (approved):
The simplest solution is often the best. I think saying that the repo is created "in x days/hours/etc." is an elegant and clear way to call out a time traveler.

Is there an emoji combo that can represent a time traveler? :thinking:

src/info/utils/mod.rs
```diff
@@ -27,16 +27,14 @@ fn to_human_time(time: Time) -> String {
...
-    let duration = since_epoch_duration.checked_sub(ts).expect(
-        "Achievement unlocked: time travel! \
-        Check your system clock and commit dates.",
-    );
-    let ht = HumanTime::from(-(duration.as_secs() as i64));
+    // Calculate the distance from the current time. This handles
+    // future dates gracefully and will simply return something like `in 5 minutes`
+    let duration = (ts.as_secs() as i64) - (since_epoch_duration.as_secs() as i64);
+    let ht = HumanTime::from(duration);
     ht.to_string()
```
o2sh, line 38: Here is a suggestion to simplify the logic and handle commit time before epoch :
```rust
fn to_human_time(time: Time) -> String {
    let since_epoch_duration = SystemTime::now()
        .duration_since(SystemTime::UNIX_EPOCH)
        .expect("System time is before the Unix epoch");

    // Calculate the distance from the current time. This handles
    // future dates gracefully and will simply return something like `in 5 minutes`
    let delta_in_seconds = time.seconds - (since_epoch_duration.as_secs() as i64);
    HumanTime::from(delta_in_seconds).to_string()
}
```
It will also require to update the unit test `display_time_before_epoch` to avoid an underflow

MalteT:
> The simplest solution is often the best. I think saying that the repo is created "in x days/hours/etc." is an elegant and clear way to call out a time traveler.
> 
> Is there an emoji combo that can represent a time traveler? 🤔

🕰️🚀

Maybe those? Did you plan on having something like this?
image

spenserblack:
That emoji combo looks good! I'm not saying we have to add those, I was kind of half-joking. But it could be fun as an easter egg.

o2sh (approved):

o2sh (approved):
I think we can leave out the emoji for now 🤔 

Thanks a lot for you contribution @MalteT and @spenserblack for the review 🎉

MalteT:
> I think we can leave out the emoji for now 🤔

Depending on the requirements (is there a config option for emojis? would we always show the with the message?) I would be willing to open a mini PR to add that, @o2sh 🤔

spenserblack:
> is there a config option for emojis?

Nope, because we don't have any emojis in the output yet.

PR #1444, opened 2024-10-27
docs: polish `README.md` translation

adamperkowski:
Hey, I cooked up a Polish translation.
Feel free to @ me when the original is updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,8 @@
   <a href="docs/README.ja.md">日本語</a> |
   <a href="docs/README.kr.md">한국어</a> |
   <a href="docs/README.ru.md">Русский</a> |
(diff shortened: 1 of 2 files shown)


spenserblack (approved):
LGTM!

So I guess some phrases like "open source" don't have a Polish equivalent and the English is used and understood?

adamperkowski:
> So I guess some phrases like "open source" don't have a Polish equivalent and the English is used and understood?

Short answer: Yup.
Long answer: They kinda have (`otwartoźródłowy`/`otwarte źródło`) (`otwarty` - `open`; `źródło` - `source`) but literally nobody says that. In many cases it's just better to say  (and easier to pronounce) certain words in their original languages. Especially words that are not super common.

Issue #1452, opened 2024-10-29
Script to run onefetch whe navigating to a git repository for Powershell


kiapanahi:
This is a sample script to put in `$PROFILE` similar to the ones for `bash`, `zsh`, or `fish` for Powershell to run `onefetch` (vanilla command) upon navigating to a git repository.

```pwsh
# git repository greeter
$global:lastRepository = $null

function Check-DirectoryForNewRepository {
    $currentRepository = git rev-parse --show-toplevel 2>$null
    if ($currentRepository -and ($currentRepository -ne $global:lastRepository)) {
        onefetch
    }
    $global:lastRepository = $currentRepository
}

function Set-Location {
    param (
        [string]$path
    )

    # Use the default Set-Location to change the directory
    Microsoft.PowerShell.Management\Set-Location -Path $path

    # Check if we are in a new Git repository
    Check-DirectoryForNewRepository
}

# Optional: Check the repository also when opening a shell directly in a repository directory
# Uncomment the following line if desired
#Check-DirectoryForNewRepository
```

spenserblack:
Looks good! If you'd like, you can make a PR into [our wiki](https://github.com/o2sh/onefetch/blob/main/.github/wiki/Getting-Started.md), and it should get deployed once merge in.

kiapanahi:
I didn't quite get your point. Can you elaborate?

spenserblack:
If I understand correctly, @Wu-Felix is saying that this doesn't work with [zoxide](https://github.com/ajeetdsouza/zoxide) or [yazi](https://github.com/sxyazi/yazi).

Yazi is an interactive TUI filesystem manager similar to Ranger, using vim-like controls to navigate directories. TBH IDK if running `onefetch` on changing directories is even appropriate for Yazi. It would require writing a Yazi plugin, most likely.

kiapanahi:
I was gonna suggest the same thing, that @Wu-Felix open a new issue if they deem it necessary.

Thanks for the compliment though. I appreciate it.

PR #1453, opened 2024-10-29
Powershell snippet

kiapanahi:
Addresses #1452

### DIFF
--- a/.github/wiki/Getting-Started.md
+++ b/.github/wiki/Getting-Started.md
@@ -5,8 +5,9 @@ Onefetch is installed, then what?
 ```sh
 > onefetch /path/of/your/repo
  ```
+
  Or
- 
+
 ```sh
 > cd /path/of/your/repo
 > onefetch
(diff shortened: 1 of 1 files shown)


.github/wiki/Getting-Started.md
```diff
@@ -15,38 +16,50 @@ Onefetch is installed, then what?
...
 ```
-By [**@quazar-omega**](https://github.com/quazar-omega)
+
+### Automatic repo detection and running
+
+If you want to automate the detection and running of `onefetch` every time you `cd` into a repository you can leverage one of the methods below:
+
+#### 1. Bash / Zsh
+
+> By [**@quazar-omega**](https://github.com/quazar-omega)
```
spenserblack, line 31: If it's not too much trouble, can you undo these quotes (`> `)? I guess they were added for emphasis. My concern is that they get rendered to `<blockquote>` in HTML, which can convey unintended semantic meaning and be harmful for screen readers.

spenserblack (commented):
Looks great! I appreciate that you went the extra mile and added some more description to the scripts.

.github/wiki/Getting-Started.md
kiapanahi: sure. good point

spenserblack (approved):
Looks like this was formatted with a tool like `prettier`? Don't worry about it for this PR, but for future reference it can be helpful to make formatting commits separate from any other changes. This allows formatting commits to be registered in a `.git-blame-ignore-revs` file ([docs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)). `.git-blame-ignore-revs` is especially helpful for ignoring automated formatting so you can focus on "real" changes in a blame view. But you don't want to add a commit to `.git-blame-ignore-revs` if it also contains good, non-formatting changes :slightly_smiling_face:

spenserblack:
@o2sh Because the CI runs are required before merge, and they only get run on code changes, I can't merge documentation PRs :sweat_smile: Would you want to merge this?

PR #1468, opened 2024-11-14
Add -L/--logo-only option

yabobay:
Add the ability to print an ASCII art logo without needing a git repository.

I was kinda thinking of somehow baking this into Printer, but Printer wants an Info and the whole point is to not have an Info.

### DIFF
--- a/docs/onefetch.1
+++ b/docs/onefetch.1
@@ -143,6 +143,10 @@ If set to auto: true color will be enabled if supported by the terminal
 .IP
 [default: auto]
 [possible values: auto, never, always]
+.HP
+\fB\-L\fR, \fB\-\-logo\-only\fR
+.IP
+Only print the ascii art without retrieving any info
 .SS "IMAGE:"
 .HP
 \fB\-i\fR, \fB\-\-image\fR <IMAGE>
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -7,6 +7,7 @@ use clap::builder::TypedValueParser as _;
 use clap::{value_parser, Args, Command, Parser, ValueHint};
 use clap_complete::{generate, Generator, Shell};
 use num_format::CustomFormat;
+use onefetch_ascii::AsciiArt;
 use onefetch_image::ImageProtocol;
 use onefetch_manifest::ManifestType;
 use regex::Regex;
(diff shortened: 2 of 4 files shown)


spenserblack (changes requested):
> I was kinda thinking of somehow baking this into Printer, but Printer wants an Info and the whole point is to not have an Info.

I see 2 ways to go:

1. Build an empty `Info`. See the `InfoBuilder` struct. I think this should be pretty easy.
2. Make `-L` an alias of `--no-title --disabled-fields project head ...`

Do either of these seem reasonable?

spenserblack:
Oh, and before I forget: thanks for your PR! I agree we should have this. It pairs well with `neofetch -L`.

yabobay:
Thanks! So i tried going with building an empty `Info`, and i think it would involve having to give bogus options to `InfoBuilder.build`. They don't get seen by the user, and i think it'd work, but it's kind of inelegant. So maybe `Printer::new` should accept an `Option<Info>`? I think it makes more sense than having a check inside `build_info`, and it keeps the logic of the -L flag inside `main` with the other early-exit options.

spenserblack:
I pushed https://github.com/o2sh/onefetch/pull/1468/commits/aba656db6c5fa8e91226e6bf23ce69bd93369ff6 to show what I meant. The benefit is that language detection still works, so users aren't forced to specify the language (forcing users to use the `-a` option with `-L` isn't great IMO).

The change I pushed isn't complete. It should also

* Be refactored to avoid unused calculations.
* Be formatted.
* Hide the color palette

But how do you feel about continuing with this?

spenserblack:
Got it. Since this tool is so tied to git repos for most of its behavior, I don't know about making it work outside a repo in this case. I'll have to think about that. I could kind of see it automatically skipping all checks if `-a` and `-L` are passed together as an edge-case optimization.

However, have you considered using the [onefetch-ascii crate](https://crates.io/crates/onefetch-ascii)? You could write a small executable that uses that library to print a logo. You could read from our YAML file to populate a list and default colors. Then you can follow pretty much any logic you want for picking a logo (e.g. randomization).

yabobay:
Will probably do that then! And my usecase is basically just "i like looking at pretty colors"

PR #1492, opened 2024-12-19
Add script to preview/validate Nerd Fonts

spenserblack:
Getting these PRs for Nerd Fonts, I'd like to be able to preview them without having to manually look them up. This script will display the Nerd Font glyphs in your terminal as long as you're using a compatible font.

I used Ruby because:

* Built-in YAML support, so no need to install a YAML-parsing dependency (and deal with a package manifest + recommended package manager)
* Some similarities to Rust, like a focus on using closures and "everything is an expression" behavior

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,3 +17,7 @@ indent_style = tab
 [*.json]
 indent_style = space
 indent_size = 2
+
+[*.rb]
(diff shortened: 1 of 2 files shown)


o2sh (approved):
works great!

image

PR #1515, opened 2025-01-20
More idiomatic way to fetch HEAD refs

o2sh:
Tried to simplify the logic by taking advantage of Gitoxide's dedicated APIs.

We only care about the head's local and remote ref.


Byron (approved):
Neat! Definitely also a performance improvements as one won't go through a possibly large amount of refs anymore.
The only refs collected now are the checked out reference, and its remote tracking ref (e.g. `refs/remotes/origin/foo`) if available.

PR #1516, opened 2025-01-24
More idiomatic way to fetch repository remote URL

o2sh:
Tried to simplify the logic to get the repository remote URL

### DIFF
--- a/src/info/url.rs
+++ b/src/info/url.rs
@@ -23,11 +23,11 @@ pub fn get_repo_url(repo: &Repository, hide_token: bool, http_url: bool) -> Resu
         None => return Ok(String::new()),
     };
 
-    if let Some(url) = remote.url(gix::remote::Direction::Push) {
-        Ok(format_url(&url.to_string(), hide_token, http_url))
-    } else {
-        Ok(String::new())
-    }
+    Ok(remote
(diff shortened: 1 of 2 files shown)


src/info/url.rs
```diff
@@ -16,28 +17,16 @@ impl UrlInfo {
...
-        }
-    }
-
-    match remote_url {
-        Some(url) => format_url(&url, hide_token, http_url),
-        None => String::default(),
+    if let Some(url) = remote.url(gix::remote::Direction::Push) {
+        Ok(format_url(&url.to_string(), hide_token, http_url))
+    } else {
+        Ok(String::new())
```
spenserblack, line 29: IDK if you have a preference, but this can also be written with `Option::map` and `Option::unwrap_or_default` I think.

spenserblack (commented):
LGTM! But, just to confirm, we're now changing from always falling back to an empty string to now distinguishing between "no remote found" vs "invalid remote found", right?

spenserblack (approved):
Oops, hit submit before switching to "approve"

Byron (approved):
LGTM!

src/info/url.rs
o2sh: done

PR #1521, opened 2025-01-28
add Razor language support

SrS2225a:
Let me know if you require any changes


spenserblack (approved):

languages.yaml
```diff
@@ -2307,6 +2307,33 @@ Raku:
       - "#FFFFFF"
       - "#00FF39"
     chip: "#0000FB"
+Razor:
+  type: programming
```
o2sh, line 2311:
```suggestion
  type: markup
```

according to linguist: https://github.com/github-linguist/linguist/blob/1647b5be1658fb15a5e09bf89067f79b392301e2/lib/linguist/languages.yml#L2886

languages.yaml
```diff
@@ -2307,6 +2307,33 @@ Raku:
...
+    {0}  @@@@   @@@@@@@@@@@@@@@@@
+    {0}    @@@@      @@@@@         @@
+    {0}      @@@@@@          @@@@@
+    {0}         @@@@@@@@@@@@@@                 
+  colors:
+    ansi:
+      - magenta
+    hex:
+      - "#5c2994"
+    chip: "#A64AC9"
```
o2sh, line 2336:
```suggestion
    chip: "#512be4"
```

according to linguist: https://github.com/github-linguist/linguist/blob/1647b5be1658fb15a5e09bf89067f79b392301e2/lib/linguist/languages.yml#L2887C11-L2887C18

languages.yaml
```diff
@@ -2307,6 +2307,33 @@ Raku:
...
+    {0} @@   @@@@           @@@   @@@@@@@@@@@
+    {0}@@@  @@@@  @@@@@@@   @@@@  @@@@@@@@@@
+    {0}@@@  @@@@  @@@@@@@   @@@@  @@@@@@@@@
+    {0}@@@  @@@@   @@@@@@   @@@@  @@@@@@@
+    {0} @@   @@@@                @@@@@@@
+    {0} @@@   @@@@@@   @@@@@@@@@@@@@@
+    {0}  @@@@   @@@@@@@@@@@@@@@@@
+    {0}    @@@@      @@@@@         @@
+    {0}      @@@@@@          @@@@@
+    {0}         @@@@@@@@@@@@@@
```
o2sh, line 2330:
```suggestion
    {0}                                     @
    {0}                               @    @@
    {0}                            @@@@   @@@@
    {0}           @@@@@@@@@@@@@@@@@@@@  @@@@@@
    {0}       @@@@@@       @@@@@@@@@@@@@@@@@@@
    {0}     @@@               @@@@@@@@@@@@@@@@
    {0}   @@@    @@@@@@@@@@@    @@@@@@@@@@@@@@
    {0}  @@@   @@@@@@@@@@@@@@@   @@@@@@@@@@@@
    {0} @@@   @@@@         @@@@   @@@@@@@@@@
    {0}@@@   @@@   @@@@@@   @@@@   @@@@@@@@
    {0}@@@   @@   @@@@@@@   @@@@   @@@@@@@
    {0}@@@   @@@   @@@@@@   @@@@   @@@@@@
    {0} @@@   @@@           @@    @@@@@
    {0}  @@@   @@@@@    @@      @@@@@
    {0}   @@@   @@@@@@@@@@@@@@@@@@@    @
    {0}    @@@      @@@@@@@@@       @@
    {0}      @@@@@             @@@@@
    {0}         @@@@@@@@@@@@@@@@
```

cleaner design suggestion:

image

o2sh (changes requested):
left a few suggestions

languages.yaml
spenserblack: By that logic we should probably change Vue's type to "markup". It's "markup" in the permalink provided, and Vue and Razor seem to have a vaguely similar at a high level (a combo of HTML and a programming language).

languages.yaml
SrS2225a: I felt that #A64AC9 would be more correct as that is the official color they use from their logo. The other one looks a little too blueish to me. Here is logo for reference

image

SrS2225a: Would #A64AC9 still be acceptable?

languages.yaml
SrS2225a: > By that logic we should probably change Vue's type to "markup". It's "markup" in the permalink provided, and Vue and Razor seem to have a vaguely similar at a high level (a combo of HTML and a programming language).

Thats correct.` .razor` files are indeed a combo of HTML and a programming language. I wasn't sure if it should of went in markup, or programming since technically you can run C# code within the markup file with the `@code` block. https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor has great documentation on what you can do within an `.razor` file

spenserblack: In this situation where it can be either markup or programming, because it's a mix of both, I think you should look at what's most important.

For example, with Vue, a `.vue` file is perfectly valid and normal with only HTML and no JS. But a `.vue` file with only JS and no HTML is uncommon.

SrS2225a: Also, see https://learn.microsoft.com/en-us/aspnet/core/blazor/components/. It goes into more detail about `@code` blocks

o2sh: You're right, I'll fix it

o2sh (approved):
Thanks @SrS2225a 🎉

Issue #1467, opened 2024-11-12
PowerShell snippet to run onefetch on directory change can break tools that wrap `Set-Location` (`cd`)


spenserblack:
> zoixde yazi can not use ,Unable to jump

_Originally posted by @Wu-Felix in https://github.com/o2sh/onefetch/issues/1452#issuecomment-2470462841_

spenserblack:
I think I know what the problem is for Yazi (and likely zoxide, too). Yazi has this recommended wrapper for PowerShell to CD into the last directory used:

```powershell
function y {
    $tmp = [System.IO.Path]::GetTempFileName()
    yazi $args --cwd-file="$tmp"
    $cwd = Get-Content -Path $tmp
    if (-not [String]::IsNullOrEmpty($cwd) -and $cwd -ne $PWD.Path) {
        Set-Location -LiteralPath $cwd
    }
    Remove-Item -Path $tmp
}
```

I don't know much PowerShell, but I'm guessing that @Wu-Felix's issue is caused because your custom `Set-Location` takes `-Path`, but the Yazi wrapper uses `-LiteralPath`. So the `y` wrapper fails to CD ("jump").

Perhaps this can be easily fixed by accepting and passing *any* parameter from the custom `Set-Location` to the builtin `Set-Location` for better compatibility? Here's a simple example:

```powershell
function Set-Location-Proxy {
    # do something
    Set-Location @args
}
```

cc @kiapanahi

I'd happily accept a PR from either of you 🙂

FallenDeity:
@spenserblack I was working with `zoxide` encountered this exact issue this is a script I modified from the original one in the docs 

```pwsh
$global:lastRepository = $null

function Check-DirectoryForNewRepository {
    $currentRepository = git rev-parse --show-toplevel 2>$null
    if ($currentRepository -and ($currentRepository -ne $global:lastRepository)) {
        [Console]::OutputEncoding = [System.Text.Encoding]::UTF8
        onefetch | Write-Host
    }
    $global:lastRepository = $currentRepository
}

function Set-Location {
    param (
        [string]$Path,
        [string]$LiteralPath
    )

    if ($LiteralPath) {
        Microsoft.PowerShell.Management\Set-Location -LiteralPath $LiteralPath
    }
    else {
        Microsoft.PowerShell.Management\Set-Location -Path $Path
    }

    Check-DirectoryForNewRepository
}

Check-DirectoryForNewRepository
```

[...]

PR #1542, opened 2025-03-05
docs: :memo: Update wiki powershell-snippet

FallenDeity:
Update powershell-snippet in wiki to play nice with zoxide, yazi etc

With reference to #1467 @spenserblack

### DIFF
--- a/.github/wiki/Getting-Started.md
+++ b/.github/wiki/Getting-Started.md
@@ -126,32 +126,21 @@ An adaptation of the above snippet suited for `Powershell`. Put this script in t
 
 ```pwsh
 # git repository greeter
+
+# Set the console output encoding to UTF-8, so that special characters are displayed correctly when piping to Write-Host
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
 $global:lastRepository = $null
 
 function Check-DirectoryForNewRepository {
     $currentRepository = git rev-parse --show-toplevel 2>$null
     if ($currentRepository -and ($currentRepository -ne $global:lastRepository)) {
-        if ([Console]::OutputEncoding -ne [System.Text.Encoding]::UTF8) {
-            [Console]::OutputEncoding = [System.Text.Encoding]::UTF8
-        }
         onefetch | Write-Host
     }
     $global:lastRepository = $currentRepository
 }
 
 function Set-Location {
-    param (
-        [string]$Path,
-        [string]$LiteralPath
-    )
-
-    if ($LiteralPath) {
-        Microsoft.PowerShell.Management\Set-Location -LiteralPath $LiteralPath
-    }
-    else {
-        Microsoft.PowerShell.Management\Set-Location -Path $Path
-    }
-
+    Microsoft.PowerShell.Management\Set-Location @args  
     Check-DirectoryForNewRepository
 }
 
--- a/.github/wiki/Getting-Started.md
+++ b/.github/wiki/Getting-Started.md
@@ -131,20 +131,27 @@ $global:lastRepository = $null
 function Check-DirectoryForNewRepository {
     $currentRepository = git rev-parse --show-toplevel 2>$null
     if ($currentRepository -and ($currentRepository -ne $global:lastRepository)) {
-        onefetch
+        if ([Console]::OutputEncoding -ne [System.Text.Encoding]::UTF8) {
+            [Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+        }
+        onefetch | Write-Host
     }
     $global:lastRepository = $currentRepository
 }
 
 function Set-Location {
     param (
-        [string]$path
+        [string]$Path,
+        [string]$LiteralPath
     )
 
-    # Use the default Set-Location to change the directory
-    Microsoft.PowerShell.Management\Set-Location -Path $path
+    if ($LiteralPath) {
+        Microsoft.PowerShell.Management\Set-Location -LiteralPath $LiteralPath
+    }
+    else {
+        Microsoft.PowerShell.Management\Set-Location -Path $Path
+    }
 
-    # Check if we are in a new Git repository
     Check-DirectoryForNewRepository
 }
 


.github/wiki/Getting-Started.md
```diff
@@ -131,20 +131,27 @@ $global:lastRepository = $null
 function Check-DirectoryForNewRepository {
     $currentRepository = git rev-parse --show-toplevel 2>$null
     if ($currentRepository -and ($currentRepository -ne $global:lastRepository)) {
-        onefetch
+        if ([Console]::OutputEncoding -ne [System.Text.Encoding]::UTF8) {
+            [Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+        }
```
spenserblack, line 136: Is it perhaps too opinionated to set the output encoding? Is there an issue that this resolves?

.github/wiki/Getting-Started.md
```diff
@@ -131,20 +131,27 @@ $global:lastRepository = $null
...
     )
 
-    # Use the default Set-Location to change the directory
-    Microsoft.PowerShell.Management\Set-Location -Path $path
+    if ($LiteralPath) {
+        Microsoft.PowerShell.Management\Set-Location -LiteralPath $LiteralPath
+    }
+    else {
+        Microsoft.PowerShell.Management\Set-Location -Path $Path
+    }
```
spenserblack, line 153: Instead of explicitly capturing certain known arguments, perhaps it would be better to just pass *all* arguments to the built-in. Look up `@args`.

spenserblack (commented):
Thanks! Let me know what you think of this feedback.

.github/wiki/Getting-Started.md
FallenDeity: Yep, so basically as I mentioned if don't pipe to `Write-Host` nothing prints via onefetch if i `z` into a git repo, but if I don't send the console encoding some symbols get garbled, for example the circles, also just added that if statement to avoid setting the encoding multiple times redundantly

image

.github/wiki/Getting-Started.md
FallenDeity: Just this directly?

```pwsh
Microsoft.PowerShell.Management\Set-Location @args
```

.github/wiki/Getting-Started.md
spenserblack: Oh, makes sense. Perhaps we could recommend users set their encoding in their `$PROFILE`,  instead of in this function? Seems like something that belongs in startup configuration. What do you think?

.github/wiki/Getting-Started.md
spenserblack: Yeah, I think so, unless you see any problems with that. It should pass *every* parameter to the builtin AFAIK.

spenserblack: In Python terms, it's the difference between doing this:
```python
def my_print(value, end='\n'):
    print(value, end=end)
```

and this:

```python
def my_print(*args, **kwargs):
    print(*args, **kwargs)
```

AFAIK

FallenDeity: Yep, tested it out with different cases, didn't cause any issues, we can go ahead with `@args`

.github/wiki/Getting-Started.md
FallenDeity: How about we add a comment regarding this, or just it globally above this line?

```pwsh
$global:lastRepository = $null
```

spenserblack: Yeah, setting it globally makes sense, and maybe a comment explaining why they need to do that.

spenserblack:
BTW, I don't have `zoxide` installed, would you mind posting the output of `zoxide init powershell`? I'd just like to look into the hook and maybe figure out why `z` requires the output encoding workaround.

FallenDeity:
I ran a few variations of the experiment so basically if I do `z` into some git repo while the function gets triggered onefetch doesn't display anything in console unless, I explicitly pipe it to `Write-Host` but if I do this `Set-Location -LiteralPath ".\VsCodeProjects\profile\"` it works as expected without piping.

image

image

This leads me to believe this silencing behavior might have something to do with zoxide's command when changing directories.

```
 zoxide init powershell
# =============================================================================
#
# Utility functions for zoxide.
#

# Call zoxide binary, returning the output as UTF-8.
function global:__zoxide_bin {
    $encoding = [Console]::OutputEncoding
    try {
        [Console]::OutputEncoding = [System.Text.Utf8Encoding]::new()
        $result = zoxide @args
        return $result
    } finally {
        [Console]::OutputEncoding = $encoding
    }
}

# pwd based on zoxide's format.
function global:__zoxide_pwd {
    $cwd = Get-Location
    if ($cwd.Provider.Name -eq "FileSystem") {
        $cwd.ProviderPath
    }
}

# cd + custom logic based on the value of _ZO_ECHO.
function global:__zoxide_cd($dir, $literal) {
    $dir = if ($literal) {
        Set-Location -LiteralPath $dir -Passthru -ErrorAction Stop
    } else {
        if ($dir -eq '-' -and ($PSVersionTable.PSVersion -lt 6.1)) {
            Write-Error "cd - is not supported below PowerShell 6.1. Please upgrade your version of PowerShell."
        }
        elseif ($dir -eq '+' -and ($PSVersionTable.PSVersion -lt 6.2)) {
            Write-Error "cd + is not supported below PowerShell 6.2. Please upgrade your version of PowerShell."
        }
        else {
            Set-Location -Path $dir -Passthru -ErrorAction Stop
        }
    }
}

# =============================================================================
#
# Hook configuration for zoxide.
#

# Hook to add new entries to the database.
$global:__zoxide_oldpwd = __zoxide_pwd
function global:__zoxide_hook {
    $result = __zoxide_pwd
    if ($result -ne $global:__zoxide_oldpwd) {
        if ($null -ne $result) {
            zoxide add -- $result
        }
        $global:__zoxide_oldpwd = $result
    }
}

# Initialize hook.
$global:__zoxide_hooked = (Get-Variable __zoxide_hooked -ErrorAction SilentlyContinue -ValueOnly)
if ($global:__zoxide_hooked -ne 1) {
    $global:__zoxide_hooked = 1
    $global:__zoxide_prompt_old = $function:prompt

    function global:prompt {
        if ($null -ne $__zoxide_prompt_old) {
            & $__zoxide_prompt_old
        }
        $null = __zoxide_hook
    }
}

# =============================================================================
#
# When using zoxide with --no-cmd, alias these internal functions as desired.
#

# Jump to a directory using only keywords.
function global:__zoxide_z {
    if ($args.Length -eq 0) {
        __zoxide_cd ~ $true
    }
    elseif ($args.Length -eq 1 -and ($args[0] -eq '-' -or $args[0] -eq '+')) {
        __zoxide_cd $args[0] $false
    }
    elseif ($args.Length -eq 1 -and (Test-Path $args[0] -PathType Container)) {
        __zoxide_cd $args[0] $true
    }
    else {
        $result = __zoxide_pwd
        if ($null -ne $result) {
            $result = __zoxide_bin query --exclude $result -- @args
        }
        else {
            $result = __zoxide_bin query -- @args
        }
        if ($LASTEXITCODE -eq 0) {
            __zoxide_cd $result $true
        }
    }
}

# Jump to a directory using interactive search.
function global:__zoxide_zi {
    $result = __zoxide_bin query -i -- @args
    if ($LASTEXITCODE -eq 0) {
        __zoxide_cd $result $true
    }
}

# =============================================================================
#
# Commands for zoxide. Disable these using --no-cmd.
#

Set-Alias -Name z -Value __zoxide_z -Option AllScope -Scope Global -Force
Set-Alias -Name zi -Value __zoxide_zi -Option AllScope -Scope Global -Force

# =============================================================================
#
# To initialize zoxide, add this to your configuration (find it by running
# `echo $profile` in PowerShell):
#
# Invoke-Expression (& { (zoxide init powershell | Out-String) })
```

FallenDeity:
Edit: I played around with zoxide's script a bit and fixed both issues

- Reason for no output was because they were reassigning any output from Set-Location to a variable `$dir` and not returning it.
```
function global:__zoxide_cd($dir, $literal) {
    $dir = if ($literal) {
        Set-Location -LiteralPath $dir -Passthru -ErrorAction Stop
    } else {
        if ($dir -eq '-' -and ($PSVersionTable.PSVersion -lt 6.1)) {
            Write-Error "cd - is not supported below PowerShell 6.1. Please upgrade your version of PowerShell."
        }
        elseif ($dir -eq '+' -and ($PSVersionTable.PSVersion -lt 6.2)) {
            Write-Error "cd + is not supported below PowerShell 6.2. Please upgrade your version of PowerShell."
        }
        else {
            Set-Location -Path $dir -Passthru -ErrorAction Stop
        }
    }
    #  This fixes it
    return $dir
}
```

- Reason for garbled characters, they are changing the default console output encoding in the finally block back from utf-8
```
function global:__zoxide_bin {
    # default encoding utf-8
    $encoding = [System.Text.Encoding]::UTF8
    try {
        [Console]::OutputEncoding = [System.Text.Utf8Encoding]::new()
        $result = zoxide @args
        return $result
    } finally {
        [Console]::OutputEncoding = $encoding
    }
}
```

onefetch setup

```
$global:lastRepository = $null

function Check-DirectoryForNewRepository {
    $currentRepository = git rev-parse --show-toplevel 2>$null
    if ($currentRepository -and ($currentRepository -ne $global:lastRepository)) {
        onefetch
    }
    $global:lastRepository = $currentRepository
}

function Set-Location {
    Microsoft.PowerShell.Management\Set-Location @args  
    Check-DirectoryForNewRepository
}
```

image

I believe they silence the output because -PassThru they use returns a dictionary which they don't really wanna log, so maybe a sort of configurable option there would be nice

So to summarize the current stuff not printing is more of a zoxide issue 😔 and the this pr fixes the path not changing issue due to other required params not being propagated through the `Set-Location`, and adds a workaround to prevent unintentional silencing by other scripts

spenserblack (approved):
> I believe they silence the output because -PassThru they use returns a dictionary which they don't really wanna log, so maybe a sort of configurable option there would be nice

Without looking more into it, I'm not really sure why zoxide is using `-PassThru` at all. If they don't want output, [simply excluding `-PassThru` can achieve that](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-location?view=powershell-7.5&WT.mc_id=ps-gethelp#-passthru). Discarding the output of `-PassThru` is like doing `some-tool --verbose > /dev/null` :sweat_smile:

Maybe `-PassThru` behaves differently on older versions of PowerShell :shrug:

spenserblack:
:thinking: Also, as a workaround when `onefetch` is wrapped by a tool/function/alias that might consume output (like `z`), it might make sense to redirect onefetch's output to stderr. This is a somewhat common practice to separate text meant for humans from text meant for machines.

FallenDeity:
I don't think it would be a good idea to redirect onefetch's output to stderr in case of powershell
```pwsh
onefetch | Write-Error
onefetch | Write-Error : TSXTSXTSXTSXTSXTSXTSXTSXTSXTSXTSXTSXTSX   FallenDeity ~ git version 2.48.1.windows.1
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException

onefetch | Write-Error : TSXTSXTSXTSXTSXTSXTSXTSXTSXTSXTSXTSXTSX   ------------------------------------------
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException
 ```
 
 wouldnt want this 😭, also afaik piping to stderr might mess with onefetch styles since a lot of terminals/or terminal config frameworks apply custom ansi styles to error text for theming

spenserblack:
@o2sh This will need an admin to merge because the required CI jobs won't run with the ignored paths.

PR #1481, opened 2024-12-06
Update cargo_toml from 0.20.5 to 0.21.0

musicinmybrain:
This crate does not maintain a changelog, but here is the source diff:

https://gitlab.com/lib.rs/cargo_toml/-/compare/v0.20.5...ab52d4f02ed72452a742264e5f877b8373f6b6b9

I verified that `cargo test --workspace` still passes.

PR #1545, opened 2025-03-11
Update cargo_toml from 0.21.0 to 0.22.1

musicinmybrain:
This crate does not maintain a changelog, but here is the source diff:

https://gitlab.com/lib.rs/cargo_toml/-/compare/ab52d4f02ed72452a742264e5f877b8373f6b6b9...v0.22.1

I verified that `cargo test --workspace` still passes.

I’m patching `rust-cargo-manifest` in Fedora to allow the new `cargo_toml` version, and it’s my habit and Fedora’s policy to send patches upstream wherever possible. Because this is not a particularly urgent update, I waited for the weekly batch of PR’s from dependabot (as suggested in https://github.com/o2sh/onefetch/pull/1481), but there wasn’t an PR for `cargo_toml` in today’s batch. I’m not sure why.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,9 +331,9 @@ checksum = "325918d6fe32f23b19878fe4b34794ae41fc19ddbe53b10571a4874d44ffd39b"
 
 [[package]]
 name = "cargo_toml"
-version = "0.21.0"
+version = "0.22.1"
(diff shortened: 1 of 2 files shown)


o2sh (approved):
Thanks @musicinmybrain

PR #1548, opened 2025-03-25
Toml impl

Sk7Str1p3:
Very WIP around #745
@spenserblack I wish you make some review on code. I'm kinda newbie with rust, this code may seems ugly asf so please help me write it cool. 
Code is uncompilable for now (as i had no time to impl some stuff yet) and all changes are under one commit(ignore flake lol) , I will refactor commits later. Waiting for your advices


spenserblack:
> Code is uncompilable for now (as i had no time to impl some stuff yet) and all changes are under one commit(ignore flake lol)

Did you forget to commit your changes to the Rust code? I'm only seeing changes related to flake.

I'm not familiar with flake at all, and only played around with Nix a little bit, but if the addition of a `flake.nix` file can be helpful for users or developers somehow, that can be added in a different PR.

spenserblack:
> About rust code, yes it seems like I simply forgot to push it lol. I will push asap

OK, no problem!

I think it might be helpful to take a test-driven development approach for this one. Write the tests for decoding the config (and possibly combining it with the CLI options) *before* the actual implementation. So it would look vaguely like this:

```rust
fn decode_from_toml(toml: &str) -> Config {
    todo!() // This panics, but lets the code compile so tests can run
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_decode_from_toml() {
        let config = decode_from_toml("types = [\"language\"]");
        assert_eq!(config.types, ["language"]);
}
```

This is really messy, but I hope this at least gives you a general idea of how you can write this PR in a test-driven way.

Sk7Str1p3:
@spenserblack I pushed changes. Code is compileable, don't look at CI complaints I forgot to remove few imports

So, I took previous pull request as base, this is a really raw draft I have work a lot on. Any advice and criticism is appreciated

src/cli.rs
```diff
@@ -186,59 +213,71 @@ pub struct TextForamttingCliOptions {
...
     #[arg(long, short)]
+    #[merge(strategy = merge::bool::overwrite_false)]
     pub package_managers: bool,
 }
 
 impl Default for CliOptions {
     fn default() -> CliOptions {
         CliOptions {
             input: PathBuf::from("."),
+            config_path: Default::default(),
```
spenserblack, line 280:
```suggestion
            config_path: None,
```
Nitpick: I think an explicit `None` is a bit better, here.

src/configuration.rs
```diff
@@ -0,0 +1,46 @@
...
+                Err(anyhow!("Configuration file at {:?} does not exist!", default_path))
+            }
+        }
+    }
+}
+
+pub fn write_default_cfg<P: AsRef<Path>>(default_path: &P) {
+    let toml = toml::to_string(&CliOptions::default()).expect("Config should be serializable");
+    fs::write(default_path, toml).expect("Unable to write file");
+}
```
spenserblack, line 36: Looks mostly good, but I would move these to be implemented on an appropriate struct.
```rust
impl Config {
    fn load_from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
        // ...
    }
}
```

src/configuration.rs
```diff
@@ -0,0 +1,46 @@
...
+            if default_path.exists() {
+                read_cfg_file(&default_path)
+            } else {
+                create_dir(&default_path)?;
+                write_default_cfg(&default_path);
+                Err(anyhow!("Configuration file at {:?} does not exist!", default_path))
+            }
+        }
+    }
+}
```
spenserblack, line 31: IMO this can just take a `P` instead of an `Option<P>`. Any unwrapping can happen before this is called, which can be simplified with:

```rust
path.unwrap_or_else(|| default_config_dir())
```

We may also want to replace some of these `unwrap`s and `expect`s with non-panicking `Result::Err` returns, instead.

spenserblack (commented):
This is starting to look good!

BTW, you might want to install EditorConfig on whatever editor you're using, because I'm noticing a lot of files that don't have final newlines.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
+/.direnv
```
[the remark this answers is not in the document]
Sk7Str1p3: got you, i will create separate PR for this

Sk7Str1p3:
Why would it close bruh

spenserblack:
> Why would it close bruh

Looks like you force-pushed all the way back to b37959c0ae1767e298d65c27a41978012354bb88. The PR auto-closed because there were no changes to merge after that force-push.

Sk7Str1p3:
I pushed changes but it stacked... I'll just open another one

PR #1549, opened 2025-03-27
Flake devShell

Sk7Str1p3:
little flake for development shell with nix

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,36 @@
+{
+  description = "onefetch";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+  };
+
+  outputs = {
+    self,
+    nixpkgs,
+    ...
+  }: let
+  forAllSystems = fn: nixpkgs.lib.genAttrs [
+    "aarch64-darwin"
+    "aarch64-linux"
+    "i686-linux"
+    "x86_64-linux"
+    "x86_64-darwin"
+  ] (system: fn nixpkgs.legacyPackages.${system});
+  in {
+    devShells = forAllSystems (pkgs: {
+      default = pkgs.mkShell {
+        name = "onefetch";
+        packages = with pkgs; [
+          cargo
+          rustc
+          clippy
+          rustfmt
+          rust-analyzer
+          cmake
+        ];
+        RUST_SRC_PATH = "${pkgs.rust.packages.stable.rustPlatform.rustLibSrc}";
+      };
+    });
+  };
+}
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,6 +18,10 @@ indent_style = tab
 indent_style = space
 indent_size = 2
 
+[*.nix]
+indent_style = space
+indent_size = 2
+
 [*.rb]
 indent_style = space
 indent_size = 2
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /stage
 /parts
 /prime
+.direnv
 .gitignore.swp
 .DS_Store
 result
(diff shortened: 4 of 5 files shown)


spenserblack (approved):
Looks good, but I'll read up on flake dev shells before merging.

I think you need to add `.direnv` to the `.gitignore`, don't you?

spenserblack:
Oh, maybe add a `.nix` entry to `.editorconfig` so our editors follow the appropriate standard if we ever update `flake.nix`.

flake.nix
```diff
@@ -0,0 +1,36 @@
+{
+  description = "onefetch";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+  };
+
+  outputs = {
+    nixpkgs,
+    self,
```
spenserblack, line 10:
```suggestion
    self,
    nixpkgs,
```
As I said I'm not too familiar, but based on the docs I'm seeing it seems standard to list `self` first. IDK if position matters here.

spenserblack (approved):
LGTM!

I'll leave this open for a bit for @o2sh's thoughts.

spenserblack (approved):

o2sh (approved):
LGTM! I tested it locally, and it seems to work as expected.

Sk7Str1p3:
@o2sh we forgot something important. flake.lock pinned nixpkgs, so after a week or less commit get outdated and lock should be refreshed. I have no idea how to work with Workflows yet so i leave this to you. Flake.lock should be updated approximately once in 10 days

o2sh:
Good point, @Sk7Str1p3. I've reverted the PR for now.

On second thought, it might not be ideal to commit these files directly to the repo since they’re tied to a specific setup (Nix).

Instead, we could add a section in the Wiki for Nix users, including the `flake.nix` file along with some explanations.

spenserblack:
> @o2sh we forgot something important. flake.lock pinned nixpkgs, so after a week or less commit get outdated and lock should be refreshed. I have no idea how to work with Workflows yet so i leave this to you. Flake.lock should be updated approximately once in 10 days

Could it just be added to the gitignore? We're not like an enterprise team with strict environment requirements, so I think it would be reasonable to just let the dev shell users manage their own lock file.

Sk7Str1p3:
Yep, we could left lock out. Also we could set workflows for updating lock. I found few solutions with that already

Sk7Str1p3:
https://github.com/DeterminateSystems/update-flake-lock this seems to be best solution as it comes from Nix developers

o2sh:
In the meantime, I created a page in the wiki explaining how to setup a local dev environment for onefetch using Nix: https://github.com/o2sh/onefetch/wiki/nix-local-setup

Hope that's ok with you @Sk7Str1p3

Sk7Str1p3:
Yep that's okay

spenserblack:
We could probably add .envrc and flake.nix to the .gitignore so nobody accidentally commits them, then

Sk7Str1p3:
@o2sh im willing come back with much greater and comprehensive nix toolchain. [Crane](https://github.com/ipetkov/crane) is library around nix with devShell, automatic tests generator (not sure about this one?) and incremental builds, it also should support cachix which would not even build crates locally but simply download pre-built crates leading to much faster builds. 
flake.lock won't be a problem because of workflows (renovate/determinate)

Sk7Str1p3:
@o2sh im willing come back with much greater and comprehensive nix toolchain. [Crane](https://github.com/ipetkov/crane) is library around nix with devShell, automatic tests generator and incremental builds, it also should support cachix which would not even build crates locally but simply download pre-built crates leading to much faster builds. 
flake.lock won't be a problem because of workflows (renovate/determinate)

spenserblack:
I think the Helix editor has a flake devshell configured, so we could probably just do what they do

PR #1550, opened 2025-03-28
config impl

Sk7Str1p3:


Sk7Str1p3:
@spenserblack I'm sorry for pinging you that many times... I opened new PR

Current problem is config is not spawning at $XDG_CONFIG_HOME, but it spawns folder. I am not able to check how does config parsed because is not created. Waiting for your answer, thx

src/config.rs
```diff
@@ -0,0 +1,52 @@
...
+pub fn load_cfg<P: AsRef<Path>>(path: Option<&P>) -> Result<CliOptions> {
+    match path {
+        Some(config_path) => read_cfg(config_path),
+        None => {
+            let mut default_path = config_dir().unwrap();
+            default_path.push("/onefetch/config.toml");
+            if default_path.exists() {
+                read_cfg(&default_path)
+            } else {
+                create_dir(&default_path)?;
```
spenserblack, line 28: You probably want to use [`create_dir_all`](https://doc.rust-lang.org/stable/std/fs/fn.create_dir_all.html). But you also want to make sure you're creating `path/to/config/onefetch/` as the directory, *not `path/to/config/onefetch/config.toml/`*. Look into [`Path::parent`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.parent).

spenserblack (changes requested):
See review comment. I think it should help you figure out why the config file isn't being created.

src/config.rs
Sk7Str1p3: This makes sense... I will look into this tomorrow

Sk7Str1p3:
i thing i gotta write a todo with what are my plans on customisation

config sample: `
separator = " -> "
number-separator = comma

# this should support custom ascii and images
# if images not supported will fallback to custom ascii if provided and standard if not
[logo] 
ruby = ./ruby.png
c = ./c.txt
rust = { image = ./rust.png, ascii = ./rust.txt }

# like --disabled-field, but lets you change order
modules = [
  name
  break

  project
  created
  authors
  break

  head
  last-change
  pending
  churn
  lines-of-code
  break

  size
]

# detailed settings on module
name = "____$(whoami) | $(git --version)"
`

Sk7Str1p3:
Nah, we won't get any good customization if we would parse config as CLI arguments. I reverted all commits

I am looking forward to CrabFetch approach. All customisation made by configuration file, and CLI just lets you some minimal settings. 
We can also do things like FastFetch does. It has enormous amount of CLI arguments and uses values in config as arguments. I do NOT like that. 

@spenserblack what's your opinion on that?

spenserblack:
> Nah, we won't get any good customization if we just parse config as CLI arguments.

Could you explain this a bit more? You hint at overriding logos per-language in https://github.com/o2sh/onefetch/pull/1550#issuecomment-2763387426, which wouldn't work if config and CLI are identical. Is that the main issue, or are there others?

> I am looking forward to CrabFetch approach. All customisation made by configuration file, and CLI just lets you some minimal settings.
> We can also do things like FastFetch does. It has enormous amount of CLI arguments and uses values in config as arguments. I do NOT like that.

Each repository is different, and can require different options. For example, if a repository has a logo, show that with `-i`. In many repositories, the user would want to use `-T` to get stats that better represent the "real" dominant language. Frankly, if I had to write a config file to set these, a config file that I may never use again (likely if I'm not a maintainer of the repo I'm analyzing), I would find it a bit annoying. IMO, simply allowing users to store and not repeat some CLI arguments is a big win. But I'm still of the opinion that the design should be CLI first, config file secondary.

spenserblack:
OK, I've just reread the idea that was documented in https://github.com/o2sh/onefetch/issues/745. I had kind of forgotten the original intent of that issue :sweat_smile:

I've noticed that the original issue body was mainly focusing on *display* configuration options. This makes sense to me, since those are the types of things that I'd likely only want to set once, not per-repo. Also, some options would be kind of awkward to set via the CLI, like if I was trying to set Nerd Font icons for the info lines.

So, yeah, I'm backtracking on the config file being a 1:1 with the CLI. Instead, we should focus on a limited set of options that we're very confident would only be set once, regardless of repo.

@o2sh Thoughts?

Sk7Str1p3:
> Instead, we should focus on a limited set of options that we're very confident would only be set once, regardless of repo

Those could be key-value separator, number separator, nerd-fonts switch and natural colors (use colors set by terminal) switch (to be implemented). 

Im working on those and possibly finish in just few hours.

Sk7Str1p3:
Speaking of key-value separator, check this out


This confusing me, why that would happen?

spenserblack:
> This confusing me, why that would happen? Why I could not use -> as separator?

That's going to be caused by clap, the library for argument parsing. It's reading the next argument after `--separator` and seeing `->`. I believe from there it's checking that the next argument starts with `-` (it does), and interpreting that as another CLI switch/option instead of the value passed for `--separator`.

There's perhaps an option that would force clap to take whatever argument is next, regardless of format.

Sk7Str1p3:
AYO I MADE BASE FOR CONFIG!!1! 
currently slowly creating commits

Sk7Str1p3:
@spenserblack I got one annoying error with config file - values MUST be set there, otherwise onefetch is crashing

src/config.rs
Sk7Str1p3: Resolved in new commits

spenserblack:
> @spenserblack I got one annoying error with config file - values MUST be set there, otherwise onefetch is crashing

I think the easiest thing to do would be to just wrap them all in `Option` to make them nullable.

Also, ideally, let's not remove options from the CLI. Here's roughly how shared options should be resolved.

```rust
let opt = cli.opt.or(config.opt).unwrap_or(default_opt);
```

Check out the huge amount of helper methods defined on `Option<T>`.

Sk7Str1p3:
image
why does this freaky ':' appears?

Sk7Str1p3:
@spenserblack could not really find why ":" is hardcoded as separator, do you have any ideas?

spenserblack:
> @spenserblack could not really find why ":" is hardcoded as separator, do you have any ideas?

I don't think there's a particularly strong reason why, just that we hadn't considered that as a customizable option.

I think, before adding new customization features, you should focus on supporting a config file for existing features. Making the label/value separator customizable can then be handled later.

spenserblack:
> 2\. tests failed due to snaps and i have no idea how can i update them (skill issue basically)

Look up the CLI for insta (snapshot testing tool). https://insta.rs/docs/cli/

src/config/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+            separator: Some(":".to_string()),
+            number_separator: Some(NumberSeparator::Plain),
+        }
+    }
+}
+
+pub fn read_cfg<P: AsRef<Path>>(path: &P) -> Result<Configuration> {
+    let file = File::open(path)?;
+    let mut buf_reader = BufReader::new(file);
+    let mut content = String::new();
```
kakserpom, line 29: fs::read_to_string()

tests/snapshots/repo__repo.snap
```diff
@@ -12,7 +12,9 @@ expression: info
       "ProjectInfo": {
         "repoName": "repo",
         "numberOfBranches": 0,
-        "numberOfTags": 1
+        "numberOfTags": 1,
+        "numberSeparator": "Dot",
+        "separator": "->"
```
spenserblack, line 17: These probably don't need to be in JSON outputs (or other machine-readable outputs).

Sk7Str1p3:
@spenserblack I need some help with Tera for per-language configuration field. It's very first time working with generators and for now I have no clue how could I implement those. Maybe there're already some references could help me? If so, may you show them or at least send link to documentation? thx

tests/snapshots/repo__repo.snap
spenserblack: I think serde has an attribute to tell it to ignore when converting to JSON, YAML, etc.

spenserblack:
> Maybe there're already some references could help me? If so, may you show them or at least send link to documentation?

These are probably what you need:

- https://keats.github.io/tera/docs/#control-structures
- https://keats.github.io/tera/docs/#built-ins

Could you tell me what you're planning to write that involves writing with tera instead of Rust? I don't think you need to unless you're iterating over language entries.

Sk7Str1p3:
Sorry for late answer

> I don't think you need to unless you're iterating over language entries.

That's exactly what I planned to implement. As you could see in my "todo", it contains custom image/ascii logo fields on every language. I don't want to write 100+ identical structures and implementation. I investigate on documentation and onefetch source code.

Sk7Str1p3:
Lol I think i implemented the most useless feature imaginable. Shall I revert commit or better keep it?

Sk7Str1p3:
@spenserblack I really need your review on tests I never worked with them and don't really understand how should I edit them according to changes. Also I feel like I made many shitcode

spenserblack:
> That's exactly what I planned to implement. As you could see in my "todo", it contains custom image/ascii logo fields on every language.

I think it would be wiser to consider this a per-project configuration than a per-language configuration. So, for example, instead of configuring your favorite Rust logo, you would configure your logo for the onefetch repository. But, regardless, we haven't discussed this sort of configuration AFAIK, so I think it's best to save that for later and focus on the basic stuff. You don't need to add all the coolest features all at once. You can add just the simplest config that supports really basic options, and once that's done, we can say "cool, now what's next?"

spenserblack:
> I really need your review on tests. I never worked with them and don't really understand how should I edit them according to changes.

Looks like the CI can't run until merge conflicts on `Cargo.toml` and `Cargo.lock` are resolved. Think you could try resolving those? Merge conflicts are kind of a pain with lock files -- I would recommend using `git checkout --theirs -- Cargo.lock` when resolving merge conflicts, then let `cargo` update it by calling `cargo check`.

At a very quick glance I didn't see anything too bad. Were you concerned about the snapshot test? Or some other tests?

Sk7Str1p3:
It's not a bug but an unimplemented feature, I will workaround it a bit later

spenserblack:
@Sk7Str1p3 Just want to say that the "todo" in your latest changes to the PR body looks perfect :+1:

benches/repo.rs
```diff
@@ -10,10 +10,16 @@ fn bench_repo_info(c: &mut Criterion) {
         input: repo.path().to_path_buf(),
         ..Default::default()
     };
+    let toml = Configuration {
+        separator: Some("->".to_string()),
+        number_separator: Some(onefetch::config::NumberSeparator::Dot),
+        nerd_fonts: Some(false),
+        percent_verbosity: Some(1),
+    };
```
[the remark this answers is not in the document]
Sk7Str1p3: Resolved

Cargo.toml
```diff
@@ -52,6 +53,7 @@ gix-features = { version = "0.41.0", features = ["zlib-ng"] }
 globset = "0.4.16"
 human-panic = "2.0.2"
 image.workspace = true
+merge = "0.1.0"
```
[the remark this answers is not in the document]
Sk7Str1p3: Oh, true, I used to think it's only option to merge params, now I did my own implementation. Will remove with last commit

Sk7Str1p3:
@spenserblack i consider reverting all my commits refactoring this pull request to onefetch reimplementation. I found current code non-extendable and hard to edit, especially BuildInfo struct. Also some stuff could be optimised. 

By the way, i found out new config format, RON. It could be also candidate to be config format, as more native to rust? 

Now, I need you to confirm or deny. Waiting for answer...

spenserblack:
> By the way, i found out new config format, RON. It could be also candidate to be config format, as more native to rust?

I wouldn't really use RON unless you can expect users to almost entirely be Rust developers. I think TOML is a happy medium between "easy to use" and "familiar to the Rust ecosystem."



> I found current code non-extendable and hard to edit, especially BuildInfo struct. Also some stuff could be optimised.

For now, I think just focus on the configuration options that are already easy to work with (so the things that are also configurable via CLI). Adding configuration for more difficult things can then be worked on individually, instead of stressing on how to add a bunch of difficult config options at once.

src/config.rs
```diff
@@ -0,0 +1,35 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::{fs, path::Path};
+
+use crate::info::utils::info_field::InfoType;
+
+#[derive(Serialize, Deserialize)]
+pub struct ConfigOptions {
+    pub disabled_fields: Option<Vec<InfoType>>,
```
spenserblack, line 9: I don't believe this needs to be wrapped on an `Option`. If your goal is to avoid breakage when the user doesn't define this value, there should be serde options to fall back to a default.

Also, since `Vec::default()` will always be empty, for now you could probably just derive `Default`.

src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl Default for ConfigOptions {
+    fn default() -> Self {
+        Self {
+            disabled_fields: Some(vec![]),
+        }
+    }
+}
+
+pub fn read_cfg(path: &Path) -> ConfigOptions {
```
spenserblack, line 20: This should probably return a `Result`. When the value is `Result::Err`, we can fall back to the default config with `unwrap_or_default()`.

spenserblack (changes requested):
You're making good progress! Here's some feedback.

src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+    let contents = fs::read_to_string(path).expect("Failed to read config file!");
+    toml::from_str(&contents).expect("Config could not be deserialized!")
+}
+
+pub fn write_default_cfg(path: &Path) -> Result<()> {
```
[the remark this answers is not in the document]
Sk7Str1p3: Oh really I forgot about struct methods xd

spenserblack:
> @spenserblack i think this is smth odd
> image
> 

That is odd. Try running `cargo clean` to delete build files and rebuild it.

That folder isn't backed up by OneDrive, is it?

Sk7Str1p3:
Didn't help. I think its something about filesystem, will check tomorrow

src/config.rs
Sk7Str1p3: Can't really remember why did I write function in that way, but it cannot return error because of expect()

src/info/mod.rs
```diff
@@ -217,11 +218,15 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
...
+    fn new(cli_options: &CliOptions, config_options: &ConfigOptions) -> Self {
         Self {
             title: None,
             info_fields: Vec::new(),
-            disabled_fields: cli_options.info.disabled_fields.clone(),
+            disabled_fields: cli_options
+                .info
+                .disabled_fields
+                .clone()
+                .unwrap_or(config_options.disabled_fields.clone().unwrap_or_default()),
```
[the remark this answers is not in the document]
spenserblack: Ah, OK, sorry, I didn't read this correctly. I didn't notice that it was *`cli_options.info`* and *`config_options`:* I thought `unwrap_FOO` was being called on the same value twice.

I think this can be simplified a bit by using [`Option::or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.or).

src/config.rs
spenserblack: Yeah, I'm thinking it shouldn't panic via `.expect()`, but instead return a `Result::Err` when either `fs::read_to_string()` or `toml::from_str` fail.

src/main.rs
```diff
@@ -30,7 +31,13 @@ fn main() -> Result<()> {
         return Ok(());
     }
 
-    let info = build_info(&cli_options)?;
+    if cli_options.config.generate_config {
+        return config::write_default_cfg(&cli_options.config.config_path.unwrap_or_default());
```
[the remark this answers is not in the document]
Sk7Str1p3: the `None` value on default PathBuf is placeholder, Im gonna use $XDG_CONFIG_HOME (with fallback to $HOME/.config? I couldn't find any info if $XDG_CONFIG_HOME always exist)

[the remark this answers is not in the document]
spenserblack: Yeah, using the `dirs` crate is best for compatibility

src/config.rs
Sk7Str1p3: It should not panic at all but fallback to defaults IMO. Also I believe we should have `--verbose` cli flag to show errors only then asked.

Sk7Str1p3:
Broken code for your review, I experienced some problems

Sk7Str1p3:
It seems to compile, finally

Sk7Str1p3:
Lmfao it thinks `config.toml` is a directory. I forgot to use `parent()` method, I will fix tomorrow

Sk7Str1p3:
@spenserblack I experienced some troubles with writing config - the default path is EMPTY. I set it in `default()` for `ConfigCliOptions`but then i use `unwrap_or_default()`, path presents empty. 
Second issue is that config is being overwritten, I thought `fs::write` would tell me if file exist

spenserblack:
> @spenserblack I experienced some troubles with writing config - the default path is EMPTY. I set it in `default()` for `ConfigCliOptions`but then i use `unwrap_or_default()`, path presents empty. 
> Second issue is that config is being overwritten, I thought `fs::write` would tell me if file exist

Yeah, you're going to have to combine it with `dirs`. The home/config directory is kind of an edge case, so it might be OK to have a panicking unwrap when `dirs` returns `None`

Sk7Str1p3:
hm @spenserblack something odd happens. it cant unwrap to default because the value is PRESENT, but it is an empty path. i cannot figure out the cause of this problem

src/config.rs
```diff
@@ -0,0 +1,65 @@
...
+        match contents {
+            Err(_) => Self::default(),
+            Ok(contents) => {
+                // I wish to print exact error here, like syntax errors
+                toml::from_str(&contents).unwrap()
+            }
+        }
+    }
+
+    pub fn write_default<P>(path: P) -> Result<()>
```
spenserblack, line 37: If you refactor this to `write(&self, path)`, then we can call `::default().write("path")` instead. A bonus from this is that we're already set up to write *other* configs, for example if we needed to dump the config for debugging.

spenserblack:
> hm @spenserblack something odd happens. it cant unwrap to default because the value is PRESENT, but it is an empty path. i cannot figure out the cause of this problem

:thinking: I'll have to look into where that's happening, but `Option::<T>::default()` should always be `Option::None`.

src/main.rs
```diff
@@ -30,7 +32,15 @@ fn main() -> Result<()> {
         return Ok(());
     }
 
-    let info = build_info(&cli_options)?;
+    if cli_options.config.generate_config {
+        // what the actual FUCK is happening here?
+        // why default path is EMPTY?
+        return ConfigOptions::write_default(&cli_options.config.config_path.unwrap_or_default());
```
spenserblack, line 38: At, here it is, if I understand your problem correctly. Don't do `.unwrap_or_default()`, do `unwrap_or_else(|| dirs::config_path().expect("config path to exist")`.

spenserblack: `Option::<PathBuf>::unwrap_or_default()` is equivalent to
```rust
let path_buf = match config_path {
    Some(p) => p,
    None => PathBuf::default(),
};
```

I'm thinking you were expecting the default `PathBuf` as defined in *your* `Default` implementation, instead of the `PathBuf::default()`.

Sk7Str1p3: exactly what i expected to get, is it supposed to work like that or it is a bug?

src/config.rs
Sk7Str1p3: done;

i also think you found my commits chaotic and weird, i will recommit in more pretty way a bit later

src/main.rs
spenserblack: I'm pretty sure `PathBuf::default()` is supposed to return an empty `PathBuf`, if that's what you're asking.

src/info/mod.rs
```diff
@@ -217,10 +218,12 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
 }
 
 impl InfoBuilder {
-    fn new(cli_options: &CliOptions) -> Self {
+    fn new(cli_options: &CliOptions, config_options: &ConfigOptions) -> Self {
         Self {
             title: None,
             info_fields: Vec::new(),
+            // I have totally NO idea how to properly override values
+            // if cli_options.info.disabled_fields is NOT type of Option<T>
```
Sk7Str1p3, line 226: @spenserblack not sure if i should wrap all types into `Option <T>` as i suppose this would require major changes in code but i cant see another way to override values

spenserblack: Check out these docs for telling serde what the value should default to if it's not available: https://serde.rs/attr-default.html

That way you don't have to wrap everything in `Option` and manually unwrap to a default.

Sk7Str1p3: Hm, yes, clap also provide [`default_value`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_value)

src/main.rs
Sk7Str1p3: No, I mean why does `unwrap_or_default`return `PathBuf::default()`and not `ConfigCliOptions::default().config_path`

src/info/mod.rs
Sk7Str1p3: Tested in last commit. Currently I got vectors concatenation instead of overwrite, and clap throws error if `--disabled-fields` is not set with CLI. I think wrap into `Option` would be better decision

Sk7Str1p3:
@spenserblack I tried to implement parameters override but without wrapping into `Option` this is near to impossible or at least much more complicated than it could be

Sk7Str1p3:
Yep as I expected, wrapping into `Option` is most convenient, logical, **_right_** way to overrite. Currently im trying to write a function which would look like 
```Rust
let disabled_fields = merge_with_cli(/* Value to merge */)
// or maybe CliOptions::merge? 
```
Tho I also must decide whether to merge entire struct or just one it's value... 

Anyway I'm on final stretch and this PR is going to be ready to get merged really soon

Sk7Str1p3:
@o2sh ready to merge (just will have to recommit in prettier way), and I would like to get some advices on my code

SirJiga (commented):
Nothing interesting

Sk7Str1p3:
@spenserblack may I have a review?

spenserblack:
> @spenserblack may I have a review? 

My availability is limited right now, but don't worry, your PRs are still on my radar 🙂

Sk7Str1p3:
Ah alright I'm sorry

Sk7Str1p3:
By the way, speaking of optimisations, I noticed there are many `clone()` methods are used (especially then `cli_options` are being parsed). I think we can easily replace them with references as they have same lifetime. 
I would start working on it after this gets merged

Sk7Str1p3:
Oh it's already 2 months since I opened PR lol

src/main.rs
spenserblack: `config_path()` returns `Option<PathBuf>`. `Option::<T>::unwrap_or_default()` will return either the `t` in `Some(t)` or `T::default()` for `None`. So `Option::<PathBuf>::unwrap_or_default()` can return `PathBuf::default()`. Hope that explains things.

src/cli.rs
```diff
@@ -210,7 +212,19 @@ pub struct VisualsCliOptions {
...
+}
+
+#[derive(Clone, Debug, Args, PartialEq, Eq)]
+#[command(next_help_heading = "CONFIG")]
+pub struct ConfigCliOptions {
+    /// Path to the config file
+    #[arg(long, value_hint = ValueHint::FilePath)]
+    pub config_path: Option<PathBuf>,
+    /// Creates a default config file
+    /// Writes to $XDG_CONFIG_HOME/onefetch/config.toml but can be overridden with --config-path
```
spenserblack, line 225: This is only true for Linux. If you want to get fancy, you can change these docs with conditional compilation. Otherwise, you can just generically say something like `onefetch/config.toml in your home directory`.

src/cli.rs
```diff
@@ -302,6 +317,20 @@ impl Default for ImageCliOptions {
     }
 }
 
+impl Default for ConfigCliOptions {
+    fn default() -> Self {
+        ConfigCliOptions {
+            // Not sure about unwrap
+            config_path: Some(
+                dirs::config_dir()
+                    .expect("Could not find $HOME!")
```
spenserblack, line 326:
```suggestion
                    .expect("OS to have a home directory")
```
`$HOME` isn't used in Windows, and this can only fail on operating systems that don't have standard home paths AFAIK. Also, I switched to the style where `expect` is part of the sentence, though you don't have to follow that.

src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+    cli::{NumberSeparator, When},
+    info::utils::info_field::InfoType,
+};
+
+#[derive(Serialize, Deserialize)]
+pub struct ConfigOptions {
+    // THIS IS JUST A RAW, REALLY WIP CONFIG STRUCTURE
+    #[serde(default)]
+    pub disabled_fields: Vec<InfoType>,
+    // Lol is this really will turn into comment?
```
spenserblack, line 15: `//` for normal comments, `///` for documentation comments. More specifically:

```rust
pub mod foo {
    //! This documents the thing that this comment is *inside.* So this
    //! documents `mod foo`
    /// This documents the thing *underneath* the comment, so this documents
    /// `struct Foo`
    pub struct Foo;
}
```

src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Or maybe this?
+    #[serde(default)]
+    pub no_title: bool,
+    #[serde(default)]
+    pub number_of_authors: usize,
+    #[serde(default)]
+    pub number_of_languages: usize,
+    #[serde(default)]
+    pub number_of_file_churns: usize,
```
spenserblack, line 24: `usize::default()` is `0`, which I don't think we want. There should be a serde option to specify what the default value should be, instead of falling back to `Default::default()`, which I believe this is doing.

src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+            no_bold: false,
+            true_color: When::Auto,
+            nerd_fonts: false,
+        }
+    }
+}
+
+impl ConfigOptions {
+    pub fn read<P>(path: P) -> Self
+    // Is it even should panic?
```
spenserblack, line 62: > Is it even should panic?

Probably not :laughing: At least not *here.*
This can return a `Result<Self, _>`, and then somewhere else we can decide what to do when it returns an `Err(_)`.

src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    pub fn write<P>(self, path: P) -> Result<()>
+    // I believe user would like to generate config with CLI flags
+    // I mean to write disabled_fields with --disabled-fields flag
+    where
+        P: AsRef<Path>,
+    {
+        // I dont think this can panic so i simply unwrapped it
+        let contents = toml::to_string(&self).unwrap();
+        match fs::create_dir_all(&path.as_ref().parent().unwrap_or(Path::new("/"))) {
```
spenserblack, line 84: Not sure how I feel about unwrapping to `/` here. I think it's best to just return an error if `parent()` can't get a parent path.

src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+                Err(e) => {
+                    let path = path.as_ref().display();
+                    eprintln!("Failed to write config file at {path}: {e}")
+                }
+            },
+            Err(e) => {
+                let path = path.as_ref().display();
+                eprintln!("Failed to create config directory {path}: {e}");
+            }
+        }
```
spenserblack, line 99: The logic here is OK, but it can be reorganized so that you don't have to nest `match` expressions.
```rust
fs::create_dir_all(parent)?;
fs::write()?;
```

src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+        // Im not sure it should return simple Ok(())?
+        Ok(())
```
spenserblack, line 101: Returning `Ok(())` is fine.

spenserblack (changes requested):
Almost there!

Besides my review comments, there are also merge conflicts that would block this PR.

src/config.rs
Sk7Str1p3: I wanted to write comments directly to config file and was testing which type would be written. I could find any info about it (maybe I'm bad searcher lol)

Sk7Str1p3: I know about rustbook, this is how I learned rust. And I also acknowledged about different comment types

But the question was, how do I turn `///` doc into comment in `.toml` file.
So this
```
struct foo {
    /// this is commented line
    bar: String
}
```
turns into this
```
# this is a commented line
bar = <default value of bar>
```

src/config.rs
Sk7Str1p3: Wait, wait, wait, doesn't serde fallback to values set in struct's `Default` trait implementation?

src/config.rs
Sk7Str1p3: oh, i made wrong question. Is it even *should* panic, or rather in place return `Self::default()`? this was the quiestion, im sorry for confusing

src/config.rs
Sk7Str1p3: hm, yes, it seems to fallback to `Default::default()`. 

I think i shouldn't switch from `Option<T>` to serde fallbacks.

spenserblack: I believe that serde allows you to specify a function to call to get the default.
https://serde.rs/attr-default.html

PR #1560, opened 2025-04-12
Improve error handling

Mahdiglm:
Adds more context to error messages

### DIFF
--- a/src/info/churn.rs
+++ b/src/info/churn.rs
@@ -74,31 +74,35 @@ fn compute_file_churns(
     globs_to_exclude: &[String],
     number_separator: NumberSeparator,
 ) -> Result<Vec<FileChurn>> {
+    // Build a glob matcher for all the patterns to exclude
     let mut builder = GlobSetBuilder::new();
-    for glob in globs_to_exclude {
-        builder.add(Glob::new(glob)?);
+    for pattern in globs_to_exclude {
+        builder.add(glob::Pattern::new(pattern)?);
     }
-    let glob_set = builder.build()?;
+    let matcher = builder.build().context("Failed to build glob matcher for file exclusions")?;
+
     let mut number_of_commits_by_file_path_sorted = Vec::from_iter(number_of_commits_by_file_path);
+    number_of_commits_by_file_path_sorted.sort_by(|(_, a), (_, b)| b.cmp(a));
 
-    number_of_commits_by_file_path_sorted
-        .sort_by(|(_, a_count), (_, b_count)| b_count.cmp(a_count));
-
-    Ok(number_of_commits_by_file_path_sorted
+    let file_churns = number_of_commits_by_file_path_sorted
         .into_iter()
-        .filter_map(|(file_path, nbr_of_commits)| {
-            if !glob_set.is_match(file_path.to_string()) {
-                Some(FileChurn::new(
-                    file_path.to_string(),
-                    *nbr_of_commits,
-                    number_separator,
-                ))
-            } else {
-                None
+        .filter_map(|(file_path, count)| {
+            let path = std::str::from_utf8(&file_path).ok()?;
+            if matcher.is_match(path) {
+                return None;
             }
+
+            let file_name = path.split('/').last().unwrap_or(path);
+            Some(FileChurn {
+                file_path: file_name.to_string(),
+                nbr_of_commits: *count,
+                number_separator,
+            })
         })
         .take(number_of_file_churns_to_display)
-        .collect())
+        .collect();
+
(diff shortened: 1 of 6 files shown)


src/info/langs/mod.rs
```diff
@@ -22,7 +22,12 @@ pub fn get_loc_by_language_sorted(
     let locs = get_locs(dir, globs_to_exclude, language_types, include_hidden);
 
     let loc_by_language =
-        get_loc_by_language(&locs).context("Could not find any source code in this repository")?;
+        get_loc_by_language(&locs).with_context(|| {
+            format!(
+                "Could not find any source code in this repository at path '{}'. Try using --include-hidden if your files are hidden, or check your exclude patterns.",
```
spenserblack, line 27: The most common point of confusion I've seen is languages that are excluded by default not being shown. For example, JSON, a data language, is excluded by default, so a repo that is only JSON files would have "no source code" unless they use the `-T` option. Perhaps you should mention that instead.

src/info/churn.rs
```diff
@@ -16,6 +16,7 @@ pub struct FileChurn {
 }
 
 impl FileChurn {
+    #[allow(dead_code)]
```
spenserblack, line 19: Why are we allowing dead code here?

spenserblack (changes requested):

Mahdiglm:
hey @spenserblack thank you for your feedback.
About the language detection message, that's a good point about data languages. I was just trying to improve the error message but didn't know that was actually the more common issue users run into. I'd be happy to update it to mention the "-T" option instead of the hidden files thing.
And for the #[allow(dead_code)] I added that because the compiler was complaining about FileChurn::new being unused in the actual codebase. It's only used in the tests (we're using direct struct initialization in compute_file_churns). Just seemed cleaner than rewriting all the tests or changing the implementation. But if you'd prefer a different approach I'm open to it.

I'm actually still learning Rust, so I Thanks for the guidance! Should I update the PR with the language suggestion?

src/info/churn.rs
```diff
@@ -74,31 +74,35 @@ fn compute_file_churns(
...
-                None
+        .filter_map(|(file_path, count)| {
+            let path = std::str::from_utf8(&file_path).ok()?;
+            if matcher.is_match(path) {
+                return None;
             }
+
+            let file_name = path.split('/').last().unwrap_or(path);
+            Some(FileChurn::new(
+                file_name.to_string(),
```
o2sh, line 97: Why did you change `file_path` to `file_name`?

Mahdiglm: Actually, I didn't change the behavior there - the original code was already extracting and storing just the file name. I just switched from direct struct initialization to using the constructor method.
The original code was:

`let file_name = path.split('/').last().unwrap_or(path);
Some(FileChurn {
    file_path: file_name.to_string(),
    nbr_of_commits: *count,
    number_separator,
})`

And I changed it to:

`let file_name = path.split('/').last().unwrap_or(path);
Some(FileChurn::new(
    file_name.to_string(),
    *count,
    number_separator,
))`

The variable name file_name was already there, and it was already being assigned to the file_path field. I just switched to using the constructor to eliminate the "unused method" warning since FileChurn::new() wasn't being used in the main code (only in tests).

o2sh: I'm confused, this is not what I see in the diff.

Mahdiglm:
ah sorry I misclicked start a review button

spenserblack:
While improving the error messages to provide more info is a good idea, I think this PR does too much.
Besides what's mentioned in https://github.com/o2sh/onefetch/pull/1560#discussion_r2041097264, I'm seeing variables and imports being renamed when that's not really relevant to the error handling IMO.

The best PR does *only* what it says, and doesn't include possibly unrelated changes.

spenserblack:
> And for the #[allow(dead_code)] I added that because the compiler was complaining

When the compiler complains, it's best to *handle* the complaint instead of silence it. In this case, if the code is only used in tests, then you can move the implementation to inside the test scope.

spenserblack:
> I see what you mean. Yeah, the PR did end up touching more than just error messages.
> Happy to simplify it - I can revert all the import/variable changes and just focus on adding better error contexts. Would that work better?
> For the dead_code warning, I'll drop that change too. Makes sense to fix it properly rather than suppress it.
> Let me know if you want me to update the PR with just the error message improvements!

This all sounds good, thanks!

Mahdiglm:
> Thanks for the PR @Mahdiglm 🎉

sorry I was kinda busy with my own project totally forgot about the pr

PR #1573, opened 2025-05-04
The Nix pipelines

Sk7Str1p3:
This pull request supersedes #1549 with much better approaches.
Features:
 - Development shell
 - Automatic checks (just ran all checks and it found 3 vulnerabilities in gix, wth?)
 - Incremental builds - download ready crates from cachix/garnix
 - Automatic flake.lock update (not sure if this even works?)
 
 this can also be used in CI for much more powerful checks

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,4 @@
+# Doesn't work in the sandbox
+[yanked]
+enabled = false      # Warn for yanked crates in Cargo.lock (default: true)
+update_index = false # Auto-update the crates.io index (default: true)
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,94 @@
+{
+  "nodes": {
+    "advisory-db": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1746174207,
+        "narHash": "sha256-/ZPWm1dsz3tXREXGQfa/W6UsQ+J3Pvi0A0U1DZVgLqU=",
+        "owner": "rustsec",
+        "repo": "advisory-db",
+        "rev": "4584ad9a5ea16ce196317cf4d3593e974fb4a8a1",
+        "type": "github"
+      },
+      "original": {
+        "owner": "rustsec",
+        "repo": "advisory-db",
+        "type": "github"
+      }
+    },
+    "crane": {
+      "locked": {
+        "lastModified": 1746291859,
+        "narHash": "sha256-DdWJLA+D5tcmrRSg5Y7tp/qWaD05ATI4Z7h22gd1h7Q=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "dfd9a8dfd09db9aad544c4d3b6c47b12562544a5",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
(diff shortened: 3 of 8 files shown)


Sk7Str1p3:
By the way, as I am newbie to rust ecosystem, I suppose I did some bad configuration in toml configs. If you think so, let me know.

.cargo/audit.toml
spenserblack, line 1: > Doesn't work in the sandbox

What happens in the sandbox? Does it fail to send HTTP requests to crates.io?

.envrc
spenserblack, line 1: This doesn't need a `layout` command?

taplo.toml
spenserblack, line 1: Setting up TOML formatting sounds good to me, but TBH I think this can go in a separate PR. That way we can get TOML formatting merged in faster while we still discuss flake.

.envrc
Sk7Str1p3: No, this is how direnv works

flake.lock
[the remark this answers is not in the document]
Sk7Str1p3: Same pros of tracking cargo.lock in repository - to be sure the shell will be SAME, and to be sure it is going to behave same way anywhere.

.cargo/audit.toml
Sk7Str1p3: Did not checked (this is crane generated defaults) but I suppose yes it fails

doprz:
I would also recommend using [nix-direnv](https://github.com/nix-community/nix-direnv)
> A faster, persistent implementation of `direnv`'s `use_nix` and `use_flake`, to replace the built-in one.

Sk7Str1p3:
I think I should recreate flake from scratch. Now I worked much more with crane and I can say it can be better. Much better

.cargo/audit.toml
spenserblack: Got it. Makes sense if this file was generated by some tool.

But I don't think we'd want to disable audits for yanked crates just because a dev tool doesn't work well with it.

flake.lock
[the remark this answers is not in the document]
spenserblack: I suppose the question is: do we *need* that amount of consistency? Or can we trust that the requirements and constraints in `flake.nix` are most likely good enough?

Continuing with the comparison to `Cargo.lock`, there are many times where you *don't* commit that lock file, because you don't need to use exact versions. This is common when authoring libraries, for example.

Sk7Str1p3:
I believe we should include clippy, deny and audit checks in our CI. Also it is possible to build CI with nix (this should speed up checks because of dependency caching)

spenserblack:
@Sk7Str1p3 Try to separate formatting from behavioral changes (features, fixes) in your commits. I.e. formatting should in a different commit from other tasks. The reason is that formatting commits can be added to `.git-blame-ignore-revs`. I'm happy to explain this further if needed 🙂

cardboardcpu (approved):
Maybe it could be interesting to consider [treefmt-nix](https://github.com/numtide/treefmt-nix) since it goes well with flakes.

P.D: Gotta love being able to comfortably build this project from latest git :p

image/src/kitty.rs
```diff
@@ -31,11 +31,7 @@ impl KittyBackend {
 
         // generate red rgba test image
         let mut test_image = Vec::<u8>::with_capacity(32 * 32 * 4);
-        test_image.extend(
-            std::iter::repeat([255, 0, 0, 255].iter())
-                .take(32 * 32)
-                .flatten(),
-        );
+        test_image.extend(std::iter::repeat_n([255, 0, 0, 255].iter(), 32 * 32).flatten());
```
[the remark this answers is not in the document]
Sk7Str1p3: This was also already committed on upstream lol, my fork is simply very outdated 😅

Sk7Str1p3:
@spenserblack will it be okay to have pull requests with same name like 'flake.lock: update'? I (almost) succeeded with workflow but I couldn't came up with idea how should we distinguish them. Do you have any ideas on that?

Sk7Str1p3:
OH I MANAGED TO ADD DATE THERE YAY

spenserblack:
I don't think we should use Nix in the CI. CI should, IMO, be close to a standard user building the project. It's like proving "can this project be built" vs "can this project be built with Nix".

However, a new CI job to validate that the Nix configuration is good might make sense. But I think that would be for a `default.nix`, wouldn't it?

flake.nix
```diff
@@ -0,0 +1,193 @@
...
+    flake-utils.lib.eachDefaultSystem (
+      system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
+
+        inherit (pkgs) lib;
+        craneLib = crane.mkLib pkgs;
+
+        # This filter prevent project from being rebuilded then changing
+        # unrelated files ,e.g. README
```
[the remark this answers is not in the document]
Sk7Str1p3: Yes I'm bad with English, sorry. I'll fix soon

Sk7Str1p3:
Completely forgot about all this lol. I'll fix typos in few minutes

PR #1594, opened 2025-08-05
Update rstest dev-dependency from 0.25 to 0.26

musicinmybrain:
Nothing special here, just bumping the dependency downstream in Fedora’s [`rust-onefetch` package](https://src.fedoraproject.org/rpms/rust-chromaterm) and sending the corresponding change upstream. For whatever reason, it looks like dependabot didn’t propose this promptly.

Nothing in https://github.com/la10736/rstest/blob/v0.26.1/CHANGELOG.md looks like it should affect this project, and `cargo test --workspace` still passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2992,21 +2992,20 @@ dependencies = [
 
 [[package]]
 name = "rstest"
-version = "0.25.0"
+version = "0.26.1"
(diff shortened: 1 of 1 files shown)


o2sh:
done in https://github.com/o2sh/onefetch/pull/1595

PR #1633, opened 2025-11-23
Add colors to --help

starsep:
I used [clap-cargo](https://github.com/crate-ci/clap-cargo) as it's easy to use and provides sensible default styles.
User can disable colors via `NO_COLOR=1` environment variable

Output of `onefetch -h`:
| before | after | `NO_COLOR=1` |
| - | - | - |
| onefetch-before | onefetch-colors | onefetch-no-color |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -440,6 +440,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap-cargo"
+version = "0.15.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d546f0e84ff2bfa4da1ce9b54be42285767ba39c688572ca32412a09a73851e5"
+dependencies = [
+ "anstyle",
+ "clap",
+]
+
+[[package]]
 name = "clap_builder"
 version = "4.5.43"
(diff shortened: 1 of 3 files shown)


spenserblack:
>  I think this could be done without adding one more dependency.

Yeah, looks like the value's type comes from clap-builder. https://docs.rs/clap-cargo/latest/clap_cargo/style/constant.CLAP_STYLING.html

So we could probably define our own if we wanted.

starsep:
I don't know whether it's compatible with i18n PR.

---

Yes, it's possible to do without dependency.
Similar effect can be done with
``` rust
use clap::builder::styling::{AnsiColor, Effects};
use clap::builder::Styles;

// ...

const STYLES: Styles = Styles::styled()
    .header(AnsiColor::Green.on_default().effects(Effects::BOLD))
    .usage(AnsiColor::Green.on_default().effects(Effects::BOLD))
    .literal(AnsiColor::Cyan.on_default().effects(Effects::BOLD))
    .placeholder(AnsiColor::Cyan.on_default());

// ...
#[command(styles = STYLES)]
pub struct CliOptions {
```

I can prepare PR if you prefer this approach.

PR #1646, opened 2025-11-27
No unsafe

Sk7Str1p3:
We all love rust for its safety and soundness. Until now, `onefetch-image` had few unsafe blocks, and i fixed that.


spenserblack (approved):

Sk7Str1p3:
@o2sh shall I add `#![deny(unsafe_code)]` to all onefetch crates?

PR #1736, opened 2026-04-19
Prevent churn calculation from returning an error on partial clones

lawrence3699:
Fixes #1706.

Before this change, `onefetch` could fail on tree-filtered partial clones with:

```text
Error: Failed to traverse Git commit history

Caused by:
    sending on a closed channel
```

The churn worker assumed parent and current commit trees were always present locally. In a `git clone --filter=tree:0` partial clone, those tree objects can be missing, which caused the worker to exit early and the main traversal loop to hit the closed-channel error.

This keeps the rest of the repository info working by skipping churn diffing for commits whose tree objects are not available locally. The regression coverage now extends the existing partial-clone fixture with a `tree:0` variant.

Validation:
- `cargo fmt --all -- --check`
- `cargo test --test repo`
- `cargo test`
- `git clone --filter=tree:0 https://github.com/sharkdp/fd /tmp/onefetch-fd-tree0-fixed && target/debug/onefetch /tmp/onefetch-fd-tree0-fixed`

### DIFF
--- a/src/info/git/mod.rs
+++ b/src/info/git/mod.rs
@@ -180,6 +180,11 @@ fn update_signature_counts(
     Ok(())
 }
 
+
+/// This will get the trees for a commit and its parent and compute the diff. On success, this
+/// returns `Ok(true)`. If the trees are not available due to a partial clone, this will return
+/// `Ok(false)`. Essentially, this function will have successfully determined that there is no diff
+/// to compute. Other issues that prevent calculation will return `Err`. 
 fn compute_diff_with_parent(
     change_map: &mut HashMap<BString, usize>,
     commit: &Commit,
--- a/src/info/git/mod.rs
+++ b/src/info/git/mod.rs
@@ -130,8 +130,9 @@ fn get_churn_channel(
                 if is_bot_commit(&commit, &mailmap, bot_regex_pattern.as_ref())? {
                     continue;
                 }
-                compute_diff_with_parent(&mut number_of_commits_by_file_path, &commit, &repo)?;
-                diffs_computed += 1;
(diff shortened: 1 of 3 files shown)


src/info/git/mod.rs
spenserblack, line 183: Since this now returns a value on success, I wonder if there's a better name to use.

spenserblack (approved):
LGTM! Thanks for the informative PR description.

src/info/git/mod.rs
Byron: To me it still looks meaningful in its new `if compute_diff_with_parent() {}` context, but maybe there is an even better name.

What I'd like to see, more as a nit, is a little doc string that says what each boolean value means and why. Otherwise, it's great that partial clones now work.

PS: If it was me, I'd make the change myself and merge. Otherwise, it's probably fine to merge nonetheless.

spenserblack: I agree with @Byron about the doc comment. That would be better than changing the function name.

@lawrence3699 Could you add a doc comment that explains what it means when the function returns `Ok(true)` or `Ok(false)`? Also, I can't really think of a better function name myself, but if you think of a better name feel free to change it.

Edit: Maybe something like `compute_diff_if_available` would be a bit more descriptive.

spenserblack, line 188:
```suggestion

/// This will get the trees for a commit and its parent and compute the diff. On success, this
/// returns `Ok(true)`. If the trees are not available due to a partial clone, this will return
/// `Ok(false)`. Essentially, this function will have successfully determined that there is no diff
/// to compute. Other issues that prevent calculation will return `Err`. 
fn compute_diff_with_parent(
```

PR #1747, opened 2026-05-11
fix: resolve #1460 — Can't install on Debian 12 - libc version issue

### LINKED ISSUE #1460: Can't install on Debian 12 - libc version issue
### Duplicates

### Current behavior 😯

Installation of the .deb file fails due to a dependency issue. Debian 12 does not currently have required version of libc

```
dpkg -i onefetch_2.22.0_amd64.deb
Selecting previously unselected package onefetch.
(Reading database ... 42987 files and directories currently installed.)
Preparing to unpack onefetch_2.22.0_amd64.deb ...
Unpacking onefetch (2.22.0) ...
dpkg: dependency problems prevent configuration of onefetch:
 onefetch depends on libc6 (>= 2.39); however:
  Version of libc6:amd64 on system is 2.36-9+deb12u8.

dpkg: error processing package onefetch (--install):
 dependency problems - leaving unconfigured
Processing triggers for man-db (2.11.2-2) ...
Errors were encountered while processing:
 onefetch
```

### Expected behavior 🤔

Onefetch should operate on the current version of libc 

### Steps to reproduce 🕹

See current behavior 

_No response_

### Possible Solution 💡

_No response_

chinhkrb113:
## Summary

The Debian package currently depends on `libc6 (>= 2.39)`, which means the `.deb` artifact was built on a system with glibc 2.39, likely `ubuntu-24.04` or another recent runner. Debian package dependency metadata is inferred from the binary’s linked glibc symbols, so building on too-new a distribution makes the generated `.deb` unusable on Debian 12, which ships libc 2.36. The release workflow should build Linux artifacts, especially `.deb` packages, on a stable older Linux image such as `ubuntu-22.04`, or in a Debian 12 container, to keep the minimum libc requirement compatible with Debian 12 and most supported distributions

Fixes #1460

## Changes

- `.github/workflows/cd.yml`
- `tests/cd_workflow.rs` *(new)*

## Why

`.github/workflows/cd.yml`: The Debian package currently depends on `libc6 (>= 2.39)`, which means the `.deb` artifact was built on a system with glibc 2.39, likely `ubuntu-24.04` or another recent runner. Debian package dependency metadata is inferred from the binary’s linked glibc symbols, so building on too-new a distribution makes the generated `.deb` unusable on Debian 12, which ships libc 2.36. The release workflow should build Linux artifacts, especially `.deb` packages, on a stable older Linux image such as `ubuntu-22.04`, or in a Debian 12 container, to keep the minimum libc requirement compatible with Debian 12 and most supported distributions.

### DIFF
--- a/tests/cd_workflow.rs
+++ b/tests/cd_workflow.rs
@@ -1,0 +1,11 @@
+use std::fs;
+
+#[test]
+fn linux_release_uses_ubuntu_22_04() {
+    let workflow = fs::read_to_string(".github/workflows/cd.yml").unwrap();
+
+    assert!(workflow.contains("os: [ubuntu-22.04, macos-latest, windows-latest]"));
+    assert!(workflow.contains("if: matrix.os == 'ubuntu-22.04'"));
+    assert!(workflow.contains("Debian 12-compatible glibc baseline"));
+    assert!(!workflow.contains("if: matrix.os == 'ubuntu-latest'"));
+}
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,7 +10,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        # Pin Linux to ubuntu-22.04 so release binaries keep a Debian 12-compatible glibc baseline.
+        os: [ubuntu-22.04, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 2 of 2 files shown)


tests/cd_workflow.rs
spenserblack, line 1: I've never seen a test written in Rust assert the contents of a workflow, and it looks very unusual to me. I'm assuming that this was the result of an LLM being told to add tests for the changes, which it followed even when the changes don't really need tests.

spenserblack (changes requested):
Given the strange test that was added, I've been led to believe that this was a purely AI agent generated PR, with little to no human validation.

Under the assumption that there was no actual testing, I don't feel confident that this fixes the issue, as I highly doubt that this was actually tested to confirm that it compiles and runs properly on Debian 12.

spenserblack:
I believe that this is an AI-generated PR with little to no human review by the contributor. Per [the AI policy](https://github.com/o2sh/onefetch/blob/main/CONTRIBUTING.md#ai-policy), I will close this PR.

Issue #1743, opened 2026-05-07
The reference at "HEAD" could not be instantiated: Reference name cannot start with a dot


spenserblack:
### Duplicates

- [x] I have searched the existing issues

### Current behavior 😯

> one such weakness of `onefetch` is this:
> 
> ```
> > onefetch
> Error: Failed to traverse Git commit history
> 
> Caused by:
>     0: The reference at "HEAD" could not be instantiated
>     1: The path "refs/heads/.invalid" to a symbolic reference within a ref file is invalid
>     2: Reference name cannot start with a dot
> > git refs migrate --ref-format=files
> > onefetch --no-art -d pending
> groutoutlook ~ git version 2.53.0.windows.3
> -------------------------------------------
> Project: onefetch (1 branch)
> Description: Command-line Git information tool
> HEAD: 8d4e4c8 (main, origin/main)
> Version: 2.27.1
> Created: 21 hours ago
> Languages:
>            ● Rust (97.3 %) ● Shell (2.1 %)
>            ● Makefile (0.4 %) ● Ruby (0.2 %)
> Dependencies: 23 (Cargo)
> Author: 100% dependabot[bot] 1
> Last change: 21 hours ago
> URL: https://github.com/o2sh/onefetch
> Commits: 1 (shallow)
> Churn (1): .gitignore 1
>            …/git/metrics.rs 1
>            …/info/head.rs 1
> Lines of code: 5417
> Size: 2.06 MiB (130 files)
> License: MIT
> # output here.
> ```
>
> Core issue is `onefetch` still a `git` tool and lots of thing dependent on `libgit2` have to finish migrating to `reftable` backend first.

### Expected behavior 🤔

Successful execution

### Steps to reproduce 🕹

N/A

### Additional context/Screenshots 🔦

Bringing over this bug report I got in https://github.com/spenserblack/gengo/issues/680#issuecomment-4396440462

cc @groutoutlook

spenserblack:
@Byron Could this be an issue gix could/should resolve? Looks like this may be related to Windows. Are you familiar with the `refs/heads/.invalid` file? Assuming that's the actual filename and not an example, looks like it could be an intentionally invalid file that git's plumbing would create?

Just wanted to make you aware. IDK if there's anything that actually should be done on gix's side.

Byron:
Git marks `HEAD` with an invalid value when the `reftable` backend is used to fail older clients early.

[...]

PR #1755, opened 2026-05-21
fix: add reftable HEAD migration hint

puneetdixit200:
Related to #1743.

This adds a targeted hint for the reftable placeholder failure discussed in the issue. When the commit traversal error chain shows `HEAD` resolving through `refs/heads/.invalid` or `Reference name cannot start with a dot`, onefetch now keeps the original error chain and adds a concrete recovery step:

```text
git refs migrate --ref-format=files
```

This does not try to implement reftable support in onefetch. It just turns the current low-level `gix` failure into an actionable message while that support is still pending upstream.

Tests cover both the reftable-style error chain and an unrelated error that should not receive the hint.

Checks run locally:

- `rustfmt --check src/info/mod.rs`
- `CARGO_HOME=$PWD/.cargo-home CARGO_TARGET_DIR=$PWD/target cargo test info::tests::test_add_reftable_head_hint --lib`
- `CARGO_HOME=$PWD/.cargo-home CARGO_TARGET_DIR=$PWD/target cargo test test_display_head_refs --lib`
- `CARGO_HOME=$PWD/.cargo-home CARGO_TARGET_DIR=$PWD/target cargo test --test repo test_repo -- --exact`
- `CARGO_HOME=$PWD/.cargo-home CARGO_TARGET_DIR=$PWD/target cargo clippy --lib -- -D warnings`

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -132,6 +132,7 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
         cli_options.info.churn_pool_size,
         cli_options.info.no_merges,
     )
+    .map_err(add_reftable_head_hint)
     .context("Failed to traverse Git commit history")?;
     let manifest = get_manifest(&repo_path)?;
     let repo_url = get_repo_url(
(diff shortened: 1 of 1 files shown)


src/info/mod.rs
```diff
@@ -206,6 +207,31 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
...
+    }
+}
+
+fn has_reftable_invalid_head_error(error: &anyhow::Error) -> bool {
+    let mut has_invalid_head_ref = false;
+    let mut has_invalid_dot_ref = false;
+
+    for cause in error.chain() {
+        let message = cause.to_string();
+        has_invalid_head_ref |=
```
spenserblack, line 226: :thinking: I thought `|=` was only a valid operator for numbers, not `bool`s. Neat.

spenserblack (approved):
I think it would've been better to match on the *gix error,* so that cheap pattern and value matching could be done instead of string matching on error messages.

But overall this is good enough IMO.

spenserblack:
https://github.com/o2sh/onefetch/pull/1853 has addressed this, so closing.

PR #1760, opened 2026-05-27
Use flame NF icon for Mojo

spenserblack:
While previewing the icons, I noticed that Mojo wasn't using a Nerd Font icon, but the emoji 🔥. This PR replaces the emoji with a flame icon (``).

Mojo is an interesting case, because IIRC 🔥 was (and maybe still is) a valid file extension. E.g. `hello-world.🔥`. So we may want to keep the emoji to reflect Mojo's unique quirk.

### DIFF
--- a/languages.yaml
+++ b/languages.yaml
@@ -1816,7 +1816,7 @@ Mojo:
       - "#FF4500"
       - "#FFD700"
     chip: "#FF4C1F"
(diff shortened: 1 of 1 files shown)


o2sh (approved):
Good Catch! It's better to be consistent and only use nerd fonts icon

PR #1829, opened 2026-08-07
fix: strip control characters from info field values

### LINKED ISSUE #1828: Info field values aren't stripped of terminal escape sequences
## Summary

onefetch's "Version" field falls back to a project manifest's (`package.json`/`Cargo.toml`) `version` field when the repository has no git tags, and the "Project" field's repo name falls back to the manifest's `name` field when the repository has no configured remote. Both are parsed with standards-compliant parsers and displayed via `InfoField::style_value` (`src/info/utils/info_field.rs`) with no filtering of control characters. Since a manifest's string field has no character restrictions, and running `onefetch` on a directory is the tool's entire purpose, any escape sequence embedded in these fields is written to the terminal and interpreted live.

## POC

(available upon request)

## Impact

I validated this by creating a git repository with no tags and a `package.json` whose `version` field contained a JSON-escaped OSC title-set sequence, then running `onefetch` on it inside a tmux session and capturing the literal byte stream written to the pty via `tmux pipe-pane` (which taps the raw bytes before tmux's own screen-buffer interpretation) and inspecting it as a hex dump. The captured stream contained the exact injected sequence, unmodified, directly adjacent to the version text, confirming it reaches the terminal without any stripping. I used a benign title-set marker specifically to confirm the sequence reaches output unstripped, rather than a more aggressive payload; the specific range of what an injected OSC/CSI sequence can do depends on the victim's terminal emulator and was not enumerated further.

## Suggested Fix

Strip control characters, particularly the ESC byte, from a field's value before it's styled and written to the terminal. I'd suggest fixing this in `InfoField::style_value`, the single point every field already passes through, rather than only in the version field used to demonstrate this, since the project-name fallback has the same root cause and any future manifest-derived field would otherwise reintroduce the same gap.

carfeii:
Fixes #1828

`Version` and `Project` (repo name) fall back to a `package.json`/`Cargo.toml` manifest field when git tags/remote are absent, and those values were styled and written to the terminal in `InfoField::style_value` with no filtering of control characters. A manifest field has no character restrictions, unlike a git tag/ref, so a crafted manifest could inject terminal escape sequences into onefetch's own output.

## Change

Adds `sanitize_for_display` (`src/info/utils/info_field.rs`), which replaces control characters (other than `\n`, since some fields intentionally span multiple lines) with the Unicode replacement character. Applied inside `style_value`, the single choke point every field's value already passes through, so the fix covers the project-name fallback and any other manifest-derived field, not just the version field used to demonstrate this.

## Testing

- Reproduced the issue with a git repo (no tags) and a `package.json` whose `version` field contained a JSON-escaped OSC title-set escape sequence, confirmed via a `tmux pipe-pane` capture (hex-dumped, not just visually inspected) that the decoded raw escape bytes reached the terminal adjacent to onefetch's own SGR color codes.
- Re-ran the same reproduction against this fix; the injected control bytes are gone (replaced with U+FFFD), and unrelated legitimate color codes nearby are untouched.
- Full test suite (132 tests) passes unchanged.

### DIFF
--- a/src/info/utils/info_field.rs
+++ b/src/info/utils/info_field.rs
@@ -2,6 +2,18 @@ use crate::{info::utils::get_style, ui::text_colors::TextColors};
 use owo_colors::OwoColorize;
 use std::fmt;
 
+/// Replaces control characters (which can include terminal escape sequences)
+/// with the Unicode replacement character, since some field values (a
+/// project manifest's version, name, description, or license, for example)
+/// come from data with no character restrictions and shouldn't be trusted
+/// to display as-is. `\n` is preserved since some fields intentionally span
+/// multiple lines.
+fn sanitize_for_display(s: &str) -> String {
+    s.chars()
+        .map(|c| if c != '\n' && c.is_control() { '\u{FFFD}' } else { c })
+        .collect()
+}
+
 #[typetag::serialize]
 pub trait InfoField {
     fn value(&self) -> String;
@@ -46,8 +58,7 @@ pub trait InfoField {
             return None;
         }
         let style = get_style(false, text_colors.info);
-        let styled_lines: Vec<String> = self
-            .value()
+        let styled_lines: Vec<String> = sanitize_for_display(&self.value())
             .lines()
             .map(|line| format!("{}", line.style(style)))
             .collect();
--- a/src/info/utils/info_field.rs
+++ b/src/info/utils/info_field.rs
@@ -131,4 +131,40 @@ mod test {
         info.write_styled(&mut buffer, false, &colors).unwrap();
         assert_eq!(buffer, "", "It should not write anything");
     }
+
+    #[test]
+    fn test_sanitize_for_display_strips_control_chars() {
+        // ESC ] 0 ; PWNED BEL, an OSC title-set sequence
+        let input = "1.0.0\u{1b}]0;PWNED\u{07}";
+        let sanitized = sanitize_for_display(input);
+        assert_eq!(sanitized, "1.0.0\u{FFFD}]0;PWNED\u{FFFD}");
+        assert!(!sanitized.contains('\u{1b}'));
+        assert!(!sanitized.contains('\u{07}'));
+    }
+
+    #[test]
+    fn test_sanitize_for_display_preserves_newlines() {
+        let input = "line one\nline two";
+        assert_eq!(sanitize_for_display(input), input);
+    }
+
+    #[test]
+    fn test_sanitize_for_display_leaves_normal_text_untouched() {
+        let input = "some normal description, with punctuation! 42";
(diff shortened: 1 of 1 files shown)


src/info/utils/info_field.rs
```diff
@@ -2,6 +2,18 @@ use crate::{info::utils::get_style, ui::text_colors::TextColors};
 use owo_colors::OwoColorize;
 use std::fmt;
 
+/// Replaces control characters (which can include terminal escape sequences)
+/// with the Unicode replacement character, since some field values (a
+/// project manifest's version, name, description, or license, for example)
+/// come from data with no character restrictions and shouldn't be trusted
+/// to display as-is. `\n` is preserved since some fields intentionally span
+/// multiple lines.
+fn sanitize_for_display(s: &str) -> String {
```
o2sh, line 11: can you add some unit tests for this?

spenserblack:
This would only realistically happen with a repository that isn't trustworthy IMO, so I'd argue that this would be a symptom of a much greater problem, which is the user running `onefetch` on unsafe and malicious repositories.

Git has had a feature for repository trust/ownership for a while now. Should we perhaps rely on that to determine if it's safe to read and output repository contents?

Also, rather than *clean* the contents, perhaps it would be better to wrap the string in a result? `Ok(s)`, where `s` is the original string, if it's safe and valid, `Err` otherwise. I think this makes sense since AFAIK control characters would violate most package manifest specs, anyway. It also avoids allocating a new string, so it's a bit more performant.

carfeii:
Added unit tests for `sanitize_for_display` in 4af5377, covering the control-byte-to-replacement-character behavior and confirming existing SGR color codes and normal printable text are left untouched.

On the trust question: Perhaps a repo does not need to already look untrustworthy for this to matter. Running onefetch is often how someone forms that first impression, cloning a new dependency or browsing an unfamiliar repo before deciding whether to trust it. Escape sequences in a manifest field let the repo act on the terminal (change the window title, and depending on the emulator, worse) before that judgment ever gets made. git's safe.directory/ownership check is also a different boundary: it stops git from running config or hooks from a repo you don't own, not display of arbitrary file content like a manifest's version or description field.

On `Result<String, _>` vs cleaning: fair point on the allocation, and I'm not attached to the specific approach as long as the control bytes never reach the terminal. The behavioral difference is what happens when one field has a single bad byte in it: cleaning keeps the rest of the field and swaps out just the bad byte, `Err` would drop the whole field.

spenserblack:
> The behavioral difference is what happens when one field has a single bad byte in it: cleaning keeps the rest of the field and swaps out just the bad byte, Err would drop the whole field.

Unless there's a legitimate reason to have a bad byte, I think it's safe to assume the field is malformed at best, malicious at worst, and is an edge case where it's okay to just drop the whole field.

PR #1842, opened 2026-08-18
Add pyproject.toml manifest support (PEP 621)

### LINKED ISSUE #1590: Add parsing of pyproject.toml
### Summary 💡

UV uses the PEP 518 and PEP 621 specs for its `pyproject.toml`

In `pyproject.toml`,
- The project's description is at `project.description`
- Dependencies are at `project.dependencies`, with dev dependencies at `dependency-groups.dev`
- License is at `project.license.text`
- Version is at `project.version`

However, if you want to add support for poetry,
- The project's description is at `tool.poetry.description`
- Dependencies are at `tool.poetry.dependencies` with dev dependencies at `tool.poetry.group.dev.dependencies` (these are formatted as a dictionary, with a `<dependency> = <version>` arrangement
- License is at `tool.poetry.license`
- Version is at `tool.poetry.version`

There are a couple other key-value pairs available, but these seem to be the main ones.

### Motivation 🔦

Adding support for Python's `pyproject.toml` specification.

I noticed that #345 was opened for the same thing, but it was marked as completed even though I don't see it in the repository.

ChrisJr404:
Closes #1590. This adds `pyproject.toml` to the set of manifests onefetch understands, so Python projects show a dependency count, description, version and license just like Cargo and npm projects already do.

Following the discussion on the issue, I focused on the format specified by the PEPs rather than tool-specific layouts. A new `ManifestType::PyProject` reads the PEP 621 `[project]` table:

- `number_of_dependencies` from `project.dependencies`
- `name` / `version` / `description` from their `project.*` fields
- `license` from `project.license`, handled as either a PEP 639 SPDX string or the older `{ text = "..." }` table. A `{ file = "LICENSE" }` table carries no identifier, so it's left unset and onefetch falls back to detecting the license from the repo as before.

Tool-specific tables such as `[tool.poetry]` use a different shape and are intentionally out of scope here.

Parsing uses `toml`, which was already in the tree via `cargo_toml`, so `Cargo.lock` only gains the existing version as a direct dependency of the manifest crate. In the Dependencies field this reads as e.g. `3 (pyproject.toml)`.

### Tests

- `manifest/tests/pyproject.rs` with a `tests/fixtures/pyproject/` fixture, mirroring the existing cargo/npm integration tests.
- A unit test in `manifest/src/lib.rs` covering the three PEP 621 `license` forms (SPDX string, `{ text }` table, `{ file }` table).
- Verified end to end against a scratch git repo containing a `pyproject.toml`: `onefetch` prints `Dependencies: 3 (pyproject.toml)` and picks up the description, version and license. `cargo fmt --all --check`, `cargo clippy`, and the full test suite pass.

### DIFF
--- a/manifest/Cargo.toml
+++ b/manifest/Cargo.toml
@@ -13,3 +13,4 @@ cargo_toml = "1.0.0"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.150"
 strum = { version = "0.28.0", features = ["derive"] }
+toml = "1.0.0"
--- a/manifest/Cargo.toml
+++ b/manifest/Cargo.toml
@@ -14,3 +14,6 @@ serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.150"
 strum = { version = "0.28.0", features = ["derive"] }
 toml = "1.0.0"
+
+[dev-dependencies]
+rstest = "0.26.1"
--- a/manifest/Cargo.toml
+++ b/manifest/Cargo.toml
@@ -16,4 +16,4 @@ strum = { version = "0.28.0", features = ["derive"] }
 toml = "1.0.0"
 
 [dev-dependencies]
-rstest = "0.26.1"
+rstest.workspace = true
--- a/manifest/src/lib.rs
+++ b/manifest/src/lib.rs
@@ -21,7 +21,6 @@ pub struct Manifest {
 pub enum ManifestType {
     Npm,
     Cargo,
-    #[strum(to_string = "pyproject.toml")]
     PyProject,
 }
 
@@ -108,7 +107,16 @@ struct PyProjectTable {
 }
 
 /// PEP 621 allows `license` to be either an SPDX expression string (PEP 639) or
-/// a table pointing at the license text or a file.
+/// a table with a `text` or `file` key.
+///
+/// - `Spdx` is the PEP 639 form, e.g. `license = "MIT"`.
+/// - `Table.text` is the older PEP 621 form, e.g. `license = { text = "MIT" }`.
+///   Its value is free-form: it is conventionally an SPDX identifier, but the
+///   spec permits any string (for example the full text of a non-standard
+///   license). We surface it verbatim and let the caller decide how to render
+///   it, rather than trying to validate it as SPDX.
+/// - `license = { file = "LICENSE" }` points at a file and carries no
+///   identifier, so it deserializes to `text: None` (the `file` key is ignored).
 #[derive(Deserialize)]
 #[serde(untagged)]
 enum PyProjectLicense {
@@ -127,8 +135,11 @@ fn parse_pyproject_manifest(path: &Path) -> Result<Manifest> {
         .project
         .context("pyproject.toml has no [project] table")?;
 
-    // A `license = { file = "LICENSE" }` table carries no identifier, so leave it
-    // unset and let onefetch fall back to detecting the license from the repo.
+    // The SPDX string and a `{ text = "..." }` table both carry a license
+    // identifier (usually SPDX, though `text` may be any free-form string), so
+    // use them directly. A `{ file = "LICENSE" }` table has no identifier, so
+    // leave it unset and let onefetch fall back to detecting the license from
+    // the repo.
     let license = project.license.and_then(|license| match license {
         PyProjectLicense::Spdx(spdx) => Some(spdx),
         PyProjectLicense::Table { text } => text,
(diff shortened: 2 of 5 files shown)


manifest/src/lib.rs
```diff
@@ -21,6 +21,8 @@ pub struct Manifest {
 pub enum ManifestType {
     Npm,
     Cargo,
+    #[strum(to_string = "pyproject.toml")]
```
spenserblack: Do we want to set the `to_string` to `"pyproject.toml"`? We don't call the `Npm` variant `"package.json"`, or the `Cargo` variant `"Cargo.toml"`.

manifest/src/lib.rs
```diff
@@ -85,10 +88,92 @@ fn parse_npm_manifest(path: &Path) -> Result<Manifest> {
...
+    let project = pyproject
+        .project
+        .context("pyproject.toml has no [project] table")?;
+
+    // A `license = { file = "LICENSE" }` table carries no identifier, so leave it
+    // unset and let onefetch fall back to detecting the license from the repo.
+    let license = project.license.and_then(|license| match license {
+        PyProjectLicense::Spdx(spdx) => Some(spdx),
+        PyProjectLicense::Table { text } => text,
+    });
```
spenserblack: I was a bit confused by this for a moment, but, if I understand it correctly, the value of `license.text` (if `license` is a table) may *also* be the SPDX identifier? But it seems like `license.text` may be any string if the project uses a non-standard license? Could you clarify this in the comment?

manifest/src/lib.rs
```diff
@@ -85,10 +88,92 @@ fn parse_npm_manifest(path: &Path) -> Result<Manifest> {
...
+            matches!(text.license, Some(PyProjectLicense::Table { text: Some(t) }) if t == "Apache-2.0")
+        );
+
+        // `{ file = "LICENSE" }` table carries no identifier
+        let file: PyProjectTable = toml::from_str("license = { file = \"LICENSE\" }").unwrap();
+        assert!(matches!(
+            file.license,
+            Some(PyProjectLicense::Table { text: None })
+        ));
+    }
```
spenserblack: Instead of using 3 `assert!`, I think you can use `rstest` to make this 3 test cases instead. That way a prior failure wouldn't block the following test cases from running.

spenserblack (commented):
Thanks! Mostly looks good, I just have a few nitpicks.

ChrisJr404:
Thanks for the review! Pushed a commit addressing all three:

1. **`to_string = "pyproject.toml"`** — good catch, that was inconsistent with `Npm`/`Cargo`. Dropped the override so `PyProject` renders via the default variant name, matching the others.

2. **`license.text` comment** — you've got it exactly right: `text` is free-form. It's *conventionally* an SPDX identifier (e.g. `MIT`), but PEP 621 permits any string, including the full text of a non-standard license. We surface whatever's there verbatim rather than trying to validate it as SPDX. I expanded the doc comment on `PyProjectLicense` and the inline comment in `parse_pyproject_manifest` to spell this out, and to note that a `{ file = ... }` table carries no identifier (so we fall back to repo license detection).

3. **`assert!` → `rstest`** — converted to three `#[case]`s so a failure in one form no longer blocks the others. Added `rstest` as a dev-dependency on the `onefetch-manifest` crate (the workspace already pins `0.26.1`).

manifest/Cargo.toml
```diff
@@ -13,3 +13,7 @@ cargo_toml = "1.0.0"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.150"
 strum = { version = "0.28.0", features = ["derive"] }
+toml = "1.0.0"
+
+[dev-dependencies]
+rstest = "0.26.1"
```
spenserblack: I believe you should be able to use `rstest.workspace = true` to ensure that this version and the overall project's version of `rstest` remain synchronized.


```suggestion
[dev-dependencies]
rstest.workspace = true
```

manifest/src/lib.rs
```diff
@@ -85,10 +87,101 @@ fn parse_npm_manifest(path: &Path) -> Result<Manifest> {
...
+    use super::*;
+    use rstest::rstest;
+
+    #[rstest]
+    // SPDX expression string (PEP 639)
+    #[case("license = \"MIT\"", Some("MIT"))]
+    // `{ text = "..." }` table (older PEP 621 form)
+    #[case("license = { text = \"Apache-2.0\" }", Some("Apache-2.0"))]
+    // `{ file = "LICENSE" }` table carries no identifier
+    #[case("license = { file = \"LICENSE\" }", None)]
```
spenserblack: Looks good! Rather than use comments, I think we can just name the test cases:

```suggestion
    #[rstest]
    #[case::pep_639("license = \"MIT\"", Some("MIT"))]
    #[case::pep_621_text("license = { text = \"Apache-2.0\" }", Some("Apache-2.0"))]
    #[case::pep_621_file("license = { file = \"LICENSE\" }", None)]
```

spenserblack (changes requested):
Thanks! Just these last two changes and LGTM!

spenserblack:
I hate to do this, but one more thing, since it looks like you've managed to author over 90 pull requests today alone: have you read our contributing guidelines (including our AI policy)?

spenserblack (changes requested):
Oops, I thought just using `rstest.workspace = true` was enough to synchronize the versions, but looks like a little bit more work is needed.

TBH you can just revert bd4491b05ac006fe99447790c373fc9e1857601a. I'll look into utilizing that feature some other time.

PR #1847, opened 2026-08-24
Support non-colocated Jujutsu repositories

### LINKED ISSUE #1846: Jujutsu Support
### Summary 💡

It should be able to interact with jujutsu only repos not just git & jj colocated.

### Motivation 🔦

I need to do jj git init --colocate just so I can use onefetch in repos where it wouldn't have been necessary

lainon1:
Closes #1846.

Native Jujutsu repositories keep their Git store inside `.jj`, so normal Git discovery cannot find them.

This change:

- detects non-colocated JJ workspaces;
- uses `jj git root` to locate the backing Git store;
- obtains the current JJ working-copy commit;
- reuses the existing Git history and statistics implementation;
- uses the JJ workspace root for language, manifest, and license detection;
- avoids snapshotting or modifying the JJ working copy.

The transient JJ working-copy commit is displayed as HEAD but excluded from commit statistics. Its ancestors, including multiple merge parents, are still traversed.

`Pending` and `Size` are omitted because the hidden backing repository has no Git worktree or index.

Testing:

- `cargo test -p onefetch` — 133 passed
- crate-level Clippy — clean
- manually tested with `jj git clone --no-colocate`

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -15,6 +15,7 @@ use self::license::LicenseInfo;
 use self::loc::LocInfo;
 use self::pending::PendingInfo;
 use self::project::ProjectInfo;
+use self::repository::Repository as DiscoveredRepository;
 use self::size::SizeInfo;
 use self::title::Title;
 use self::url::UrlInfo;
@@ -46,6 +47,7 @@ mod license;
 mod loc;
 mod pending;
 mod project;
+mod repository;
 mod size;
 mod title;
 mod url;
@@ -109,8 +111,9 @@ impl std::fmt::Display for Info {
 }
 
 pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
-    let repo = gix::discover(&cli_options.input)?;
-    let repo_path = get_work_dir(&repo)?;
+    let discovered_repo = DiscoveredRepository::discover(&cli_options.input)?;
+    let repo = discovered_repo.git();
+    let repo_path = discovered_repo.work_dir().to_owned();
     // Compute LOC in a separate thread so it runs in parallel with commit-graph traversal.
     let loc_by_language_sorted_handle = std::thread::spawn({
         let globs_to_exclude = cli_options.info.exclude.clone();
@@ -127,19 +130,17 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
         }
     });
     let git_metrics = traverse_commit_graph(
-        &repo,
+        repo,
+        discovered_repo.head_id()?,
+        discovered_repo.is_jujutsu(),
         cli_options.info.no_bots.clone(),
         cli_options.info.churn_pool_size,
         cli_options.info.no_merges,
     )
     .context("Failed to traverse Git commit history")?;
     let manifest = get_manifest(&repo_path)?;
-    let repo_url = get_repo_url(
-        &repo,
-        cli_options.info.hide_token,
-        cli_options.info.http_url,
-    )
-    .context("Failed to determine repository URL")?;
+    let repo_url = get_repo_url(repo, cli_options.info.hide_token, cli_options.info.http_url)
+        .context("Failed to determine repository URL")?;
     let true_color = match cli_options.ascii.true_color {
         When::Always => true,
         When::Never => false,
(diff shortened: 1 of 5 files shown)


src/info/mod.rs
```diff
@@ -238,8 +239,8 @@ impl InfoBuilder {
         self
     }
 
-    fn pending(mut self, repo: &Repository) -> Result<Self> {
-        if !self.disabled_fields.contains(&InfoType::Pending) {
+    fn pending(mut self, repo: &Repository, is_jujutsu: bool) -> Result<Self> {
+        if !is_jujutsu && !self.disabled_fields.contains(&InfoType::Pending) {
```
spenserblack, line 249: I think it's pretty obvious, but, just to be clear: we're going to skip the pending changes info for Jujutsu, and this is because Jujutsu's concept of "pending changes" is different from Git's, right? IIRC Jujutsu *always* tracks changes?

If so, maybe add a comment explaining why we're skipping the pending field for Jujutsu.

src/info/mod.rs
```diff
@@ -284,8 +288,13 @@ impl InfoBuilder {
...
 
-    fn size(mut self, repo: &Repository, number_separator: NumberSeparator) -> Self {
-        if !self.disabled_fields.contains(&InfoType::Size) {
+    fn size(
+        mut self,
+        repo: &Repository,
+        number_separator: NumberSeparator,
+        is_jujutsu: bool,
+    ) -> Self {
+        if !is_jujutsu && !self.disabled_fields.contains(&InfoType::Size) {
```
spenserblack, line 303: :thinking: I'm a bit concerned that some Jujutsu users might be confused if they don't get the size field. I wonder if it would be enough to just simply `eprintln!("Jujutsu support is experimental: some fields are not yet supported")` or something like that.

src/info/repository.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    pub fn head_id(&self) -> Result<ObjectId> {
+        match self.jujutsu_head {
+            Some(head_id) => Ok(head_id),
+            None => Ok(self
+                .git
+                .head_id()
+                .context("Failed to retrieve HEAD ID")?
+                .detach()),
+        }
```
spenserblack: Stylistic nitpick: instead of using `match` to convert an `Option` to a `Result`, I think we can use some of the `Option` methods like `ok_or_else`.

src/info/repository.rs
```diff
@@ -0,0 +1,135 @@
...
+fn find_jujutsu_root(input: &Path) -> Result<Option<PathBuf>> {
+    let input = input
+        .canonicalize()
+        .with_context(|| format!("Failed to resolve repository path '{}'.", input.display()))?;
+    let start = if input.is_dir() {
+        input.as_path()
+    } else {
+        input
+            .parent()
+            .context("The repository path has no parent directory")?
```
spenserblack: Should we allow inputs to be file paths? Should we just always fail here, instead, as the user has failed to provide a path to a directory?

spenserblack (approved):
This looks very good, thanks! Just have some nitpicks and comments.

src/info/repository.rs
lainon1: Agreed. I simplified this with Option::map_or_else in 8751184.

src/info/repository.rs
lainon1: Agreed. INPUT is documented as a directory path, so accepting files only through Jujutsu discovery would be inconsistent. find_jujutsu_root now rejects file inputs with a clear error, and the behavior is covered by the unit test in 8751184.

src/info/mod.rs
lainon1: Good point. I added a stderr notice in 8751184 stating that Jujutsu support is experimental and that pending changes and size are not yet supported. Keeping it on stderr leaves normal and serialized stdout intact.

src/info/mod.rs
lainon1: Yes. Jujutsu snapshots working-copy changes into the @ commit, while the existing pending field reads Git index/worktree status. A native Jujutsu repository has a bare backing Git store, so that calculation is not meaningful. I added a comment explaining why the field is skipped in 8751184.

spenserblack (approved):
Thanks!

lainon1:
@o2sh

PR #1853, opened 2026-09-01
Suggest reftable migration when HEAD can't be read

### LINKED ISSUE #1743: The reference at "HEAD" could not be instantiated: Reference name cannot start with a dot

Jorge-Polanco-Roque:
Fixes #1743.

## Problem
On a repository that uses git's **reftable** backend, onefetch crashes with an opaque error:
```
The reference at "HEAD" could not be instantiated
  → The path "refs/heads/.invalid" ... is invalid
  → Reference name cannot start with a dot
```

## Root cause
Git's reftable backend intentionally points `HEAD` at the sentinel `refs/heads/.invalid` so older clients fail early. `gitoxide` cannot read the reftable format yet, so every `HEAD` resolution fails.

## Approach
Detect that specific case and turn the crash into an actionable message suggesting `git refs migrate --ref-format=files`. Detection matches the git-hardcoded sentinel `refs/heads/.invalid` (not a gix message), so it stays stable across gitoxide versions. Applied via `.map_err(...)` at the traversal call — the earliest, user-facing HEAD failure.

## Testing
- `reftable_sentinel_error_gets_migration_hint` and `unrelated_error_is_left_untouched` (the latter guards against over-matching). Verified the hint test **fails without the fix**.
- `cargo fmt --check` clean · `cargo clippy --all-targets` zero warnings · `cargo test --lib` 102 passed.

Note: this is the graceful-error fix (as suggested in the thread); reading reftable itself belongs upstream in gitoxide.

### DIFF
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -132,6 +132,7 @@ pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
         cli_options.info.churn_pool_size,
         cli_options.info.no_merges,
     )
+    .map_err(reftable_head_hint)
     .context("Failed to traverse Git commit history")?;
     let manifest = get_manifest(&repo_path)?;
     let repo_url = get_repo_url(
@@ -469,4 +470,52 @@ pub fn get_work_dir(repo: &gix::Repository) -> Result<std::path::PathBuf> {
         .workdir()
         .context("please run onefetch inside of a non-bare git repository")?
         .to_owned())
+}
+
+/// When a repository uses git's `reftable` backend, `HEAD` is intentionally pointed at the
+/// invalid sentinel `refs/heads/.invalid` to make older clients fail early. `gitoxide` cannot
+/// yet read the `reftable` format, so any `HEAD` resolution fails with an opaque error. Detect
+/// that specific case and turn it into an actionable suggestion.
+///
+/// `refs/heads/.invalid` is a sentinel hard-coded by git itself, so matching on it is stable
+/// across `gitoxide` versions.
+fn reftable_head_hint(err: anyhow::Error) -> anyhow::Error {
+    if err
+        .chain()
+        .any(|cause| cause.to_string().contains("refs/heads/.invalid"))
+    {
+        return err.context(
+            "This repository uses git's reftable backend, which onefetch cannot yet read. \
+             Convert it with `git refs migrate --ref-format=files`.",
+        );
+    }
+    err
+}
+
+#[cfg(test)]
+mod tests {
+    use super::reftable_head_hint;
+    use anyhow::anyhow;
+
+    #[test]
+    fn reftable_sentinel_error_gets_migration_hint() {
+        // Mimics the gitoxide error chain seen on reftable repos (issue #1743).
+        let err = anyhow!("Reference name cannot start with a dot").context(
+            "The path \"refs/heads/.invalid\" to a symbolic reference within a ref file is invalid",
+        );
+        let hinted = reftable_head_hint(err);
+        assert!(
+            hinted
+                .to_string()
+                .contains("git refs migrate --ref-format=files"),
+            "expected reftable migration hint, got: {hinted:#}"
+        );
+    }
+
+    #[test]
+    fn unrelated_error_is_left_untouched() {
+        let err = anyhow!("some other traversal failure");
+        let out = reftable_head_hint(err);
+        assert_eq!(out.to_string(), "some other traversal failure");
+    }
 }
--- a/src/info/mod.rs
+++ b/src/info/mod.rs
@@ -473,17 +473,23 @@ pub fn get_work_dir(repo: &gix::Repository) -> Result<std::path::PathBuf> {
 }
 
 /// When a repository uses git's `reftable` backend, `HEAD` is intentionally pointed at the
-/// invalid sentinel `refs/heads/.invalid` to make older clients fail early. `gitoxide` cannot
-/// yet read the `reftable` format, so any `HEAD` resolution fails with an opaque error. Detect
-/// that specific case and turn it into an actionable suggestion.
+/// sentinel `refs/heads/.invalid` to make older clients fail early. `gitoxide` cannot yet read
+/// the `reftable` format, so resolving `HEAD` fails when it decodes that symbolic reference and
+/// its target name (`.invalid`) fails ref-name validation. Detect that specific typed error and
+/// turn it into an actionable suggestion.
 ///
-/// `refs/heads/.invalid` is a sentinel hard-coded by git itself, so matching on it is stable
-/// across `gitoxide` versions.
+/// We match on `gix`'s [`decode::Error::RefnameValidation`] rather than the error text, so this
+/// stays correct regardless of how the messages are worded.
 fn reftable_head_hint(err: anyhow::Error) -> anyhow::Error {
-    if err
-        .chain()
-        .any(|cause| cause.to_string().contains("refs/heads/.invalid"))
-    {
+    use gix::refs::file::loose::reference::decode;
+
+    let is_invalid_symref = err.chain().any(|cause| {
+        matches!(
+            cause.downcast_ref::<decode::Error>(),
+            Some(decode::Error::RefnameValidation { .. })
+        )
+    });
+    if is_invalid_symref {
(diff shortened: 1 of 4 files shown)


spenserblack (changes requested):
Duplicate of https://github.com/o2sh/onefetch/pull/1755, but the solution here looks a bit cleaner, so I'll leave this open.

Is there any chance that, rather than pattern matching on a string generated by `to_string()`, you could check the actual error type that the `gix` utilities return in `traverse_commit_graph`? Let me know if you feel that this would be overly complex, and I'll approve.

Jorge-Polanco-Roque:
Good call — switched to checking the typed error instead of the message string. `reftable_head_hint` now downcasts the error chain to `gix::refs::file::loose::reference::decode::Error` and matches the `RefnameValidation` variant (an invalid symbolic-ref target), which is exactly what gitoxide raises when it resolves `HEAD` → `refs/heads/.invalid` on a reftable repo — independent of the message wording.

The unit test builds the real typed error via `gix::validate::reference::name(...)`, and I re-verified end-to-end on a `git init --ref-format=reftable` repo. Thanks for keeping this open!

spenserblack (changes requested):
.

spenserblack (approved):
I suspect that this PR violates the AI policy since the replies have been very LLM-like, but overall the code is still decent I suppose.

I'm curious if the user is copying and pasting comments and LLM responses, or if the account is a fully autonomous AI agent. If you are a fully autonomous AI agent, you *must* disclose.

Jorge-Polanco-Roque:
Hi @spenserblack — thanks for approving it, and for asking directly. I'd rather be upfront.

Yes, I use AI, but I'm not an autonomous agent: there's a real person (me) behind this account, and I review and approve every change and every comment before it's posted. Nothing here is unattended.

I lean on AI in two ways. First, for **code quality and security review** — a second pair of eyes on edge cases and potential issues before I open a PR. Second, to help me **write my replies**, because English isn't my first language and I want my comments to be clear and correct.

If any of that doesn't fit the project's AI policy, tell me and I'll adjust or step back — no problem at all. Thanks again for the thorough review.

spenserblack:
As long as there's a human involved in the process, and the PR isn't fully automated, you're good. I don't mean to block people from contributing.

Jorge-Polanco-Roque:
Thanks, that's a very fair line to draw — and yes, there's always a human (me) reviewing and approving every change and comment before it goes out. Since it's approved and CI is green, is there anything else you'd like me to tweak, or is it good to go? No rush.

spenserblack:
LGTM. But I typically don't merge PRs until @o2sh has had a chance to review, unless it's a small change.

spenserblack:
FWIW sometimes it's a good idea to post both the original text in your native language *and* the translated text, just so you have a record of what you originally said in case there's a misunderstanding.

o2sh:
Sorry for the late answer.

It's too bad gitoxide doesn't support reftable yet.

It looks goot to me but maybe @Byron you can take a look to confirm this is the proper way to handle this case.

We should probably also add a comment saying that it is temporary until support is added upstream by gitoxide

Byron:
Supporting Reftable is the next big feature to be implemented, and it will happen this year.

Meantime, what's problematic for applications is that there is no obvious error message:

```
> gix status
Error: The reference at "HEAD" could not be instantiated

Caused by:
    0: The path "refs/heads/.invalid" to a symbolic reference within a ref file is invalid
    1: Reference name cannot start with a dot
```

The problem is that it allows to open the repository fine, but trying to read HEAD will fail. I think it's worth a remedy to special-case the `.invalid` ref name in HEAD and say something about reftable not being supported yet.

Meanwhile, I would certainly make the call-to-action "Convert it with `git refs migrate --ref-format=files`." weaker, as it casually demands to make an impactful decision about how references are handled just to use `onefetch`.

o2sh:
Yes the call to action should be removed IMO.

BTW, I think it would be cleaner to just bail if the repo uses reftable instead of waiting for gitoxide to fail on this case.

@Jorge-Polanco-Roque 

you can add a function in git/mod.rs (taken from https://github.com/starship/starship/pull/7154)

```
pub fn uses_reftables(repo: &gix::Repository) -> bool {
    repo.config_snapshot()
        .string("extensions.refstorage")
        .is_some_and(|kind| kind.as_bstr() == "reftable")
}
```

and bail in build_info if it returns true:

```
    if uses_reftables(&repo) {
        // TODO: remove once gitoxide supports reftable
        bail!("reftable repositories are not supported");
    }
```

also it would be nice to add an integration test (cf. tests/fixtures), you can add a sh script `make_reftable_repo.sh`

```
set -eu -o pipefail

mkdir reftable
cd reftable
git init -q --ref-format=reftable
```

and assert it fails:

```
#[test]
fn test_reftable_repo_is_rejected() -> Result<()> {
    let repo = named_repo("make_reftable_repo.sh", "reftable")?;
    let config = CliOptions {
        input: repo.path().to_path_buf(),
        ..Default::default()
    };
    let error = match build_info(&config) {
        Ok(_) => panic!("reftable repository should be rejected"),
        Err(error) => error,
    };

    assert_eq!(error.to_string(), "reftable repositories are not supported");
    Ok(())
}
```

Jorge-Polanco-Roque:
Reworked it the way you suggested. Ditched the HEAD-failure hint and now bail up front: added uses_reftables() in git/mod.rs and the early bail in build_info with the TODO. Also added an integration test with a make_reftable_repo.sh fixture so it's covered. The starship reference helped, this reads a lot cleaner than what I had. Thanks for taking the time.

src/info/mod.rs
```diff
@@ -110,6 +111,10 @@ impl std::fmt::Display for Info {
 
 pub fn build_info(cli_options: &CliOptions) -> Result<Info> {
     let repo = gix::discover(&cli_options.input)?;
+    if uses_reftables(&repo) {
+        // TODO: remove once gitoxide supports reftable
+        bail!("reftable repositories are not supported");
```
spenserblack, line 116:
```suggestion
        bail!("reftable repositories are not yet supported");
```
nitpick

spenserblack (approved):
Overall LGTM! This does look a lot cleaner.

I do have one small concern. My concern is that users might see this and get confused, wondering *why* a "reftable repository" is not supported, or what a "reftable repository" even is.

This is a temporary fix, so maybe it would make sense to include a link to https://github.com/o2sh/onefetch/issues/1743 in the error message for more context? A user can follow that link, read up on the issue, and decide on their own if they want to migrate the repository or wait for Gix and Onefetch to be updated.

PR #1856, opened 2026-09-06
fix: count C++ modules as C++

### LINKED ISSUE #1855: C++ Modules doesn't count as a part of C++ source code
### Duplicates

- [x] I have searched the existing issues

### Current behavior 😯

Tokei already supports C++ modules, but onefetch does not display them as part of the C++ source code.

### Expected behavior 🤔

Onefetch should include C++ Modules in C++ statistic.

### Steps to reproduce 🕹

1. Create any .ixx/.cppm/.cxxm/.c++m/.ccm file
2. Done

As we can see here:
Image

### Possible Solution 💡

_No response_

be-student:
Closes #1855.

Adds a reusable Tokei alias mapping, maps `CppModule` into Onefetch’s existing C++ language, and sums aliased LOC buckets. Tests cover both scan selection and aggregation.

Validation: `cargo test`; `cargo clippy --all-targets --all-features -- -D warnings`; `cargo fmt --check`

### DIFF
--- a/src/info/langs/language.tera
+++ b/src/info/langs/language.tera
@@ -46,6 +46,11 @@ impl From<tokei::LanguageType> for Language {
         match language {
             {% for language, _ in languages -%}
                 tokei::LanguageType::{{ language }} => Self::{{ language }},
+                {% if languages[language].tokei_aliases is defined -%}
+                    {% for alias in languages[language].tokei_aliases -%}
+                        tokei::LanguageType::{{ alias }} => Self::{{ language }},
+                    {% endfor %}
+                {% endif -%}
             {% endfor %}
             _ => unimplemented!("Language {:?}", language),
         }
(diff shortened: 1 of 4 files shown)


src/info/langs/language.tera
```diff
@@ -63,6 +68,21 @@ impl From<Language> for tokei::LanguageType {
 }
 
 impl Language {
+    pub fn get_tokei_types(&self) -> Vec<tokei::LanguageType> {
```
spenserblack: Because the contents of each collection are known at compile time, I believe we can use a static reference to a slice rather than a `Vec`. This should avoid a heap allocation.

spenserblack (commented):
Overall LGTM, and is actually cleaner than what I was planning. But I think perhaps a bit more discussion should occur first.

@o2sh, what do you think?

spenserblack (approved):

PR #1868, opened 2026-09-19
Update Elixir colors to the official brand color

georgeguimaraes:
Elixir now publishes an official brand color, `#8847B9`, on its trademarks page: https://elixir-lang.org/trademarks/. This updates the chip from the old linguist value and adds a true-color hex for the ASCII logo, which only had the basic magenta before.

The matching linguist change is open at github-linguist/linguist#8212.


spenserblack (approved):

o2sh (approved):
Thanks a lot @georgeguimaraes 🎉

